PR #2, opened 2018-04-26
rfc: package aliases

zkat:


Dwood15:
Does this allow me to alias a package for use in another library? For example, blessed hasn't been updated in ~3 years. In order to change it to support Windows, I manually fork the library and change one dependency, and publish to npm.

Then, I reference the forked blessed in my own code. This process is kind of silly when the end goal is a drop-in replacement of a dependency of a project that is, itself, a dependency.

zkat:
This RFC explicitly calls out that it is -not- intended to inject packages as transitive dependencies. You can only alter what -your-package will import as a certain name -- transitive deps (and dependents) will not be affected.

iarna:
Do we want to allow publication of modules that have aliases as dependencies or do we want to limit this to unpublished (application level) packages?

The advantage of limiting this means that we won't be introducing modules to the ecosystem that users of older package managers (most users) won't be able to install.

If we do that, there probably does need to be some sort of escape valve for private publication and private registries. It could be as simple as allowing `npm publish $(npm pack)`, but disallowing `npm publish`.

ljharb:
It might be a good idea to limit it initially - that way bugs can be worked out, and userland patterns can develop, before exposing it to the wider npm ecosystem.

If it is limited, hopefully it would be either stripped from packages on publish, or better, `npm publish` would error out if the fields are present.

Dwood15:
The problem with limiting features initially and then expanding those later is that there will always be people + organizations behind in versions of npm.

We could give package maintainers a warning that publishing an alias sets an implicit (or explicit) minimum npm version of <currentversion>.

Any package which already exists that which adds an alias to their repository would require a maintainer to implement a major version bump. 

On a similar note, if this does break installing packages on older npm installs, ``engineStrict`` will need to be updated by way of minimum requirements in package.json upon publishing to reflect the then-current version of npm.

zkat:
I fully intend to make this publishable. People already publish packages that are uninstallable on earlier versions of npm. Since this syntax is compatible with Yarn's, I'm even more willing to include it in publishes.

We could help the situation a bit by making sure this patch lands with `npm@7.0.0` or some other semver-major patch, because it'll make it easier to communicate what version is needed.

zkat:
(also worth noting: users who publish packages with aliases will likely get plenty of bug reports if their target userbase hasn't upgraded enough, and it's up to them to decide whether their package will be compatible with older npm versions. I'm TOTALLY fine with this and don't consider it ecosystem-splitting because it's a single branch going forward)

accepted/0000-package-aliases.md
```diff
@@ -0,0 +1,81 @@
...
+
+1. installing multiple versions of a same-name package side-by-side (`npm i jquery3@npm:jquery@3 jquery2@npm:jquery@2` -> installs `node_modules/jquery(2|3)`
+2. installing a registry-published fork under the name of the original package (`npm i npm@npm:@zkat/npm`)
+3. shorter, more convenient import names for packages that are otherwise longer than desired (`npm i npa@npm:npm-package-arg`)
+
+This RFC should solve all three of those problems.
+
+There are other, related problems that this RFC will NOT address:
+
+1. replacing packages in transitive dependencies that have different names (`npm i underscore@npm:lodash` will only make it so `require('underscore')` _in your own project_ loads `lodash` instead. Your dependencies will receive their own (nested) version of the actual `underscore` package). This may be covered in the future by an explicit resolutions mechanism, and may be facilitated by this RFC, but is still not part of it.
```
ljharb: Flattening of deps in npm 3+ seems like it will mean that this boundary isn’t in fact strictly or reliably adhered to. If a transitive dep depends on underscore at a version matching the aliased underscore, wouldn’t it use that one?

zkat: no. That's exactly what this paragraph is saying. Dependencies are identified by _both_ name and spec version. `underscore@npm:lodash@1.0.0` and `underscore@1.0.0` are _entirely different_, and _incompatible_ dependencies, so they won't be flattened.

ljharb: ah ok gotcha, so the aliasing transparently applies to the filesystem folder name, but not to the "ideal dep graph" logic?

zkat: both.

PR #3, opened 2018-04-26
rfc: new npm link command

iarna:
This was originally discussed here: https://github.com/npm/npm/issues/19421


accepted/0003-npm-link-changes.md
```diff
@@ -0,0 +1,191 @@
...
+2. Leave all links alone. This would restore previous behavior around `npm link` and `npm install --link` but would make it impossible for npm to do the right thing if a `package.json` dependency changed from a `file:` specifier to a registry one as that would be indistinguishable from a link created with `npm link`.
+
+The second option is appealing for existing users, but introduces an inconsistency with how npm models it's use. npm views its job as synchronizing your `node_modules` with your `package.json` and lock-file.
+
+## Implementation
+
+Broken out by command:
+
+### `npm link`
+### `npm unlink
```
1j01: Missing backtick here (and is this supposed to be `npm link` & `npm unlink` together (like the heading below)? or is there something missing for `npm link`?)

gilly3:
What you are calling "ephemeral" may be a misnomer.  The distinction in that use case seems to be that you are linking to a package in a separate repo, and that it may not be linked for all (or any other) developers.  Whether or not it is a temporary state for "trying things out" doesn't matter.  I have use cases where links will differ between developers, but the links may exist long-term, even permanently.

I work on several different apps that share libraries.  Each app or library is in its own repo.  Not all developers on my team work on all apps or libraries, so it would be inappropriate to commit a package-lock dictating links to repos that may not even be cloned on other developers' systems.

Further complicating matters, we prefix all our internal packages with our company initials to avoid naming collisions.  (These are published in a private registry.)  However, when changing directories in the command line, I find it tedious typing that prefix, so I clone the repos into directories named without the prefix.  Other developers do not do this.  So, where I might want to link to `../fancylib`, my co-worker would link to `../abc-fancylib`.

The point is that for many developer workflows, linking to local packages is going to be specific _to the developer_, and not specific to the package.  Thus, committing a package-lock with links will cause headaches.

JDvorak:
I consider `npm link` to be a local configuration override on the default behavior, not a specification to be shared between developers. 

For instance, if I am working on an ecosystem of modules, then I will be (tirelessly) maintaining links between my local copies in a directory. However, when I incrementally publish my improvements, _I would not expect the links to be submitted too._ Nor, would I want them to-- especially in 1 micro-component per 1 developer architecture.

evocateur:
@JDvorak `npm publish` never includes package-lock.json, nor is it ever respected by a consuming `npm install <pkg>`, why would it matter what local links exist in it?

JDvorak:
On further reflection, I realize that there are two different understandings of what a dependency is:

1.  A dependency is a parameter of a module. Compatible versions, and entirely different modules can and should be swappable. This metaphor arises in the module developer's use of npm link as an explicit alternate binding, and the praise of semver as ideal. You can also see this metaphor in action when someone uses proxyquire for testing, or writes a shim to approximate the separation from the 'symbolic requirement' and its 'concrete instance.'  I think you also see this reasoning in most discussions of npm aliases. This metaphor naturally lends itself to compositionality and is, presumably, the preferred mode of reasoning that the functional purists think about dependencies.

2. It is a _part_ of a module, an essential attribute of its functioning. It is not a variable element but a specific value. This relies on a building block metaphor where an NPM module is an immutable block of code itself built up of immutable blocks of code. The metaphor of choice for pragmatists who recognize irreducible coupling between modules in business logic. Here I feel is where the idea of package-lock appears obvious, while in the other world it appears equivalent to caching layer. 

But, if you notice, the community seems to keep reinventing ways to express a separation between a variable's declaration and its initialization. The `npm link` I would want is one that allows me to set a different runtime initialization without changing the formal declaration of dependency, however that is because I imagine modules are parameters.

Dwood15:
@JDvorak ```The npm link I would want is one that allows me to set a different runtime initialization without changing the formal declaration of dependency.

I treat modules as parameters. I already have to swap them out on a quarterly basis.```

I'm having trouble seeing the difference between your version of npm link and the aliasing that was just ratified.

JDvorak:
If I could just do what I used to do with `npm link` (and it stops breaking when I `npm install`) but under the hood it just aliases locally, that sounds great.

 I'll miss the ease and intuitiveness of it being represented by a symlink tho, but that's a different fight.

iarna:
@gilly3 

> What you are calling "ephemeral" may be a misnomer. The distinction in that use case seems to be that you are linking to a package in a separate repo, and that it may not be linked for all (or any other) developers.

By "you" I think you mean @giltayar, who I was quoting there. (Tagging them so they can clarify their intent.) What I understood them to be asking for was very-short-term links, something closer to what npm's current behavior results in (where they're kept until you run `npm i` again).

> The point is that for many developer workflows, linking to local packages is going to be specific to the developer, and not specific to the package. Thus, committing a package-lock with links will cause headaches.

If I said that your concern is that there will be package lock churn between your installation and your co-workers because you have atypical way of cloning your repos and so when they make links they'd stomp on yours and vice versa?

Would your concerns be alayed by storing this in a different file? (I'm not super keen on introducing Yet Another File, but in so far as I can tell, it's what's here, adding something new or remove the `npm link` command entirely.)

@JDvorak 

> Because it is respected by anyone running npm install from within a cloned repository.

It would not. `npm link create` would be needed to replicate your local linking.

> If I could just do what I used to do with npm link (and it stops breaking when I run npm install) but under the hood it merely aliases to local directories, well then that sounds great. And, on reading, it does seem that is the case. But, now I'm not sure why we would need a new npm link if we are getting a proper aliasing feature.

I'm confused as to how you think aliasing would actually solve this problem for you? Can you elaborate? Aliases are published as a part of the module metadata. They change the source of the module for everyone that uses them. They are not a local development tool, but a way to ease the use of forks (so you can switch without changing your source code).

wearhere:
+1 for @giltayar's "ephemeral" use case as described [here](https://github.com/npm/rfcs/pull/3/files#diff-7546538a0a13d2f1a05e788b640935f8R192). My coworkers and I use links all the time but I'm pretty sure we wouldn't be interested in sharing them with each other. All our packages are in different repos; and so even if multiple developers were working on the same branch of repo A, which branch required that each developer instantiate a link, committing that link to the branch would not be sufficient to "enable" the link for all developers and would probably be otherwise confusing.

I am very interested in not having `npm install` blow away links though (which I imagine was part of the motivation for saving to the lockfile).

Separately, very interested in having `npm unlink` undo `npm link` and restore a non-linked copy of the dependency, which I had filed as https://npm.community/t/npm-unlink-foo-doesnt-reinstall-the-real-foo/713 before being directed here.

Venryx:
I just want to add a quick note, for others using npm 5 and wanting to restore the old behavior, where "npm install" does not delete symlinked packages: try updating to the latest version of npm 5. (npm 5.10.0, atm)

I had been using the npm 5.6.0 that came with node 8.11, but then saw (this comment)[https://github.com/npm/npm/issues/17287#issuecomment-379650891]. I tried updating to the latest 5.X version of npm, and this seems to have restored the old `npm install` behavior where it just ignores symlinked packages. (or at least doesn't delete them)

This is good enough for me, until we get a revamped `npm link` system in npm 7. (the npm 5.6.0 behavior of deleting symlinked packages -- all of them, silently -- was horrendous)

thw0rted:
Does anybody else have concerns about the revised `link` having no support for `pack` / `publish`?  I currently develop an application and its dependency modules in tandem using `link` and then distribute the application to our test machines using `pack`.  Today's (6.x) behavior kind of works, but [the whole dependency directory is shoved in](https://npm.community/t/linked-bundleddependency-doesnt-respect-npmignore/1297) which is obviously sub-optimal.

I'd love to see `pack` recurse into linked deps and `pack` them, then use the resulting output in lieu of a straight directory copy (6.X) or just giving up (this RFC).  Does that make sense?  Is it too difficult to implement?

stacktracer:
Anybody know the answer to the following question (copied from [the previous thread](https://github.com/npm/npm/issues/19421#issuecomment-353697305))?

2. **What happens when the version changes in a linked package?** When is it noticed? What is the expected behavior if the new version causes it to no longer match the dependency spec of the consuming package? will npm update the symlink to an install?

giltayar:
The latest commit (npm-link: Revise for non-lock-file storage) looks perfect. Just what the doctor ordered. My monorepo build tool just got simpler! :-)

ties-s:
So what's the status on this? npm 7 still deletes linked modules

Venryx:
@ties-s What version of npm did you try it with?

I ask because, in [this issue](https://github.com/npm/cli/issues/2372), an npm dev (@nlf) suggests the bug may have been fixed by some changes to npm link between version 7.2.0 and 7.5.2, but the OP did not confirm whether those changes fixed it or not. (the lack of response suggested it was resolved, though)

It will be good for someone to confirm if the bug is still present in the latest versions. (7.5.2+)

ties-s:
@Venryx I have npm@7.6.1

ljharb:
Try v7.7.5 just in case, and if there’s still problems, please file an issue.

cyberbiont:
Just quickly tried it with npm 7.8.0 and the bug is still there.

joe-gre:
I tried with npm 7.8.0 and the bug is still there. Running `npm install` removes linked packages. Also running `npm link <package_name>` removes all other links, which is a newer issue that didn't occur on older versions of npm whereas the other issue has occurred since npm 5 I believe. The only way to setup multiple links is to run  `npm link <package_1> <package_1>` and run this command again after running `npm install`.

 This is a major issue for my team. We have to waste a lot of time relinking after installing new dependency versions.

thw0rted:
@joe-gre  it sounds like you need to comment on https://github.com/npm/cli/issues/2372 which is marked as closed, and apparently shouldn't be.

IronGeek:
Still happens with npm 7.11.2... :disappointed:

baelter:
And 7.18.1 ...

8enSmith:
And 7.24.0...

and0p:
> I tried with npm 7.8.0 and the bug is still there. Running `npm install` removes linked packages. Also running `npm link <package_name>` removes all other links, which is a newer issue that didn't occur on more recent versions of npm whereas the other issue has occurred since npm 5 I believe. The only way to setup multiple links is to run `npm link <package_1> <package_1>` and run this command again after running `npm install`.
> 
> This is a major issue for my team. We have to waste a lot of time relinking after installing new dependency versions.

This threw me off as well, since an NPM upgrade on one machine made only the last-linked stick around. I now have a `link.sh` in the repo for all of my projects that need it, which runs `npm install` followed by `npm link local1 local2 ...`. It's not ideal but saves on a lot of typing.

PR #4, opened 2018-05-01
Find unmaintained

Dwood15:
Allow users to search their package.json for old/unmaintained packages as part of an audit subcommand or flag.

Shoutout to @iarna


zkat:
I'll drop a comment about potential impact here when considering answers: There's an *immense* number of packages that haven't been updated in well over a year (or several years), which are still in widespread use -- and they're not bitrotted, but considered stable, and are perfectly usable. I'm concerned about the unnecessary noise something like this might add and I'd like to hear more about mitigations to that. I think folks would be surprised at the number of low-maintenance packages in their trees.

iarna:
It might be more appropriate to only list something as unmaintained when there are higher ranked alternatives available. That at least makes the information actionable. We do have a maintenance quality metric currently on the website, fwiw.

nickmccurdy:
I don't think this is necessary given that npm is pretty good at supporting the installation of older packages compared to other package names. Commands like `outdated`, `deprecate`, and `audit` already solve the maintenance and security concerns of using old packages well enough.

zkat:
Hey! Thanks again for filing this RFC!

We've decided to pass on this particular proposal, based on prior discussion and responses from other users.

So, I'm gonna close this, but I look forward to other ideas coming our way that address these pain points!

PR #5, opened 2018-05-02
rfc: Don't change EUID/perms when running as root

iarna:


accepted/0000-running-as-root.md
```diff
@@ -0,0 +1,108 @@
...
+
+On Unix-like systems, access to modify system installations is traditionally
+limited to user id `0` (traditionally named `root`).  Modern systems allow
+more complicated scenarios through the user of access control lists, but
+they are out of scope for this discussion.
+
+Modern best practices are that access to the `root` user is achieved through
+a command called `sudo` which authenticates the current user and can limit
+which commands may be run. It also provides logging. When running commands
+with `sudo`, information about who the real user running the command is.
```
evocateur:
```patch
-with `sudo`, information about who the real user running the command is.
+with `sudo`, information about who the real user running the command is preserved.
```

accepted/0000-running-as-root.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Administrator on Windows
+
+On Windows systems the equivelant is to run things as `Administrator`.  Like
+`sudo`, you allow this using your credentials, not the credentials of a
+separate administrator account.  Like `su`, ordinary practice with the
+terminal is to open up a new terminal window with elevated permissions and
+run commands from there.
+
+Currently, there the only way for a Node.js program to determine it is
```
zkat: s/there //

accepted/0000-running-as-root.md
```diff
@@ -0,0 +1,108 @@
...
+its scripts as root.  It will use the config and cache in root's home
+directory.
+
+## Implementation
+
+### Configuration
+
+The `.npmrc` in root's home directory is used.
+
+When running under sudo, configuration from the user's `.npmrc` is used, but
```
zkat: Do other tools follow this style? It might be nice to mention precedent. For example with git and ~/.gitconfig

accepted/0000-running-as-root.md
```diff
@@ -0,0 +1,108 @@
...
+
+Files created in the cache will use the same user and group as the top level
+cache folder, or, if that does not exist, the folder that contains it.
+
+### Installation
+
+#### Local Installation
+
+Local installation with `sudo` into a directory not owned by root
+requires `--force` to complete.
```
zkat: I don't feel like this does anything of use besides make the user repeat the command. `--force` (ironically much like `sudo`) ends up being just an automatic thing. I'd rather warn, which at least gives users a heads up, possibly with a message saying to run `sudo chown` on their project directory if they ran this by accident.

accepted/0000-running-as-root.md
```diff
@@ -0,0 +1,108 @@
...
+
+All of the standard Unix userland writes files as root when running as root.
+
+## Unresolved Questions
+
+Should npm running as root make any attempt to read the user's auth
+information?  Not doing this means you'd have to separately `sudo npm login`
+if you wanted to install private packages as root.  On the other hand, KISS
+principles suggest that we should not try to be magic here. In particular, we
+_don't_ want other user-level config, like cache directories.
```
zkat: I am totally 100% behind making users log in again and just generally treat sudo as a separate universe.

godmar:
Is there any precedence for package managers leaving files in root's home directory?  Personally, for the software I use, I have not seen any - my `/root` is empty.  The [Linux Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) describes the locations that should be used for system wide configurations (`/etc`) and caches (`/var/cache`), not `/root/.npmrc` or `/root/.npm/_cacache` as it would be under this proposal (if I understand it correctly).

PR #6, opened 2018-05-03
Drop in replacements

Dwood15:
I apologize if this is a little disorganized/eclectic. Still working through how this all fits in with npm. Stripped out a large portion of the original doc I had written after hearing there is a WIP on an rfcs for deep aliasing, and didn't want to duplicate that effort.

This rfcs is about marking a package as being a viable candidate for both aliasing and/or deep-aliasing for another package. 

Should this get accepted, being able to search for packages in lieu of another package, regardless of depth, should become pretty easy.

### DIFF
--- a/accepted/0000-drop-in-replacements.md
+++ b/accepted/0000-drop-in-replacements.md
@@ -1,0 +1,35 @@
+# {{TITLE: a human-readable title for this RFC!}}
+
+## Summary
+
+{{A concise, one-paragraph description of the change.}}
+
+## Motivation
+
+{{Why are we doing this? What pain points does this resolve? What use cases does it support? What is the expected outcome? Use real, concrete examples to make your case!}}
+
+## Detailed Explanation
+
+{{Describe the expected changes in detail, }}
+
+## Rationale and Alternatives
+
+{{Discuss 2-3 different alternative solutions that were considered. This is required, even if it seems like a stretch. Then explain why this is the best choice out of available ones.}}
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


nickmccurdy:
This would be great for tracking packages that are renamed in major releases, especially now that packages like `@babel/core` and `@material-ui/core` and moving to organizations. It's confusing when you update the old package names to the latest version and don't realize that you're missing new versions at a different package name.

styfle:
This would be amazing!

The web interface already shows scores for Popularity, Quality, and Maintenance...so adding the ability to compare "Alias Packages for X" in the web interface would be excellent when comparing these aliases 👌 

Something like this (but instead of search on keyword, it would be alias):
https://www.npmjs.com/search?q=keywords:babel

zkat:
After some discussion, we decided that the upcoming overrides API (RFC upcoming, /cc @aeschright) solves this problem without the added complication of registry-side changes.

styfle:
Can you link to the RFC when it’s available so we can follow it?

aeschright:
Here's the overrides RFC: https://github.com/npm/rfcs/pull/27

Dwood15:
@zkat I feel like it's a mistake to not add registry changes that allow searching for overrides

hutson:
The overrides API seems targeted at the consumer to empower them to replace packages in their tree.

However, it does not empower the package maintainer to communicate and encourage consumers to adopt replacement packages (in the case of renames, scope migrations, etc.)

:+1:  on re-opening this RFC please

PR #7, opened 2018-05-05
Changelog

dantman:
[View formatted RFC](https://github.com/dantman/npm-rfcs/blob/changelog/accepted/0000-changelog.md)

### DIFF
--- a/accepted/0000-changelog.md
+++ b/accepted/0000-changelog.md
@@ -1,0 +1,52 @@
+# Changelog
+
+## Summary
+
+This RFC introduces a `"changelog"` field in `package.json` and a `changes` command in `npm`.
+
+## Motivation
+
+The location where a project documents changes varies to an extreme extent, for every single package you want to upgrade you have to hunt this down separately. Depending on the package the changelog can be: a CHANGELOG.md file on GitHub, HISTORY.md file on GitHub, a releases page on the project's website, the releases page on GitHub if the project documents releases there, a section at the bottom of the README, the commits list on whatever vcs hosting the project uses if the project doesn't maintain a changelog, or something else. You have to hunt for this location every time you want to update your dependencies and do it individually for each package you want to update.
+
+There are also multiple 3rd party commands for doing interactive upgrades (`npm-check`, `yarn outdated`, `yarn upgrade-interactive`) which list the package's homepage when listing a package that has changes. These commands would be much more useful if they were able to list a changelog url instead of a homepage url. Then the url for the package you see while picking packages to upgrade can be a link directly to the page that tells you what has changed in between your version of the package and the newer version(s).
+
+If you are not using these commands it is still highly beneficial to be able to check the changelog of packages by opening it with a `npm changes` command.
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think the `changelog` field should use `https://github.com/<user>/<repo>/releases` automatically if not defined and the `repository` field is specified. This would mean it would automatically *just work* for most packages without any changes. Inference is used for many other fields too, like the `homepage` and `bugs` fields being inferred from the `repository` field.

zkat:
I'm gonna point out there's also an alternative proposal over at #8 that takes a different approach than this. Neither of these is ratified, and feel free to discuss how y'all see these two interacting/intersecting.

My first impression of this specific proposal so far was that even though I really want a feature like this, I felt like we would be able to do so much more for this experience (even if this proposal is still part of the story we end up with). There's also some willingness on the npm website team side to support this sort of thing, which should let folks have a smoother experience in some cases, and we definitely know we could do some nice things related to this because changelog files already get included in tarballs. npm is definitely in the sort of position where we can set _new_ standards, after all, rather than build features only for limited-but-existing standards 💚 

Thanks for posting this! I look forward to the discussion developing and seeing what we all end up with. I'm excited!

dantman:
> I think the changelog field should use https://github.com/<user>/<repo>/releases automatically if not defined and the repository field is pointing to a GitHub repo.

In my experience updating packages, only a very small number of libraries use the releases page for their changelog. Most of the time it's quicker to find the changelog from the readme instead of the releases page.

dantman:
> I'm gonna point out there's also an alternative proposal over at #8 that takes a different approach than this. Neither of these is ratified, and feel free to discuss how y'all see these two interacting/intersecting.

Auto-detection via files in npm registry instead of via VCS repository didn't occur to me before, so I thought of auto-detection as much harder. If everyone is fine with auto-detection. How about implementing the auto-detection described in #8 and also adding a `changelog` field.

We can use this pattern:

- If a `changelog` field is present, open it
- If a `CHANGELOG`, `HISTORY`, `CHANGES` file was detected in the registry, open the package page to to the changelog section
- Likewise handle GitHub releases automatically somehow (up to you whether the package page should try to read this info, or we should just open the GitHub releases page if it appears that the package documents releases in it)
- Otherwise open the homepage

This will give us automatic handling for the vast majority of packages. When it fails we'll send the user to the package's homepage where it is most likely that they can find their way to wherever a changelog is located. And any package with a custom changelog can use the `changelog` field to provide an explicit URL to wherever they host their changelog.

CC: @jefflembeck

g0t4:
- Auto detection will encourage people to use one of said locations which makes it easier to find without tooling. 
- Are there any efforts to standardize a changelog format to make it parseable? I imagine a future where `npm outdated` could show a list of changes per dependency (filterable too: breaking or major changes, new features, bug fixes)

iarna:
We are ratifying the very similar #8, which covers the cli command and existing practices around the bundling of changelogs. The biggest difference is that it declines to add a new package.json field.

PR #9, opened 2018-05-25
rfc: Suggest similar script names for npm run

JoshuaKGoldberg:
This was originally discussed here: https://github.com/npm/npm/issues/20741

Equivalent Yarn issue: https://github.com/yarnpkg/yarn/issues/5867


accepted/0000-run-suggestions.md
```diff
@@ -0,0 +1,142 @@
...
+
+Hard mode: `src:lint`❌ flips `lint:src`✔ around the `:`.
+How common is the `group:specific` pattern?
+It could be useful to also check reverses of the scripts and consider them in the edit distances.
+
+## Rationale and Alternatives
+
+> This is required, even if it seems like a stretch
+
+I'm sorry, but I can't come up with alternatives to help with this problem.
```
legodude17: Alternatives: List all scripts, keep current behavior.

zkat:
I like this, tbh. We also already have a package we use for this sort of thing called [`meant`](https://npm.im/meant) that you can probably add to the implementation section. I believe it should work as you expect. Kudos to @watilde for it :)

zkat:
Also, could you please create a corresponding thread for this in https://npm.community/c/ideas/rfcs?

JoshuaKGoldberg:
https://npm.community/t/suggest-similar-script-names-for-npm-run/147

:+1:

watilde:
I also support this idea to save more time of debugging for developers. Then I made a patch at https://github.com/npm/npm/pull/20879 and it's shipped already :)

We might want to close this issue as resolved.

PR #10, opened 2018-05-31
Creating a npm audit xml report

ChristianStornowski:
Added RFC for generating an xml report via `npm audit --xml`.


accepted/0000-npm-audit-xml-report.md
```diff
@@ -0,0 +1,27 @@
...
+No need to install https://www.npmjs.com/package/nsp and https://www.npmjs.com/package/@ninjaneers/nsp-reporter-owasp as a devDependency.
+
+## Detailed Explanation
+
+Add a possibility to create an OWASP Dependency Check XML (https://www.owasp.org/index.php/OWASP_Dependency_Check) to reuse this xml for SonarQube with this plugin https://github.com/stevespringett/dependency-check-sonar-plugin.
+This should be done by executing `npm audit --xml`.
+
+## Rationale and Alternatives
+
+Instead of implementing an OWASP Dependency Check XML generation it should be possbile to add support for custom audit reporters. So third party reporters can do this job.
```
legodude17: I think you should name other alternatives, such as leaving it the way things are, external tool that turns the output of `npm audit --json` into `xml`.

accepted/0000-npm-audit-xml-report.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Implementation
+
+`npm audit --xml` has to use this XML Schema Definition https://jeremylong.github.io/DependencyCheck/dependency-check.1.7.xsd to generate vulnerable dependencies in a xml file.
```
legodude17: It would be nice if you could go into more detail about this.

accepted/0000-npm-audit-xml-report.md
```diff
@@ -0,0 +1,27 @@
...
+This is similar to `npm audit --json`.
+
+## Unresolved Questions and Bikeshedding
+
+None
```
legodude17: The name of the option could also be `--owasp` or maybe a subcommand.

zkat:
Could you create a corresponding topic in https://npm.community/c/ideas/rfcs for this RFC?

As far as the RFC itself goes: my initial reaction is that npm itself does not, and has not, supported XML. As far as the alternatives section goes, can you talk a bit about why a wrapper package that translates the JSON output to XML is not acceptable or useful here? Not even as a plugin, but as something to pipe to. And if so, what changes would need to be made to the JSON output to support this translation?

I'm curious what the @npm/security team (who are, well, the nsp folks) think of this, too :)

zkat:
Oh I realized I didn't get this across but: I think that the output being a widely-used, standardized format that other things already consumes makes this idea VASTLY more compelling than just doing plain old XML output (and is one reason why I might lean towards `--owasp` as the option, so folks don't start expecting `--xml` elsewhere -- we have a hard enough time being consistent with just `--parseable` and `--json` as-is).

So this RFC is definitely compelling, but I wanna get a sense of the potential (positive) impact :)

accepted/0000-npm-audit-xml-report.md
```diff
@@ -0,0 +1,32 @@
...
+
+Instead of implementing an OWASP Dependency Check XML generation it should be possbile to add support for custom audit reporters. So third party reporters can do this job.
+
+A tool that can process the json format report from `npm audit --json`. That's a valid alternative but why has `npm audit` a json representation and not owasp report for a direct use in a security plugin?
+
+## Implementation
+
+`npm audit --json` is already creating a json report for dependencies. So it should be possible to use a xml generator with this XML Schema Definition https://jeremylong.github.io/DependencyCheck/dependency-check.1.7.xsd instead of using a json generator.
+
+A new file named owasp.js should be created in https://github.com/npm/npm-audit-report/blob/latest/reporters/ and this could use a xml builder like `var builder = require('xmlbuilder');`
```
zkat: We need one more thing here before ratification: We need to see the data that comes through `npm audit` is sufficient to fulfill the contract for `DependencyCheck`. If so, then we're good. Otherwise, we would need to know exactly what's missing (in this section).

We can still ratify it if more data from the server is needed, but we can't promise that it'll be a priority any time soon, when it comes time to actually implement this RFCS.

ChristianStornowski:
Thanks for your comments. :-) I'm looking forward to a ratification.

accepted/0000-npm-audit-xml-report.md
ChristianStornowski: If you need some information I can create a comparison between npm audit json and owasp xml data.

zkat: Yes, please do that. I think it's important for this RFC to have that info.

ChristianStornowski: I created a mapping with pseudo code into accepted/0000-npm-audit-xml-report.md.

numbface:
any idea when this feature will be  ratified ? would be a really nice feature for CI-CD Pipelines

ysnoussi:
Do we know when this is going to be out ?

zkat:
We talked about this RFC and decided to ratify this! Congratulations!

That said, before accepting a PR implementing this, we'd like to see a standalone tool that's a proof of concept that works by consuming `npm audit --json`. Thanks for your efforts on this! I'm glad to see this one land.

jarias:
Did this ever got into NPM or is there a tool described by @zkat that uses `npm audit --json` info to generate an XML report?

bencooper222:
It's hard for me to tell. Is the intention for the outputted xml to be in the junit reporting format so services like CircleCI can consume it? I would hope yes and would be willing to build out POC if so.

PR #11, opened 2018-05-31
WIP: RFC for argument / option parsing

wyardley:
It was suggested that I submit an RFC in https://github.com/npm/npm/issues/20856
This is still missing some detail, and I'll be the first to admit that I don't have a deep knowledge of the existing codebase, so would appreciate any feedback.

### DIFF
--- a/accepted/0000-argument-option-parsing.md
+++ b/accepted/0000-argument-option-parsing.md
@@ -1,0 +1,66 @@
+# Option and Argument Parsing for npm
+
+## Summary
+
+Currently, option, and argument handling in npm is not using any kind of
+standard library (either internal or a third party lib). In addition, invalid
+arguments or flags are not detected for any of the subcommands. Where possible,
+invalid arguments or flags should throw an error.
+
+## Motivation
+
+Originally proposed in [this issue](https://github.com/npm/npm/issues/20856).
+
+Because there's no error thrown when invalid command-line flags are passed, you
+can pass in invalid flags and not realize they're invalid for a particular
+context. e.g., if you type: `npm audit --format=json` instead of `npm audit
+--json` you get the normal output of `npm audit` without any arguments, vs. an
+error that you passed an invalid flag. For `npm install` and some other
+subcommands, this may be unavoidable, for reasons described
+[here](https://npm.community/t/invalid-command-line-flags-are-ignored-vs-throwing-error-usage/164/2).
+Basically, there are cases where people rely on the behavior of additional
+command line arguments getting passed through to scripts that it runs
+indirectly.
+
+## Detailed Explanation
+
+This would pull in a third party lib, and / or create new methods in the core
+libs that would be used to declare subcommands, arguments, and options in a
+standardized way. Ideally, it should be self-documenting (so that usage text,
+auto-completion, help pages, etc. could be as auto-generated as possible).
+
+## Rationale and Alternatives
+
+The main rationale is that exiting non-0 and throwing an error is the correct
+behavior when the user passes in ambiguous or invalid input. 
+
+Additionally, if the command line flags and args are built and documented
+using a standard library, there's less chance for drift between usage / docs
+and reality as new subcommands, flags, and arguments are added. After the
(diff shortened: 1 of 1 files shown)


accepted/0000-argument-option-parsing.md
```diff
@@ -0,0 +1,42 @@
...
+I'm guessing it could also make keeping completion up to date easier.
+
+## Detailed Explanation
+
+This would pull in a third party lib, and / or have a new library (in `utils/`, probably) that would be used to declare subcommands, arguments, and options in a standardized way. Ideally, it should be self-documenting (so that usage text, auto-completion, help pages, etc. could be as auto-generated as possible).
+
+Then, npm itself, as well as its subcommands, could be cut over to use the new lib, either all at once or, more likely, one at a time, and eventually, older utilities could be deprecated or removed.
+
+## Rationale and Alternatives
+
```
zkat: this section isn't really optional.

accepted/0000-argument-option-parsing.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Implementation
+
```
zkat: at least some guesswork here might be good. tl;dr this would have to be done by `lib/config/core.js` or possible `lib/config/defaults.js`.

zkat:
worth noting: this would be a semver-major change.

accepted/0000-argument-option-parsing.md
wyardley: Thanks - I put that in verbatim, HTH. I haven't worked with npm yet, and so I'm guessing others could probably make a much more informed decision about _how_ this should be implemented; happy to put in more information in response to feedback, with or without attribution.

wyardley:
@zkat Thanks! Responded to some of the feedback. Also, formatting-wise, it looks like the other one has lines wrapped; let me know if you all have a standard / preference for wrapping lines or not.

zkat:
@wyardley I don't particularly have a preference for line wrapping in RFCs themselves. I usually wrap my .md to ~80col but as long as it renders well I don't mind.

accepted/0000-argument-option-parsing.md
```diff
@@ -0,0 +1,74 @@
...
+best option(s) would be.
+
+I don't know what specific libs would be best if something isn't written
+within utils  -- I'm not a JS / node dev, but it would seem like tools like
+[minimist](https://www.npmjs.com/package/minimist) or
+[commander](https://www.npmjs.com/package/commander) have few external deps,
+and might make sense to use for implementation. There are also some heavier
+weight options (e.g., [vorpal](https://vorpal.js.org/), but those might be
+tough to fit into an existing tool.
+
```
zkat: Unfortunately, npm can't really use an external options parser :( we have some pretty special behaviors and breaking backwards compatibility with that would be a pretty risky change. I handwave this because the config system predates me by many years and I literally don't even know what all the funny little parsing cases are but I assume some of them are pretty thingy.

zkat: I wish pretty badly that we could use [`yargs`](https://npm.im/yargs), which is probably the one that works closest to npm and is pretty swanky.

wyardley:
Possible that this may just not be able to be accepted, but did make some additional changes based on feedback here and in https://npm.community/t/invalid-command-line-flags-are-ignored-vs-throwing-error-usage/164

iarna:
As a concept, I think everyone is on board with the ideas here. In practice, it's going to require more infrastructure work before this can be approached. For example, the elimination of the config god object and its replacement with `figgy-pudding`. Once that's done, RFCs on a command-by-command basis would be welcome! (The one's that don't run any user scripts are all safe for this.)

It may even be possible to change where custom config options can be accepted, eg, env, config, or commandline, and that sort of change will need to be its own RFC as well. (It's a much higher risk proposition, which is why it needs to be discussed separately.)

I'm going to close this, but again, not because the we're not on board with the idea, but just that it needs to be sliced up a little more.  Thanks for writing this up and bringing it back into discussion!

PR #14, opened 2018-07-06
Create 0002-no-caret-prerelease-installs.md

hzoo:
> Mentioned this to @zkat last month so figured I'd write something at least, (probably want the same in yarn actually if it lands)!

### Rendered: https://github.com/hzoo/rfcs-1/blob/patch-2/0002-no-caret-prerelease-installs.md

Example gissues related to this:
- https://github.com/babel/babel/issues/7786#issuecomment-383586020
- https://github.com/vuejs/vue-cli/issues/1162 (another issue is also with integrations that may use prerelease versions with `^` but that might be a different issue)
- https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420
- https://twitter.com/left_pad/status/980545530559418369

There is a lot of confusion with prereleases in general (both from the package author and user side), so this doesn't prevent anyone from using `^` but it at least prevents people from accidentally adding it `^` given the default.

Might even suggest people get warned of keeping the `^` somehow but maybe that's too much.


0002-no-caret-prerelease-installs.md
```diff
@@ -0,0 +1,45 @@
...
+
+This may be an unexpected experience for end users, as they may not realize they have installed the prerelease version initially, or don't understand that a package author maybe in the midst of other breaking changes that a user would not want to automatically have to deal with on CI or locally.
+
+*Example issue*: https://github.com/babel/babel/issues/7786#issuecomment-383586020
+
+By changing the default behavior to pin to the latest version, a user won't be automatically opted into the latest version and will have to explicitly do it themselves. This way an end-user can decide when to upgrade.
+
+## Detailed Explanation
+
+`npm install`, when writing to `package.json` (like with `--save` or `--save-dev`), should check with semver and only add the `^` if the version isn't a prerelease. I don't believe it should affect anything else.
```
zkat: How do users opt into using ranges for prerelease, if they'd like to? This feature is currently controlled by the `--save-prefix` config, and this RFC makes no mention of how these changes will interact with that option.

iarna:
TBH, I would be quite surprised if you included breaking changes in a pre-release that wasn't a major version above the current non-prerelease? But you're right, nothing says in that case that you'd be wrong to break from pre-release to pre-release…

Something I'd like to figure out for this:

How many packages have pre-release range specifiers currently matching multiple versions? How many projects are expecting the current behavior?

0002-no-caret-prerelease-installs.md
hzoo: Sorry, I didn't know about that option, wasn't really sure what to put for this section actually!

hzoo:
> TBH, I would be quite surprised if you included breaking changes in a pre-release that wasn't a major version above the current non-prerelease?

Yeah this is mostly in the case of Babel right now where the current `latest` is `6.x`.

We are publishing `7.x` as alpha, beta, etc and aren't done making breaking changes (not all of them are intentional but there are definitely some) but yet it doesn't stop anyone from trying to use it in their projects.

> How many packages have pre-release range specifiers currently matching multiple versions? How many projects are expecting the current behavior?

Yeah, that's a good question! I think people would like it since you don't have to continually update but that assumes the non-breaking in-between pre-releases. I think many projects use pre-release as an opportunity to do breaking changes though.

Or hmm.. maybe specifically for a major release at the very least (so a `7.0.0-beta.1` vs `7.1.0-beta.2`)? So not sure if tying it specifically like that would be better then? For a major version, you might plan breaking changes in between while for a minor bump with a beta you might just want to wait a period for regressions?

ljharb:
fwiw i do generally find it surprising when the prerelease line (alpha/beta/etc) has breaking changes inside the same line.

hzoo:
Yeah, I guess this is specifically for when doing a major bump and you just want to get some testing out there but clearly not stable yet. I think it's pretty common for breaking changes in alpha/beta, and sometimes RC for some projects.

Like see [RxJS](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md), others? I guess I don't see how a package author would do it otherwise if they are releasing and landing different kinds of changes before doing the final release?

zkat:
@hzoo hey Henry, just a heads-up that this is ready to land, but we need a small implementation section in order to ratify it. Let us know when that happens! Thanks for your time and patience :)

0008-no-caret-prerelease-installs.md
```diff
@@ -0,0 +1,47 @@
...
+# if this resolves to a prerelease such as 7.0.0-beta.52
+npm install @babel/core
+```
+
+results in
+
+```diff
+"devDependencies": {
+-   "@babel/register": "^7.0.0-beta.52"
++   "@babel/register": "7.0.0-beta.52"
```
SimenB: a bit odd that it says `@babel/register` here when it's `@babel/core` in the install command above 😅

hzoo: lol good point, feel free to PR or I can later

SimenB: Can do, simple enough. Just wasn't sure if it was worth a PR

PR #15, opened 2018-07-08
Change how npm update edits package.json

Maistho:
Rendered: [0000-update-package-json.md](https://github.com/Maistho/rfcs/blob/latest/accepted/0000-update-package-json.md)

Hey, I found the rfcs repo! 🙂 

I've been thinking about writing a proposal for how to change some of my perceived unexpectedness regarding `npm update` for a while, but didn't know where to direct rfcs. The recent updates at https://github.com/npm/npm/issues/16813 pointed me here, so I've written a late night first draft regarding a way to improve the command.

Comments appreciated!


accepted/0000-update-package-json.md
```diff
@@ -0,0 +1,61 @@
+# Change how `npm update` edits `package.json`
+
+## Summary
+
+`package.json` should not be edit by `npm update` unneccesarily. When updating a package to a newer version that still satisfies the original range, package.json should not change. Only when updating to a range that does not satisfy the range specified in `package.json` should it be changed and saved to `package.json`.
+
+## Motivation
+
+It's not expected that the ranges in `package.json` changes when running `npm update`. This was more clearly shown in `npm@5` when `--save` became the default behaviour. Previously, running `npm update` would not edit `package.json`. This unexpectedness also manifests itself in an especially bad way when updating packages that are specified with a non-standard range. For example, packages using tilde (~) when the project default is caret (^), since this will save the new range with the default.
```
ljharb: I don’t agree - when i run `npm update` i assume, and want, the lower range to be lifted.

ljharb: because it establishes a new minimum, and forces deduping in consumer trees, which pushes updates through the graph faster.

Maistho: Hm, I guess that the better behaviour then would be my other considered option 2:
> Make sure that npm keeps the range when updating. For example, updating `~1.0.0` when `1.0.1` is available will yield `~1.0.1`.

It solves my main gripe (non-default ranges not really being very usable) and would still keep the current behaviour of helping consumers get updates for nested dependencies.

ljharb: That part seems fine to me :-)

DanielRuf: https://github.com/npm/npm/issues/16813 is also related (I guess).

Maistho: @DanielRuf yup, as mentioned in the document! 😃 

@ljharb Considering your comments I will be rewriting this RFC to instead suggest that NPM should keep the ranges instead of always writing with the default. Seems like that would be a better solution, all in all.

Dwood15:
Sorry if I misunderstand, but what's the use-case for running an npm update where not even the minor version in package.json gets bumped?

DanielRuf:
> running npm update can install an updated version of a package without changing package.json

I guess it is about the lockfile and the source of truth.

zkat:
So, we like this, and we'd like to have this. This would be a breaking change, so it would have to go into `npm@7`. The behavior here as described is the best formulation of `npm update` I've seen so far, and I'm happy that this RFC got put up!

The only thing this RFC needs before we ratify is generally filling out the `Implementation` section, which can be summarized thus:

* You shouldn't think of it in terms of `--save`: we still need to save the updated `package-lock.json`, and the backflips around the `--save` flag would cause way too much complication. Instead, the `update` command itself shouldn't modify the specifier when it's compatible.
* `npm update`'s behavior when using a _specifier_ (`foo@latest`) needs to change. It's currently this weird filter that no one really understands and isn't really useful in any real way tbh. We'll need to probably rewrite `npm update` such that `npm update foo@version` is the same as `npm install foo@version` -- which is kinda redundant, but probably what people expect.

Implementation-wise, this is a pretty small patch just to `lib/update.js`, and just putzing around with parameters there will be enough to prevent the `package.json` itself from being updated (unless a different package is using `foo@bar`.

Oh yeah, and keep in mind that since you can do `npm update foo bar@latest`, you'll need to separate and group the different args into the appropriate operations (`foo` behaves like an update, `bar` gets a regular installer).

Once you fill that in, I think we'll be ready to ratify this! I look forward to this change! Thank you!

broofa:
Can we add some verbiage that addresses how the `save-prefix` config option should be treated?

Currently `npm update` applies the `save-prefix` setting, but that results in [unexpected](https://github.com/npm/npm/issues/16813#issuecomment-311071479) behavior.  I think this option should just be ignored if a module is already in package.json.

zkat:
@broofa If you mean make it so this RFC doesn't count `save-prefix` changes as potentially changing the spec (and thus triggering an update-save), then yes.

If what you want is something like making it so npm reuses the `save-prefix` already in package.json whenever it updates or installs, I think that deserves its own standalone RFC, because I'd like to see discussion around how this would apply to `npm install` as well.

Maistho:
Thanks for the feedback @zkat! I'm currently on vacation and will probably not get around to updating this for maybe another week or two when I can sit down for a little while 🌴 

I have some question regarding the terminology used for things, what is this part called in package.json dependencies?
```
// in package.json dependencies
"package": ▶️"^1.0.0"◀️
```
I've been going with `range`, but I'm not sure if that is the correct word.

Thanks again!

zkat:
@Maistho no rush. This is a breaking change, and `npm@7` is probably not going to be released for another couple of months, so waiting a week or two is totally fine. Enjoy your vacation!

As to your question: we usually call that a "spec", or a "saveSpec". From "specifier". They're parsed by [`npm-package-arg`](https://npm.im/npm-package-arg).

accepted/0000-update-package-json.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Yarn
+
+Yarns `upgrade` command does pretty much the same thing. It never edits `package.json` unless a range (or `--latest`) is supplied when updating. IMO, the usability of `yarn upgrade` is better in this regard.
+
+## Unresolved Questions and Bikeshedding
+
+Maybe at the same time add a `--latest` flag, similar to yarns?
+
+If `update` is to change, should other parts of it also change? Are there any other gripes with how the command works currently? Maybe something to do with how `--depth` works, I've considered that to be a bit weird sometimes?
```
felixfbecker: ➡️ https://npm.community/t/impossible-to-update-single-package-without-updating-its-dependencies/1156

felixfbecker:
Highly agree with this RFC! I currently always revert `package.json` after running it and only commit package-lock.json. But I can't expect everyone to remember to do that, and then we end up with unnecessarily restrictive ranges in libraries.

zkat:
We talked and decided to accept this even without an implementation section (and we think we need to reevaluate how that section works, and if it should exist at all). We're hoping to land this change with `npm@7` and think it's very very good. Thank you!

felixfbecker:
Yey, RFC reviews! 🎉 😊

Maistho:
Hi! I'm sorry for not getting back with this (life is hard sometimes), but I'm glad to see it accepted! 👍

PR #17, opened 2018-08-01
[WIP] rfc: allow token exclusion for write 2fa

travi:
i still have work to do to shape this proposal out completely, but thought there is enough here to open up for early feedback.

for reference, this is a continuation of the conversation started [here](https://npm.community/t/allow-subset-of-tokens-to-bypass-2fa-requirement/536)

### DIFF
--- a/accepted/0000-tokens-with-2fa-exclusion.md
+++ b/accepted/0000-tokens-with-2fa-exclusion.md
@@ -1,0 +1,93 @@
+# Token exclusion from publish 2FA
+
+## Summary
+
+{{A concise, one-paragraph description of the change.}}
+
+Allow enabling 2FA for `auth-and-writes`, but with the ability to exclude
+specific token(s) from the requirement to provide a One Time Password (OTP) when
+publishing. This could enable enforcing full 2FA for a user account/package/organization,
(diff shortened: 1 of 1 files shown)


travi:
great to hear. i apologize for not getting back to this before now to flesh it out a bit further, but it sounds like i should hold off for now.

my team is pretty aggressive with automating things and following [continuous deployment of our packages](https://presentations.travi.org/continuous-deployment), so i'd be more than happy to help out the conversation any way i can, especially if our flow helps with walking through concrete examples.

looking forward to see where this ends up. thank you for all your efforts and for the interest in input from the community.

jrjohnson:
It might be going one step too far for this RFC, but I think limiting the non-MFA token to do one specific thing (publish just one package) would make this more secure. 

In my case I would like to automate the publication of a package that is used internally, but would not want to put more popular packages where I have publication rights at risk if the token was exposed. Allowing the creation of a special non-MFA single use publication token would solve this issue, and I think help with several of the contingencies raised by @travi's write up.

Also - It's a promise with no weight, but I would easily be able to justify my organization paying for the private package upgrade if it included this feature. 😄

owlstronaut:
Closing as overtaken

PR #18, opened 2018-08-05
Interactive audit resolver

naugtur:


accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,92 @@
...
+It should be possible to make `npm audit` a step in a CI pipeline to go red every time a new vulnerability affects the project. Managing security of dependencies should be quick to update, effective, easy to audit over time and secure in itself.
+
+For that, the user needs a new tool which makes addressing issues one by one convenient even across a whole ecosystem of projects.
+
+## Detailed Explanation
+
+`npm audit resolve` runs audit and iterates over all actions from the output with a prompt for a decision on each. 
+If fix is available, it's offered as one of the options to proceed. Other options include: ignore, remind later and delete.
+
+All decisions are stored in `audit-resolv.json` file as key-value, where key is `${dependency path}|${advisory id}` and value is a structure specific to the resolution.
```
ljharb: Why resolv instead of resolve?

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,92 @@
...
+
+`npm audit` reads `audit-resolv.json` and respects the resolution (ignores ignored issues, ignores issues postponed to a date in the future)
+
+### resolutions
+- **fix** runs the fix proposed in audit action and marks the issue as fixed in `audit-resolv.json`. On each subsequent run of `npm audit resolve` if the issue comes up again, the user is also warned that the problem was supposed to be fixed already.
+- **ignore** marks the issue as ignored and `npm audit` will no longer fail because of it (but should display a single line warning about the issue being ignored). If a new vulnerability is found in the same dependency, it does not get ignored. If another dependency is installed with the same vulnerability advisory id it is not ignored. If the same package is installed as a dependency to another dependency (different path) it is not ignored.
+- **postpone** marks the issue as postponed to a timestamp 24h in the future. Instead of ignoring an issue permanently just to make a build pass, one can postpone it when in rush, but make it show up as a CI faiure on the next working day.
```
ljharb: The next working day is often more than 24 hours later; is this just meant as an example?

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,92 @@
...
+- Editing a file to specify what to ignore will not suffice any more at that level of detail, so automation is necessary.
+
+**Why audit-resolv.json?**
+Resolutions must be recorded in a file and committed to git(or alternative) for edit history and full audit on who decided what.
+Resolution format must be readable for a JavaScript developer.
+
+- `package-lock.json` is not a good option for storing resolutions because even if we overlook the fact that it'd be adding another purpose to a single-purpose file, the community at large is still used to removing and regenerating it often. 
+- `package.json` is not a good option, because it is already overloaded for so many functionalities. The output of resolutions could be lengthy and clutter the file. User will need the resolution information much less often than they look in the package.json file. ALso, resolution file is not meant to be edited by hand.
+- `.npmrc` is not a good option because it doesn't live in the repository and being a dotfile is easy to miss, so a developer could overlook it affecting the audit. Also using it for the purpose of storing resolutions to project specific issues seems counter-intuitive.
```
ljharb: npmrc *could* live in the repo, it just also might not.

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,92 @@
...
+
+**postpone, remove, investigate**
+Other options are helpers for more elaborate actions a developer would take when confronted with an issue they can't or don't want to fix. 
+
+Why is postpone useful at all? It's designed to build a secure development culture where one didn't yet form. Without it, a developer under time pressure would mark an issue as ignored with intention to un-mark it later. 
+While shipping with a known vulnerability is a bad practice, NPM's mission with the community should be to empower people to build more secure products and trust their skill and understanding of their project's particular needs. We should also aspire to help teams introduce more secure workflows effortlessly, so letting a build pass without risking compromising security long-term is a win. 
+
+Remove is only useful as a convenience. Imagine a developer introducing `npm audit` and having to go through tens of issues. If they notice one of the first issues is caused by a dependency they no longer use, instead of remembering to clean it up later, they can choose this option.
```
ljharb: Presumably this could also narrow the list of action items, if some of them are from transitive deps of the removed item?

accepted/0003-interactive-audit-resolver.md
naugtur: I don't have a strong opinion about this. Wrote it the way it is after some well known files like resolv.conf and didn't give it a second thought. Didn't have to choose between 'resolve' or 'resolved' or 'resolver' but now that I see you ask, I think `audit-resolve.json` is probably a better name.

accepted/0003-interactive-audit-resolver.md
naugtur: Yes. I didn't intend it to count in working days. 

*If noone's at work and the build turns red, does it make a sound?*

accepted/0003-interactive-audit-resolver.md
naugtur: I'm aware of that, but we probably wouldn't want to make an exception and accept global resolutions too. That could lead to hard to debug problems.

accepted/0003-interactive-audit-resolver.md
naugtur: In my experience audit grouped these together nicely, so when I choose remove it's unlikely the dependency or any of its children would come up again.

But we could easily run audit again after the remove command and thanks to audit-resolv being written after each decision, it'd automatically continue from the same point on the list.

accepted/0003-interactive-audit-resolver.md
ljharb: Totally agree; just hoping to clarify the npmrc explanation.

naugtur:
I don't really know what's next other than moving my repo to npm org on GitHub.
There's no disagreement on anything here. 

I have more free time than usual for the next few days so I'd be happy to move this forward too.

evilpacket:
In general I'm very excited about the potential of the `npm audit resolve` feature. I think some work needs to be done to really make it a good interactive experience presenting the information in the best way possible for users.

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
evilpacket: Another though is that the `audit-resolve.json` file should not be published as part of a module publication. We will want to assert exceptions outside of the package.json and update those exceptions outside of the normal publication lifecycle.

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: I thought about a few cases of audit trail with similar conclusions. IMHO the benefit of audit-resolv file's simplicity outweighs the inconvenience of using git history. Also, people are quite familiar with git so the less we do on our end, the less learning they have to do.

I'd like to avoid the file growing to overwhelming lengths, but other than that I'm fine with adding more capability to this file format if it's useful for the server side ignores etc.

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: Will update the RFC to contain the file format spec

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: Leaving only the 24h option it should be annoying enough for most reasonable people to eventually fix. :)
Unless someone gets a job postponing vulnerabilities every morning before the team comes in.

With the audit-resolve format change I just pushed we can also add a count${dependency path}er. I'd consider that a nice touch, not the core functionality though.

zkat:
(Heads up, we _are_ tracking this internally and it's something we intend to get to, but I can't give you an exact timeline right now. Thanks for your patience.)

naugtur:
If you want to consider one bigger push to move this forward, my family is out for winter vacation 25-28Feb while I stay at work, so I'll have unusually many hours in the evenings in that period.

ersel:
This would be an awesome addition to npm CLI. We used to run `npm audit` on CI with [a custom script outlined here](https://medium.com/@ersel_aker/one-simple-trick-to-make-your-javascript-projects-safer-9257983d2efe). 

Recently we've switched to @naugtur 's npm-resolve-audit module by installing it as a dev dependency and adding a script to run `node node_modules/npm-audit-resolver/check.js`

Thanks for compiling this tool @naugtur !

naugtur:
@ersel if you're running it inside an npm script, you can use it as a command `check-audit` and it resolves from dependencies. Otherwise I'd go with `npx -p npm-audit-resolver check-audit` - feels a little cleaner and you don't need to install resolver as a dependency - npx will download and cache it. (npx is shipped with npm so no need to install it)

Check out my slides on using audit in CI (and other things) https://naugtur.pl/pres3/securedev/#/18

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: After some tweaking I updated the RFC to include a precise format definition. Same format definition is used in a branch where I'm working on refactoring.

npm-audit-resolver package after the refactoring should be trivial to integrate into npm audit for the purpose of respecting decisions made in audit-resolve.json

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,92 @@
...
+
+Investigate option has a lot of potential for the future where NPM could do things ranging from linking to resources on mitigation to generating local fixes or providing a marketplace of companies offering services fixing the issue for customers (business potential for NPM).
```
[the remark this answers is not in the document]
naugtur: fixed

naugtur:
I've updated the RFC and I'm working on a refactoring to separate the core features from npm-related and interactive ones.

Could we agree on making `npm audit` use audit-resolve.json being the first step? 
That way people could produce the file manually or with the interactive resolver, but the means of ignoring a package would be there.

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: Update - instead of letting people postpone for more than 24h, due to a popular demand, I got an option in to make ignores expire. So if you decide to ignore something because there's no fix and it doesn't seem accessible from the code that depends on it, it's possible to make it expire.

Added `expiresAt` field to the decision object in resolve file

accepted/0003-interactive-audit-resolver.md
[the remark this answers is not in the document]
naugtur: Schema is defined here: https://github.com/naugtur/npm-audit-resolver/blob/refactoring/src/core/auditFile/versions/v1.js as part of the app - for validation

naugtur:
Hi, 

I'm finishing the refactoring I mentioned before. 
My suggestion to proceed now is that `npm` (and potentially other package managers) can use the core of npm-audit-resolver to include facts from audit-reslve.json when checking the repository with `npm audit` command - therefore skipping issues that have been addressed before.
The interactive part can (initially or forever) remain in the separate package with potential competitors, while the spec for audit-resolve.json becomes something we can slowly standardize.

I will open another PR to npm some time soon introducing the line to include decisions from audit-resolve.json in the `npm audit` logic.

naugtur:
I couldn't join the RFC meetings, because I tend to give my kid a bath at that time and it's not optional ;) I'll try next time.

@ruyadorno  @darcyclarke Let's continue this.

ruyadorno:
Hi @naugtur picking up from where the conversation ended last time: https://github.com/npm/cli/issues/525#issuecomment-577907157

We would love to start working on `npm audit resolve` now that v7 is taking shape 😊 

That said, do you want to update this RFC to reflect the changes we mentioned in that issue? (specially the part about not making it an interactive command anymore) It's up to you to decide if you want to bring this RFC across the finish line - I realize it's been ages since you first opened it and totally understand if you're not interested in following up anymore 😁 We def appreciate all the work that you've put into this RFC and `npm-audit-resolver` ❤️

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,165 @@
...
+It should be possible to make `npm audit` a step in a CI pipeline to go red every time a new vulnerability affects the project. Managing security of dependencies should be quick to update, effective, easy to audit over time and secure in itself.
+
+For that, the user needs a new tool which makes addressing issues one by one convenient even across a whole ecosystem of projects.
+
+## Detailed Explanation
+
+`npm audit resolve` runs audit and iterates over all actions from the output with a prompt for a decision on each. 
+If fix is available, it's offered as one of the options to proceed. Other options include: ignore, remind later and delete.
+
+All decisions are stored in `audit-resolve.json` file as key-value, where key is `${advisory id}|${dependency path}` and value is:
```
ruyadorno, line 26: @isaacs I'm under the impression that the `${dependencyPath}` bit here is very non-arborist and not necessary overall in the context of an audit triage system, would like to hear your thoughts on it

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,165 @@
...
+
+Paid alternatives for managing security of a node.js project are available, including Snyk, with focus on providing patches to their customers.
+
+## Implementation
+
+reference implementation https://github.com/npm/cli/pull/10
+
+prototype of a working tool (in production use) https://www.npmjs.com/package/npm-audit-resolver
+
+The implementation is, and should remain, runnable standalone as a separate package with minor wrapping code - useful for testing new features without bundling unfinished work with npm cli versions and therefore node.js
```
ruyadorno, line 88: It would be nice to describe a bit more the implementation over here, an example of the cli UX would be very welcomed, maybe just copy or start from the example I floated over https://github.com/npm/cli/issues/525#issuecomment-573928594


```
$ npm audit resolve list
1325 [high]     handlebars: Prototype Pollution
1324 [high]     handlebars: Arbitrary Code Execution
788  [moderate] js-yaml: Denial of Service

$ npm audit resolve get 788
Status: Pending
Level: Moderate
Type: Denial of Service
Package name: js-yaml
Dependency of: tap [dev]
Path: tap > tap-parser > js-yaml
More info: https://npmjs.com/advisories/788

$ npm audit resolve set 788 --status=Read

$ npm audit resolve list
1325 [high]     handlebars: Prototype Pollution
1324 [high]     handlebars: Arbitrary Code Execution

$ npm audit resolve get 788
Status: Read
Level: Moderate
Type: Denial of Service
Package name: js-yaml
Dependency of: tap [dev]
Path: tap > tap-parser > js-yaml
More info: https://npmjs.com/advisories/788
```

accepted/0003-interactive-audit-resolver.md
naugtur: I'm not a native speaker and I'm not familiar enough with the culture to grasp what you mean exactly when calling something non-arborist. I'm going to focus on the "~necessary" part.

Mixing advisory and dependency path as a key in the audit-resolve file is the core of the whole idea.

It's the first time I was able to safely ignore vulnerabilities I knew won't affect me.

_A story:
My adventure with checking dependencies started with adding `nsp check` to our projects. 
The number of reports in software we used was low, so I didn't know what I'm getting myself into when I made `nsp check` a step in CI.
REDoS vulnerabilities got very popular after that and it's been an annoyance. nsp allowed creating an ignore list with just the advisoty id, so after finding out that a 5th level dependency of an old testing tool we use (or something) is unlikely to ever be fixed, we had to ignore all REDoS cases in all 20+ projects. 
Then `npm audit` showed up so I switched. It had no means of ignoring anything. CI step was removed and I started working on a wrapper that ended up being the npm-audit-resolver._

The key features I value in this solution:
- if I ignore the vulnerability in one package, it won't transfer to another
- if I ignore a vulnerability in a package coming in as a dependency of a dependency I don't really invoke in the app, I'm not automatically ignoring it elsewhere.
- if a new instance of the vulnerability shows up in dependencies of something else or in a dep tree of newly installed dependency - it doesn't get ignored 

While ignoring vulnerabilities seems like a thing that doesn't improve security when you hear about it, in this case it's a tool to really improve visibility of the reports you should actually look into and **increase** security. 

Ignoring vulnerability report is not here for convenience, but to make `npm audit` a viable step in the CI pipeline.

accepted/0003-interactive-audit-resolver.md
naugtur: I believe I found a different way to tackle this.

I think the first step is for `npm audit` to start supporting the `audit-resolve.json` file and leave creating it to external tools with various options for UI.
I'm leaning towards limiting this RFC to just that functionality and then as audit-resolve file standard is accepted, propose a way to edit it.

I've got a few ideas for the resolve command without interactions, but npm-audit-resolver is more comfortable to use than these. I wouldn't bother to implement it as interactive if I didn't need it ;)

Some ideas for non-interactive that I could develop into more detailed RFCs:
- audit resolve runs fix and marks as fixed, puts the rest in and marks as 'none' so user can edit
- audit resolve lists unresolved items and lets the user resolve all items in a certain way with one command where the filters are advisory_id, package, dev
- an almost interactive one where the --continue flag and alike are used like with git rebasing

feels like we need to chat less asynchronously too ;)

accepted/0003-interactive-audit-resolver.md
ruyadorno: ah 😄 sorry about that, I meant https://github.com/npm/arborist

accepted/0003-interactive-audit-resolver.md
ruyadorno: for sure it would be nice to catch up, I tried reaching out on twitter but let me know if there's a better place to start a conversation 😊

naugtur: I'm not seeing anything on twitter, let's switch to email for scheduling - naugtur@gmail.com
I'll explain the motivations behind what resolver is now and then we can brainstorm what to do about the UX

ruyadorno: right, twitter sucks these days... it used to notify people whenever someone starts following them

but no worries, I’ll shoot you an email them 😊

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,165 @@
...
+
+### audit-resolve.json
```
bnb, line 90: It would - to me - be extremely preferable to have a less scoped file. Whether this ends up being `package.json` or `audit.json`, I'd very strongly prefer to avoid adding one-off files. Personally, I'd prefer to be able to just have this in my `package.json` since that's a file I already actively maintain in an `audit` property but I can see the potential value in splitting this out into an `audit.json` that could also define additional audit things for the future.

naugtur: There's a section with the analysis of why the file is separate from package.json etc.  
As for `audit-resolve` vs `audit` I don't really mind and I also don't see much of a difference, but I get the sentiment. 

The tooling I built supports extending the file over time and there's already a not-yet-defined section 'rules' next to 'decisions' I introduced to the spec in the package and plan to use to automate some decision making. 

I strongly believe there's nothing blocking this file from getting used in the extent it's described here in the initial implementation and then growing in scope as we figure out more.

Den-dp:
https://www.npmjs.com/package/audit-ci is another pretty popular alternative solution for this problem. Dropping it here to give more info for inspiration.

naugtur:
Thanks for bringing it up @Den-dp 
The important difference here is audit-ci is only operating on risk levels or advisory allowlist. 

audit-resolve.json file is encoding precisely which items were ignored so if you ignore an unreachable ReDos in a sub dependency somewhere in your web app, you still get a notification if ReDos is found in your server framework used.

audit-resolve-core package (work in progress) is meant for embedding and audit-ci could easily switch to using that at any point btw.

naugtur:
Thanks for mentioning compression. Although the goal is to be very explicit not to invite future occurrences, you got me thinking. The list can be compressed by putting a wildcard in the path. So each occurrence of a package with certain advisory anywhere in dependency tree of one specific direct dependency could be collapsed into one.

That would limit the number of entities while not breaking the security promises.

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,155 @@
...
+
+```js
+{
+  "version": 1,
+  "decisions": {
+    "ADVISORY_NUMBER|DEPENDENCY_PATH":{
+      "decision": "RESOLUTION_TYPE",
+      "reason": "Reason provided by the person making the decision (optional)",
+      "madeAt": timestamp
+      "expiresAt": timestamp
```
dominykas, line 114: Can this please also (only?) support a human readable ISO date string?

Use case: extending the expiry time without using any tooling (an extremely common operation, from experience with `.snyk` policy files...)

naugtur: Personally I believe it should not be limited to ISO string because of the ending I always mess up. 
2020-09-21 should also work if we want to cater to humans :)


Next iteration of the core package that reads and writes the resolve file will have that.

Thanks for the data point!

nierob:
> Thanks for mentioning compression. Although the goal is to be very explicit not to invite future occurrences, you got me thinking. The list can be compressed by putting a wildcard in the path. So each occurrence of a package with certain advisory anywhere in dependency tree of one specific direct dependency could be collapsed into one.
> 
> That would limit the number of entities while not breaking the security promises.

It depends what kind of use case you want to solve. For _me_, partially ignoring issues doesn't make sense because it would still block CI. That means that path as such is useful when deciding if an issue can be ignored, later on it is not needed. That is because the final decision is binary; "do I accept this package or not". So instead of path I would use package version:
```js
 {
   "version": 1,
   "decisions": {
     "ADVISORY_NUMBER": [
        {
           "AFFECTED_PACKAGE_NAME_AND_SHA1_FROM_LOCK_FILE":{
             "decision": "RESOLUTION_TYPE",
...
```
If one update package version, technically the security analysis needs to be re-done, from scratch. If you put path wildcard, this situation will likely not be detected.

naugtur:
@nierob that doesn't cover when a package from your dev deps gets ignored because the vulnerable cod is obviously unreachable, but then it resurfaces as a prod dependency pulled by something else and you're busted.

On the app level (the resolver app that asks you what to do) the decision can be aggregated, but the stuff written in audit-resolve.json needs to be a snapshot of where the package was found at the time you decided to ignore it. 

And it's not affected by npm's deduplication, that happens on a differnt leve, so the entries in package-lock are not moved around.

nierob:
> @nierob that doesn't cover when a package from your dev deps gets ignored because the vulnerable cod is obviously unreachable, but then it resurfaces as a prod dependency pulled by something else and you're busted.

True. With the current config definition, the opposite is also true. One can update package version and previously unreachable code path may became reachable. 

You convinced me that it needs to contain all three; path, package version and vulnerability id :-) In such case no much can be done regarding the file size. Maybe path prefixes can be used... Still I would discuss that as a format update, in a different RFC, it looks like a side-track.

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Summary
+
+Historically this proposal was titled "Interactive audit resolver". That idea was too wide to introduce in one go and parts of it are better served in userland packages. This is now the core subset of the original proposal containing support for resolutions file.
+
+This proposal is for adding means for a human to resolve issues from `npm audit` by making and documenting decisions to ignore particular false positives. 
+The implementation should introduce support for reading and applying decisions from `audit-resolve.json` file
+
+Features:
+- Let users save tecisions about vulnerabilities and change `npm audit` behavior to accomodate that.
```
nfriedly, line 11: Was "tecisions" supposed to say "decisions" ?

naugtur: That's definitely one of the possibilities.

jfaylon:
Please allow this this for the open source libraries to have time to fix the vulnerabilities. We're using `npm audit` for our systems for security and 1 security advisory can block the entire deployment for us. If we have the ignore for x number of days/hours to give ample time for the libraries/dependencies to update their libraries, that would be great.

naugtur:
@jfaylon There's a collaboration space being set up under the OpenJS Foundation to tackle this.

For now you can use https://www.npmjs.com/package/npm-audit-resolver for your vulnerability management. My team's been doing that for a few years now.

accepted/0003-interactive-audit-resolver.md
```diff
@@ -0,0 +1,155 @@
+# Audit resolutions
+
+## Summary
+
+Historically this proposal was titled "Interactive audit resolver". That idea was too wide to introduce in one go and parts of it are better served in userland packages. This is now the core subset of the original proposal containing support for resolutions file.
+
+This proposal is for adding means for a human to resolve issues from `npm audit` by making and documenting decisions to ignore particular false positives. 
+The implementation should introduce support for reading and applying decisions from `audit-resolve.json` file
```
bnb, line 8: I would prefer either doing this through a more generic `audit.json` or through an `audit` property in package.json. Having a whole new file for a single purpose in the npm CLI feels… off.

There was discussion of additional config setups that I had with @darcyclarke that it could also potentially fit in, but they don’t exist yet so I can’t really recommend them.

naugtur: I didn't go with a package.json field because everyone kept saying we use it for too many things already and the audit file tends to be larger (from my practical experience of running the tool for 3-4 years)
Also, there's a plan to use it as vehicle for sharing recommendation. More on this here:
https://dev.to/naugtur/do-you-need-help-with-your-npm-audit-3olf

I'm not opinionated about the name. The final decision on the name will have to happen when we standardize the format in OpenJSF

ljharb: Other advantages of keeping it in a separate file include more granular codeowners targeting.

bnb: I'll make it explicit here so it can be referred to in whatever OpenJSF work you're talking about: I'm a hard -1 on `audit-resolve.json`.

naugtur: audit.json is fine. 
OpenJSF work here 
https://github.com/openjs-foundation/pkg-vuln-collab-space

ljharb: @bnb on that file name, or on a separate file outside of package.json?

bnb: @ljharb the specific file name.

Honestly, to me, the most preferable thing would be to support **both** `package.json` and an independent file to enable the widest possible set of end-users adopting this.

ljharb: I think having it in package.json, but ignored by package consumers, would be actively harmful (ie, be very surprising). package.json should be for things related to a package, which is usually published.

isiahmeadows: @ljharb I'd say it'd be useful to check recursive packages for their own resolutions, but that's not exactly good security practice - users should actually be doing their due diligence here with their dependencies and that...doesn't exactly help them do that. (At best, it makes it far harder for them to audit dependencies accurately, and at worst users just blindly trust their dependencies' maintainers' decisions without even seeing what those decisions even were.)

So yeah, it shouldn't count as package metadata - it's more like *project* metadata than *package* metadata.

naugtur: https://dev.to/naugtur/do-you-need-help-with-your-npm-audit-3olf

karlhorky: > that's not exactly good security practice

> users should actually be doing their due diligence here with their dependencies and that

I think it should be fully considered that a large number of projects will either:

- not have maintainers that have the bandwidth to do this
- not have maintainers that know how to do this
- not have maintainers that know this needs to be done at all
- not have maintainers (unmaintained / insufficiently maintained)

I work with a bunch of beginners every three months in our coding bootcamp at @upleveled and this is one example of maintainers who will not do this. And if you pick random packages on npm, more likely than not, you'll have something that is not well maintained.

So to save and check the audit decision as close as possible to the actual usage of the vulnerable code seems like actually a **better** security practice overall for the industry (at least as a fallback option).

Affecting sea change so that everyone in the industry spends time to check the security of their dependencies is unrealistic.

I think @naugtur's suggestion here could be one interesting way to implement this:

> package maintainers could publish ignore lists for the dependencies of their package when they did the research to verify the package is not affected.
>
> Such lists could be aggregated into recommendations maintained by security professionals.

To be clear, again, I'm not proposing this as the ONLY security measure - teams which have security processes, budget and practice in place should continue to do that and thus will be even more secure.

But for the majority of projects which will not have this level of focus on security, there should be a base level of security by default, provided by the ecosystem.

naugtur: The idea is to give people tools to do the best they can with the bandwidth they have. If we force them to address everything on their own, they're likely to just drop security from their pipeline

isiahmeadows: And I'm just saying we should just not hide it by default, and force the user to take some explicit action, even if it's a compromise between security and time.

ljharb: We shouldn't - the user should never see something that isn't actionable for them.

isiahmeadows: I'll agree to disagree and move on.

bnb:
> [edit] This RFC is very old, some progress on developing the idea led to a Collab Space in OpenJSF and this format proposal: openjs-foundation/pkg-vuln-collab-space#11

if this RFC is not reflective of a proposal, should it be closed? It's already been extremely misleading given that it's not the "real" representation of what the proposal is.

naugtur:
> if this RFC is not reflective of a proposal, should it be closed? It's already been extremely misleading given that it's not the "real" representation of what the proposal is.

After the last meeting I knew I should rewrite it now and I know what to put in it. Family matters and chores got in the way. I will rewrite it to reflect what it should.

rudivanbiljon:
Is there any progress updates on this? With the new move of the advisory database of NPM now on GitHub (https://github.blog/changelog/2021-10-03-the-npm-advisory-database-is-now-part-of-the-github-advisory-database/) some of our moderate vulnerabilities that we can't resolve right now have been tagged as High - breaking our builds. 

It would be great if we can accept the risks on them specifically to allow the build to pass again.

trevyn:
Looks like it didn't actually get discussed at the meeting: https://youtu.be/nkohPHyvzw4?t=3582

@bnb Sounds like this was an action item for you; did you get a chance to look at the state of this?

bnb:
> @bnb Sounds like this was an action item for you; did you get a chance to look at the state of this?

I'm not going to try to be speculative on an RFC that the creator has very specific ideas about without it actually being in a PR. I don't think that'd be particularly fair to them or me 😅

trevyn:
Ok, so we're waiting on @naugtur , got it! :)

devdev-dev:
Waiting for this like many others...

bertolo1988:
Still waiting for this.

naugtur:
BTW. I'm very near releasing a new major version of npm-audit-resolver, if anyone wants to help out.

Happy to go back to discussing getting the policy check into npm cli. Still bullish on not doing it per package name but per dependency path.

bertolo1988:
> if anyone wants to help out.

I can help, you can reach me using the contacts in my Github profile.

hax:
Any update? We need it to bypass some common packages in the third-party dependencies (eg. https://github.com/lodash/lodash/issues/5808 ) which block our build steps.

naugtur:
You can use npm-audit-resolver - I've released a version with updated support last year.

This RFC is unlikely to move forward. npm is maintained by a small team and unlikely to make major new work happen.

axi92:
For others that stumble across this. I stopped using npm audit on CI and instead use trivy and auditjs and review every CVE and whitelist when I see that I don't use that code part or its a nonissue for my app.

PR #20, opened 2018-08-20
powershell scripts for installed binaries

bterlson:
Let's get rid of "Terminate batch job (Y/N)" for Windows users using powershell!

### DIFF
--- a/accepted/0000-powershell-bins.md
+++ b/accepted/0000-powershell-bins.md
@@ -1,0 +1,55 @@
+# Powershell for Installed Binaries
+
+AKA Terminate Terminate Batch Job
+
+## Summary
+
+Today, installing binaries using npm creates an executable bash script and a .cmd file with the same name. Unfortunately, terminating cmd scripts with SIGINT causes a useless prompt for windows users - "Terminate batch job (Y/N)". Also, Powershell is the default shell for Windows. By creating .ps1 scripts in addition to bash and .cmd, powershell users will have a better command line experience.
+
+## Motivation
+
+Windows users have three main options when it comes to shells:
+
+- cmd: essentially unchanged from early versions of Windows
+- powershell: the default Windows shell, under active development
+- unixy shell: for example git-bash, zsh via WSL
+
+WSL is growing in popularity, but isn't super common. Many Windows users need to use or prefer using a native shell. And for those using Powershell, we can make their experience better by creating powershell binary wrappers.
+
+## Detailed Explanation
+
+When installing binaries (e.g. `npm install -g typescript`), NPM should create `tsc.ps1` in addition to `tsc.cmd` and the executable `tsc` bash script. Powershell will prioritize `foo.ps1` over `foo.cmd` when users type `foo` on the command line. cmd users will have to continue living with it. Users of git-bash or WSL shells will be unaffected.
+
+## Rationale and Alternatives
+
+The main alternative is to maintain the status quo, but this leaves an annoying SIGINT prompt in Windows's default shell.
+
+Creating just .ps1 scripts isn't an option. cmd can't start treating .ps1 scripts as executables (at least not by default), and NPM can't drop support for cmd.
+
+## Implementation
+
+Here's an example of a functionally identical powershell wrapper:
+
+```powershell
+$myPath = Split-Path $myInvocation.MyCommand.Path -Parent
+$node = Join-Path $myPath "node.exe"
+$binPath = Join-Path $myPath "[ path to node_module bin script, e.g. node_modules\typescript\bin\tsc]"
+if (Test-Path $node)
+{
+    & $node $binPath $args
+}
+else
+{
+    node $binPath $args
+}
+```
+
+The path to the bin script will have to be substituted into the right place.
+
+## Prior Art
+
+- [cmd-shim](https://www.npmjs.com/package/cmd-shim)
+
+## Unresolved Questions and Bikeshedding
+
+- impact of additional scripts on install times, node_modules size, etc.


accepted/0000-powershell-bins.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Motivation
+
+Windows users have three main options when it comes to shells:
+
+- cmd: essentially unchanged from early versions of Windows
+- powershell: the default Windows shell, under active development
+- unixy shell: for example git-bash, zsh via WSL
+
+WSL is growing in popularity, but isn't super common. Many Windows users need to use or prefer using a native shell. And for those using Powershell, we can make their experience better by creating powershell binary wrappers.
```
iarna: WSL is also Unix as far as npm is concerned, and so these shims are not created with it. It's lovely, but I agree, not really Windows support. =)

accepted/0000-powershell-bins.md
```diff
@@ -0,0 +1,55 @@
...
+Creating just .ps1 scripts isn't an option. cmd can't start treating .ps1 scripts as executables (at least not by default), and NPM can't drop support for cmd.
+
+## Implementation
+
+Here's an example of a functionally identical powershell wrapper:
+
+```powershell
+$myPath = Split-Path $myInvocation.MyCommand.Path -Parent
+$node = Join-Path $myPath "node.exe"
+$binPath = Join-Path $myPath "[ path to node_module bin script, e.g. node_modules\typescript\bin\tsc]"
```
dahlbyk: For a generated script, this should use single quotes.

kumarharsh:
@bterson nice job. You can also add that MS (or atleast some of their top employees) itself recommends we use ps1 scripts. See https://github.com/Microsoft/console/issues/217#issuecomment-404268654

accepted/0000-powershell-bins.md
```diff
@@ -0,0 +1,56 @@
...
+The main alternative is to maintain the status quo, but this leaves an annoying SIGINT prompt in Windows's default shell.
+
+Creating just .ps1 scripts isn't an option. cmd can't start treating .ps1 scripts as executables (at least not by default), and NPM can't drop support for cmd.
+
+## Implementation
+
+Here's an example of a functionally identical powershell wrapper:
+
+```powershell
+$myPath = Split-Path $myInvocation.MyCommand.Path -Parent
```
felixfbecker: Is there a difference between this and `$PSScriptRoot`?

joeyaiello: PM for PowerShell here, @bterlson reached out to me on this one. 

Quoting [this](https://stackoverflow.com/questions/45615239/functional-differences-between-psscriptroot-and-myinvocation) SO article, if you care about support on Windows 7 or Server 2008 R2 where PowerShell has not been updated past the inbox PowerShell 2.0, you should stick with `$myInvocation`:

bterlson: Thanks @joeyaiello! So IMO it's probably better to err on the side of caution and go with the more compatible `$myInvocation`.

accepted/0000-powershell-bins.md
```diff
@@ -54,3 +54,4 @@ The path to the bin script will have to be substituted into the right place.
 ## Unresolved Questions and Bikeshedding
 
 - impact of additional scripts on install times, node_modules size, etc.
+- default powershell permissions: will powershell users need to toggle some settings so this works without permissions errors? Can the errors be worked around? (This PR will need testing on a fresh Windows install)
```
dahlbyk: c.f. https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-6

Short answer: it depends on the version of Windows. Newer installs default to `RemoteSigned`, which should be fine.

What should happen if the user's policy won't allow running scripts? Maybe don't create the `ps1` scripts on `install`?

dahlbyk: I may have spoken too soon. Seems like newer Windows Server defaults to `RemoteSigned`, but Windows 10 is still `Restricted`.

My tentative recommendation would be to shell out with something like this:

```powershell
powershell -NoProfile -Command Get-ExecutionPolicy
```

If it returns `Unrestricted` or `RemoteSigned`, `ps1` shims should work fine.

bterlson:
Roger that, thanks @zkat! I plan to test with a fresh install of Windows. FWIW, I believe it's going to work because scripts created from a local process (i.e. node.exe) should be considered local scripts rather than remote scripts. I'm unfortunately headed out for vacation as we speak, so if someone is super amped about this and wants to test this, I'd be happy! Otherwise, I'll be back in September.

accepted/0000-powershell-bins.md
felixfbecker: I think it would be great if we could still find some solution that improves the experience for all those who have the execution policy configured - it's one of those first things you do as a dev when you setup a Windows machine, and many tools (e.g. Chocolatey package manager) don't work without it.

One thing to consider when shelling out is that PowerShell Core (v6) uses `pwsh` as the executable name. I don't know if they share the same execution policy, maybe? I guess it's stored in the registry somewhere.

be5invis:
However the more reliable way should be, like, emitting an EXE?

be5invis:
@bterlson 
I digged into my `PATHEXT` and saw this...

```
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL
```

... and I saw these

```
WSF;WSH
```

These are WSH's batch files and runs blocks of VBScript/JScript (the very old version, stopped at ES3...), and it existed in Windows for over 20 years. I think they could be a very interesting solution for this particular problem.

kumarharsh:
Using such an ancient tech which very few people even remember how to program might be counter-productive to our goals here, no?

be5invis:
@kumarharsh But, isn't CMD more ancient?

kumarharsh:
Yeah, but the issue is about migrating from CMD to something more modern... migrating to JScript would not be ideal.

be5invis:
@kumarharsh 
The best way would be generating EXEs...
Actually you **CAN** modify an EXE, **keeping it digitally signed**, adding custom data at the end of the executable file, to do the job of CMDs.
Though bad antivirus software may consider Node as a virus...

https://blog.barthe.ph/2009/02/22/change-signed-executable/

kumarharsh:
The technique sound great, apart from the virus caveat. Does Windows Defender catch these types of modifications to an EXE?

be5invis:
@kumarharsh Not tried. Maybe I can try them in the incoming holiday.

dahlbyk:
> Though bad antivirus software may consider Node as a virus...

IMHO the Windows antivirus world is still a disaster, and should absolutely preclude any notion of shipping a `.exe` for a shim.

be5invis:
@dahlbyk But if the emitted EXE is digitally signed, the situation could be different.
We may need to test them on Windows PC with different AV softwares installed.

zkat:
Hey @bterlson! I assume you're going on another vacation very soon (lol), but when you're able, can you get around to test the open questions so we can get a better idea about what we'll be dealing with in re execpolicy and how it should be addressed?

isaacs:
Is it safe to move this to "implemented" since cmd-shim landed https://github.com/npm/cmd-shim/pull/34 in cmd-shim@3.0.0, which is set for release in npm 6.11.0?

PR #21, opened 2018-08-31
Shallow updates

felixfbecker:
An RFC for https://npm.community/t/impossible-to-update-single-package-without-updating-its-dependencies/1156/3


accepted/0003-shallow-updates.md
```diff
@@ -0,0 +1,90 @@
...
+are _not_ satisfied, npm should update the indirect dependencies that _need_ to
+be updated too to satisfy all constraints.
+
+In other words, as a user I would expect `npm` to update `package-name` to the
+latest in-range version, then check if the constraints for all direct
+dependencies of `package-name` are still satisfied. If they are not for a
+package, run the same steps recursively for that package (do shallow update,
+check sub-dependencies).
+
+If and only if a conflict occurs, duplicate the package the package as npm does
```
nicksnyder: typo: `the package` is repeated

nicksnyder: I don't fully understand "deduping" in the context of NPM, so this is a bit dense for me to be able to know if I agree or disagree. An example might help. Or maybe I am not the intended audience for this document.

accepted/0003-shallow-updates.md
```diff
@@ -0,0 +1,90 @@
...
+notice, because lockfile changes are usually not reviewed (they are hidden by
+GitHub in diffs by default).
+
+[Here](https://about.sourcegraph.com/blog/the-pain-that-minimal-version-selection-solves#background)
+is a concrete case where we accidentally brought in a broken indirect dependency
+in what should have been a low-risk shallow update.
+
+## Detailed Explanation
+
+When `npm update package-name` is run, only `package-name` should be updated.
```
nicksnyder: This describes exactly what I would want/expect as a user.

accepted/0003-shallow-updates.md
felixfbecker: If you have a dependency constraint tree like this:
```
|- A@^1.0.0
| `- B@^1.0.0
`- B@^1.2.0
```
npm will only install B once ("deduplicate"/"dedupe" the package), because all constraints for `B` are compatible. So on disk `node_modules` actually looks like this:
```
|- A@1.0.0
`- B@1.2.0
```
But if `A` now bumps its constraint to `B@^1.3.0`, npm needs to update `B`. So it could either (1) maintain the deduping and update the root dependency too:
```
|- A@1.0.0
`- B@1.3.0
```
or it could (2) duplicate the package to keep the root locked:
```
|- A@1.0.0
| `- B@1.3.0
`- B@1.2.0
```
I am advocating for (1).

nicksnyder: Thanks for the clear example! I think it would be useful to inline in the rfc.

(1) makes the most sense to me because that is consistent with what npm install would do. Single package updates don't need to be a special case with respect to deduping (as I understand it).

AnyhowStep:
Having "shallow" updates would be perfect. It grates on my nerves that I update *one* package and it causes a million other packages to be updated.

And that one package didn't even update any dependencies!

accepted/0003-shallow-updates.md
```diff
@@ -0,0 +1,90 @@
...
+This causes real pains when having a frequently-updated shared library on npm
+that is used in multiple applications, because every time you want to update
+just that shared library in an application (for example for a small but
+important low-risk bug fix), it also has the potential to bring in a pile of
+bugs introduced in new versions of indirect dependencies.
+
+As a user, this behavior is counter-intuitive, because when asking for a
+specific package to update, I expect npm to only update that package (unless it
+_has_ to update indirect dependencies to satisfy new constraints) and otherwise
+respect the versions in the lockfile.
```
bains5: Thanks

PR #22, opened 2018-10-20
Add feature to show dependencies of a particular dependency

alanyinjs:


zkat:
Hey, touching base here. Are you still interested in this RFC?

alanyinjs:
@zkat Yes I'm still interested! Just trying to come up with an alternative...

mikemimik:
The root problem that you're trying to solve for is to filter the view of the dependencies in your project. This is something we very much want as a feature in `npm/cli` project. However, we don't be able to accomplish this until `npm@7` due to some performance, and timeline restrictions. We're definitely going to revisit this once we have the beta of `npm@7` out in the publics hands. Thanks so much @alanyinjs for creating this RFC. We'll circle back in Q1 of 2020 on this to get it implemented in `npm@7` 👍

PR #24, opened 2018-12-06
Unpublished modules should return 410 Gone

markstos:
Unpublished packages currently cause pain for NPM users. A recommended approach
ends up being to [delete your lock file and re-generate it](https://github.com/yarnpkg/yarn/issues/6702#issuecomment-442314783),
which can cause a huge amount of dependencies to change when only one was needed.

Unpublished packages currently return the HTTP status code "404 Not Found".
This code is designed for resources that might exist again in the future. In
the NPM ecosystem, version numbers are wisely
[immutable](https://www.npmjs.com/policies/unpublish). A better HTTP status
code to return for an unpublished package is "410 Gone", designed for resources
that will never come back.

When `npm` encounters a [410
Gone](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) resource,
it could try to resolve the related semantic version again, possibly selecting a newer
version, fixing the issue the single dependency without requiring regenerating the entire lock file.

### DIFF
--- a/accepted/0000-410-gone-for-unpublished.md
+++ b/accepted/0000-410-gone-for-unpublished.md
@@ -1,0 +1,108 @@
+# Unpublished modules should return 410 Gone 
+
+# Summary
+
+Unpublished packages currently cause pain for NPM users. A recommended approach
+ends up being to [delete your lock file and re-generate it](https://github.com/yarnpkg/yarn/issues/6702#issuecomment-442314783),
+which can cause a huge amount of dependencies to change when only one was needed.
+
+Unpublished packages currently return the HTTP status code "404 Not Found".
+This code is designed for resources that might exist again in the future. In
+the NPM ecosystem, version numbers are wisely
+[immutable](https://www.npmjs.com/policies/unpublish). A better HTTP status
+code to return for an unpublished package is "410 Gone", designed for resources
+that will never come back.
+
+When `npm` encounters a [410
+Gone](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) resource,
+it could try to resolve the related semantic version again, possibly selecting a newer
+version, fixing the issue the single dependency without requiring regenerating the entire lock file.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


zkat:
Brief update: We've moved this internally to appropriate channels and we'll be doing an internal review + prioritization of this feature. No timeline until that's done.

mikemimik:
There is some internal work happening on the registry around unpublishing packages. To that effect, there is some changes happening around the status codes as well. There is an action item from https://github.com/npm/rfcs/issues/70 that will follow up with that feature team to make sure this gets added or at least addressed.

Ethan-Arrowood:
In the similar aspect - if a package is deprecated in place of another package could we use a [redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections) code (301 or 308)? It could definitely be a security issue if we allow users to redirect their packages, so maybe instead it errors but also provides the recommended alternative library?

markstos:
I would issue a HTTP Redirect response sparingly-- only in a case where a module was found to have a security flaw and there's a known replacement which works exactly the same but lacks the security flaw.

isaacs:
Just to clarify the discussion between @seldo and me regarding url reuse, that was about unpublishing _all_ versions of a package.  Ie, if you do `npm unpublish -f foo`, then the request to `https://registry.npmjs.org/foo` should return a 404, not a 410, because someone else can publish new versions of foo (though, importantly, then _cannot_ reuse _old_ version numbers, and thus cannot take over the same tarball urls - those are gone for good.)

Responding with a 410 response code to _tarball_ urls for deleted package versions is entirely appropriate, and I don't see any reason to oppose the fallback logic described in this RFC, but may have some feedback when we start trying to implement it.

The first step, of course, is to see if we can get the registry to respond with a 410 to deleted tarball requests,

markstos:
Good clarification @isaacs. I've updated the RFC to clarify that the scope is restricted to tarball URLs and include your note that completely removing a package is a different case.

PR #26, opened 2018-12-17
Publish older versions without a tag

watson:


ljharb:
This would be great; i currently use https://npmjs.com/safe-publish-latest to prevent accidentally publishing a backport as “latest”.

Even better (on top of this) would be if the tag could be overwritten in prepublish - then i could have a package that removed the “latest” tag when it was unsafe.

ljharb:
@evocateur that’s a great workaround, but it still ends up creating a tag that i have to remove after publishing.

evocateur:
@ljharb Oh, fair enough. At least it wouldn't be "latest" by default in the scenario I described. Still doesn't avoid a tag altogether, you are correct.

ljharb:
@zkat couldn’t something like `--no-tag` be added as a semver-minor, and the breaking change later be making that the default?

ljharb:
Oh sure, maybe `--without-tag` then - bikeshed aside it seems totally doable as a semver-minor

PR #27, opened 2018-12-21
Package overrides

aeschright:


accepted/0000-package-overrides.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Rationale and Alternatives
+
+Users don't frequently need to do this sort of override because npm's nested tree structure allows for a collection of dependencies to require incompatible semvers for the same package. In many cases, the default behavior is acceptable and nothing further is needed. In other, more problematic situations, the user may encounter conflicting binaries or other components that cannot be used simultaneously. The only current options are to attempt to get the maintainers of the dependencies to make changes that will reduce these conflicts, or to use a tool like `https://www.npmjs.com/package/replace-deep-dep` to repair the requirements.
+
+## Implementation
+
+In order to do this, the npm installer will need to read in an overrides section from `package.json`. This will have a simple format such as `overrides: { 'foo': '1.2.3' }`. These values will need to be used at the end of the tree-building process to select the final version. (Is there also a pruning element to this?)
```
ljharb: What if i want to override one package with another - like underscore for lodash, etc?

It would be ideal if the RHS of this could also be a full `npm install` argument - ie, package name/url + @ + version.

zkat: You would use package aliases, which are coming soon, so: `"undescore": "npm:lodash"`

ljharb: Awesome! How soon?

zkat: the same soon as overrides

zkat:
@mkay581 Overrides is solving a different problem! We do have a [Singleton Package RFC](https://github.com/npm/rfcs/pull/23) for the "larger" problem you mention -- but overrides can be used for things like updating deep dependencies with security-patched forks, or different versions with important patches (or preventing a semver update from updating to a broken version). These are some of the examples, but ultimately, this is orthogonal to singletons, which will _benefit_ from this, but this is not the primary purpose at all.

PR #28, opened 2018-12-30
Multi target proposal

dead-claudia:
[Rendered](https://github.com/isiahmeadows/rfcs/blob/multi-target-proposal/accepted/0000-multi-target-support.md) | [Discussion](https://npm.community/t/support-multiple-targets-including-platform-and-architecture-specific-targets/4318)


zkat:
(heads up, this will probably sit here for a while until we get around to scheduling that work -- it's not on our current roadmap. Thanks for your patience while we get to this. We are definitely tracking this as a task we need to do.)

PR #31, opened 2019-01-11
chore: consistently reference @babel/core

SimenB:
Ref https://github.com/npm/rfcs/pull/14#discussion_r246963804


accepted/0008-no-caret-prerelease-installs.md
```diff
@@ -13,8 +13,8 @@ results in
 
 ```diff
 "devDependencies": {
```
jakeNiemiec: Looking at line 9, should this be dependencies?

PR #32, opened 2019-01-17
Add 0000-script-dependencies.md

edahlseng:
[Rendered](https://github.com/edahlseng/rfcs/blob/script-dependencies/accepted/0000-script-dependencies.md)

### DIFF
--- a/accepted/0000-script-dependencies.md
+++ b/accepted/0000-script-dependencies.md
@@ -1,0 +1,85 @@
+# Script dependencies
+
+## Summary
+
+This proposal introduces dependencies for NPM run scripts.
+
+## Motivation
+
+Often times an NPM run script has dependencies. For example, `npm start` may require source code to be transpiled before starting a server, meaning that `npm run build` needs to be run before `npm start`. Almost always, `npm run build` requires packages that need to be installed, meaning that `npm install` needs to be run before `npm run build`. With dependencies defined, a new user could clone a repository, run `npm start` right away, and NPM would work out that it first needs to run `npm install`, then `npm run build`, and finally `npm start`.
+
+## Detailed Explanation
+
+The `package.json` format will be extended, so that it supports objects for the `scripts` section:
+
+```json
+{
+  "scripts": {
+    "build": {
+      "command": "babel sources --out-dir dist",
+      "depends": ["install", "./sources"],
+      "produces": ["./dist"]
+    },
+    "start": {
+      "command": "node ./dist/server.js",
+      "depends": ["build"],
+    },
+	"hello-world": "echo Hello, World!"
+  }
+  ...
+}
+```
+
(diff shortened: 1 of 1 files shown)


evocateur:
Yeah, I'm not seeing any clear benefits over the existing `pre`/`post` lifecycles. All of the "benefits" presented here should live in the build tools (like `make`) that already implement them.

The `depends` and `produces` fields I am extremely `-1` on, npm is not a generic build tool, it is a package manager.

edahlseng:
@ljharb the difference is that "prestart" would _always_ be run. With this approach NPM would understand that a dependent command may not need to be run (for example, not needing to build files again if the source files haven't changed).

> npm is not a generic build tool, it is a package manager.
@evocateur totally, and by this same logic there would be no need for custom run scripts either. But those are supported because it adds to the developer experience. There are a lot of packages out there that are purely development tool. The JavaScript + NPM community makes it extremely easy to make these development tools, and make them in a shareable way. This approach would simply make it easier to run these development tools, just like NPM currently makes it easy to share these development tools.

ljharb:
How would it determine that source files haven't changed? This seems to differ widely by platform, and in fact there's a number of ecosystem projects to tackle this - all with their own challenges, tradeoffs, and bugs. It seems like a pretty large and complex thing to add to npm.

zkat:
This is literally what `devDependencies` is for, isn't it? Why not use that?

zkat:
We don't really feel like it's in scope for npm itself to solve this problem, so I'm not sure about this specific approach. There's a lot of tools out there already for doing this kind of thing, like grunt, gulp, and make itself, and we don't feel like it's a solved problem. This would be a very significant increase in scope for the CLI.

That said, I think RFCs that describe ways to more seamlessly integrate existing tools like the ones mentioned above would be an interesting avenue to explore. We're open to further discussion on this, but right now, I don't think this RFC is the sort of thing we'd like in the CLI.

edahlseng:
Thanks for all of the feedback, really appreciated!

I haven't worked with grunt or gulp specifically enough to understand how those could be better integrated. I've only worked with make, and this proposal was my best understanding of how I could see the functionality of that tool working more closely together with NPM. I'll check out those other options and do some more exploration to see if there are good ways to integrate them.

I do think that there's potential for NPM to more effectively integrate with the world of build tools, but it sounds like this direction isn't the right one. Would love to continue the conversation with anyone who's interested, whether here or elsewhere!

zkat:
Thank you! I'll close this in the meantime, and I HIGHLY encourage any further discussion on npm integration over in [npm.community](https://npm.community) and especially the [Node Tooling WG](https://github.com/nodejs/tooling) or the [DevTools Community](https://devtoolscommunity.herokuapp.com/), which are both dedicated exactly to doing this sort of integration.

Have a good one! 👋

PR #33, opened 2019-01-19
CLI new feature to install missing dependencies

tanohzana:
This is a suggested feature for NPM CLI that would allow users to install any dependency used in a file or a folder, from the terminal, without opening any of the files.

It was actually already implemented [here](https://github.com/tanohzana/cli), but was not discussed before (sorry, first adding to NPM)

### DIFF
--- a/accepted/0000-npm-missing-dependencies.md
+++ b/accepted/0000-npm-missing-dependencies.md
@@ -28,7 +28,7 @@ These solutions assume that each folder or file copied and paste from a project 
 
 The implementation was done already in a fork here:
 
-`https://github.com/tanohzana/cli`
+[https://github.com/tanohzana/cli](`https://github.com/tanohzana/cli`)
 
 Changes are the following:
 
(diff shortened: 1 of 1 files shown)


zkat:
npm itself getting in the business of statically parsing user source code is tricky, and in the real world, many of our users (at least half, if not most) do not use plain old JavaScript -- any parser that statically analyses missing dependencies must support all of that, and anything in the future.

After talking about it, we think this is the sort of thing that best lives as an `npx` tool, as `missing-dependencies` already is. I might even suggest trying to pick up the [`mp`](https://npm.im/mp) name for yourself, following our [package dispute process](https://www.npmjs.com/policies/disputes) (in summary: email support@npmjs.com and ask them to communicate with the package author and see if they're willing to hand over the package -- if they don't respond, you'll get `mp` as your own package name.

This RFC has sparked some discussion internally about how [`tink`](https://npm.im/tink) would be able to do this more reliably, and also how we can make `npx` a more first-class plugin system for npm, including a curated list of packages (like `mp`!) that we recommend for extra CLI features.

Thank you so much for taking the time to submit this! I'll leave it open for now so we can have any further discussion if needed or other people can chime in, but we will most likely not be integrating this ourselves as described. Have a good one, and thanks again!

PR #34, opened 2019-01-24
RFC for dependencies change script addition

DHedgecock:
[Rendered](https://github.com/DHedgecock/rfcs/blob/dependencies-change-script/accepted/0000-dependencies-change-script.md)

### DIFF
--- a/accepted/0000-dependencies-change-script.md
+++ b/accepted/0000-dependencies-change-script.md
@@ -1,0 +1,51 @@
+# Dependencies Change Script
+
+## Summary
+
+Provide the ability to automatically run a script anytime the the dependencies in a repo
+change.
+
+## Motivation
+
+It would be valuable to run validation scripts after dependencies have changed,
+and would enable automatically running compliance tooling in projects when
+dependencies are installed.
+
+For example, I work at company Rad Developers (not really 😉) and we have an
+internal auditing script `MakeSureWeDontGetSued.sh` we need to run on any
+dependencies we use in our applications. I want to be able to setup a
+package.json script that will be called automatically anytime the dependencies
+change so that compliance is an automated part of every team members' workflow.
+
+## Detailed Explanation
+
+I think one way this could be backwards compatible is by adding a `postChange`
+script to the set of package.json scripts that npm will call during the
+installation lifecycle.
+
+After any of these actions, the npm CLI would call any defined `postChange`
+script:
+
+- `npm install`
+- `npm install some-dep`
+- `npm uninstall some-dep`
+
+## Rationale and Alternatives
+
+I currently don't know of any alternative way to do this. I looked around and
+found a few different questions trying to do this, but no workarounds or
+solutions.
+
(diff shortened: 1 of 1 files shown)


accepted/0000-dependencies-change-script.md
```diff
@@ -0,0 +1,51 @@
...
+
+For example, I work at company Rad Developers (not really 😉) and we have an
+internal auditing script `MakeSureWeDontGetSued.sh` we need to run on any
+dependencies we use in our applications. I want to be able to setup a
+package.json script that will be called automatically anytime the dependencies
+change so that compliance is an automated part of every team members' workflow.
+
+## Detailed Explanation
+
+I think one way this could be backwards compatible is by adding a `postChange`
```
zkat: I think this should be a lifecycle called `change`, where `prechange` runs when we've determined changes will be needed, but no tree changes have been applied, and then `change` and `postchange` would be run after the tree itself gets updated (these should both run after `postinstall`).

accepted/0000-dependencies-change-script.md
```diff
@@ -0,0 +1,51 @@
...
+
+- `npm install`
+- `npm install some-dep`
+- `npm uninstall some-dep`
+
+## Rationale and Alternatives
+
+I currently don't know of any alternative way to do this. I looked around and
+found a few different questions trying to do this, but no workarounds or
+solutions.
```
zkat: This is possible using hooks, but it's nowhere near as convenient.

accepted/0000-dependencies-change-script.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Implementation
+
+_(I'm not very familiar with the npm codebase, if there is interest in this RFC
+I'd love to dive in to find out these details)_
+
```
zkat: You can remove this section. We've been talking about getting rid of it for a while.

accepted/0000-dependencies-change-script.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Prior Art
+
+I don't know of any prior art for addressing this workflow.
+
+## Unresolved Questions and Bikeshedding
+
+There may be a more self-explanatory name than `postChange`
```
zkat: See about about `change`.

accepted/0000-dependencies-change-script.md
Fishrock123: I would certainly like to be able to have a `prechange` and `postchange`!

What would be the difference be between a `change` and a `postchange`, though?

iarna: Likely none, that both exist is more of an artifact of how our lifecycle scripts work. (This is the case with most lifecycle scripts.) In a perfect world, I would only call the non pre/post named lifecycle if it was _replacing_ the action taken at that point, but given the history, keeping things consistent seems better even if it's a little odd to have two lifecycles that do the same thing.

PR #36, opened 2019-06-11
Allow script-shell config in package.json

danielbayley:
RFC proposing that `script-shell` be directly configurable from `package.json`.

[Rendered](https://github.com/npm/rfcs/blob/1afeb550da46dd189d85282aa25de6bc5824efef/accepted/0000-package-json-script-shell-config.md#allow-script-shell-config-in-packagejson)

### DIFF
--- a/accepted/0000-package-json-script-shell-config.md
+++ b/accepted/0000-package-json-script-shell-config.md
@@ -1,0 +1,54 @@
+# Allow [`script-shell`] config in `package.json`
+
+## Summary
+
+Add the ability to configure [`script-shell`] directly in `package.json`.
+
+## Motivation
+
+It would be useful to be able to specify the [`script-shell`] config option somewhere in `package.json`, without having to have a separate one-line `.npmrc` just for this.
+
+## Detailed Explanation
+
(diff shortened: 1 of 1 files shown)


ljharb:
Which would take precedence if both files are defined? Wouldn’t this make `npm config` even slower, since for the first time it’d have to read and parse package.json, instead of just npmrc files and env vars?

isaacs:
My main objection to this is that it seems weird to pick out _this_ config value as the only one to set as a project-level config in package.json.

It seems easy enough to just add a blessed package.json object that would do project-level configs.  The pattern in other node tools (babel, typescript, jest, tap, and nyc, that I'm aware of) is to use the name of the tool as the field name.  So it'd look something like this:

```json
{
  "name": "my-project",
  "version": "1.2.3",
  "npm": {
    "script-shell": "/bin/bash",
    "other-thing": "blarg"
  }
}
```

I don't think that'd be _too_ terrible performance-wise, but it would mean that we have to update the config handling  logic to treat both `./.npmrc` and `./package.json` as project-level config files.  And, if there is both a `npm` field in package.json _and_ a project-level `.npmrc` file, we'd have to decide which one has higher priority (or if it's an error to have both).

isaacs:
As discussed in open RFC meeting today, we're not going to move forward with this, and instead will evaluate #59 and #60 separately, as these would provide a more general solution to the problem identified here.

Thanks!

PR #40, opened 2019-08-02
rfc: Force version lock (Remove caret from default dependencies)

KingZee:
This change will force `npm install` to lock the version to the current installed version by default. Developers, package publishers, and users would need to manually modify their `package.json` in order for `npm update` to be relevant.

### DIFF
--- a/accepted/0000-no-caret-installs.md
+++ b/accepted/0000-no-caret-installs.md
@@ -1,0 +1,27 @@
+# No caret for default dependencies (Force lock version)
+
+## Summary
+
+This change will force `npm install` to lock the version to the current installed version by default. Developers, package publishers, and users would need to manually modify their `package.json` in order for `npm update` to be relevant.
+
+## Motivation
+
+There have been several repeat offenses abusing this issue. Examples include the [purescript package](https://github.com/purescript/npm-installer/issues/12), and the [event-stream package](https://github.com/dominictarr/event-stream/issues/115). Regardless of type of vulnerabilities, motives and accountable people, the real vulnerability here is that because developers and maybe less knowledgable people managing a Node project are unintentionally subscribing to all minor version changes, it is being abused as an attack vector.
+
+Another side-effect of this change is better version control among developers. People will eventually start to intentionally choose which packages to *depend* on, automatically upgrading on either major or minor changes, and which ones not to.
+
+## Detailed Explanation
+
+By forcing the version to be locked to the current version, if a package is compromised in the future, a large portion of the damage will be prevented by not having it installed by default on an update. Since the code is open sourced, for a vulnerability to reach a large subset of users would require it to go unnoticed from the eyes of many maintainers. As the examples above have proved, a vulnerable package only lasts days before it is found. This is largely enough time for the package publisher (and dependents) to either deprecate or remove (or lose dependencies of) the compromised package from the npm registry.
+
+## Rationale and Alternatives
+
+One solution proposed were signing of packages with AES, GPG signatures or similar. This is a non-solution in my opinion because there is already no concept of accountability for npm packages. A perpetrator would willingly admit to sharing a compromised package with bad intentions to little or no repercussions. Usernames also provide some level of anonymity both on github and npm.
+
+## Implementation
+
+https://github.com/npm/cli/blob/ab0f0260e5b6333f98062fb2d9f4f9954d3ee6cd/lib/install/deps.js#L286 I presume the change would need to be done on `computeVersionSpec` to change the default output to a caret-less version.
+
+## Unresolved Questions and Bikeshedding
+
+Has this really never been thought of before?


mkay581:
This seems reasonable given that this has always seemed to be how new users of npm would expect for it to behave anyway. I dont work for npm but, just out of curiosity, is there any public discussion around what led to this decision?

shadowspawn:
Lots of trade-offs with this proposal.

Multiple comments.

1) This will need to more package duplication in installs, but disk space is usually cheap.

2) An upside is that fresh installs or updates do not include new compromised versions of transitive dependencies (which is the focus of this RFC). But isn't there a corresponding downside that discovered vulnerabilities in existing versions will need to be manually updated and published by every package that depends on it, before you can get the suspect version out of your transitive dependencies? My experience is that currently the main obstacle and delay in clearing `npm audit` warnings is waiting for packages with strict dependencies to be republished, and this change would make that far more common.

3) What is the new normal workflow for updating my package direct dependencies? Currently I run `npm outdated` and then `npm update` to get the "fairly safe" updates. Do you expect people to check each dependency and manually specify the new version with `npm install`? Or would there need to be some corresponding new features to ease the workflow?

4) If it was decided locking versions was a good direction, would an alternative implementation be to make package-lock.json a published artefact? This would at least preserve the existing developer control over what versions are considered safe upgrades for direct maintenance, while still locking down the end install versions. (Then perhaps there could also be an option to ignore the package-lock in transitive dependencies for audit fixes...)

ljharb:
I think this is a very bad idea. Locking versions doesn’t lock transitive deps, so it gives you barely any of the determinism you’re hoping for - all it does is prevent non breaking updates from filtering through the dep graph, and also deduping by consumers.

The proper thing to do is always use the caret, and use a lockfile in apps (some also suggest using a lockfile in published things; while i disagree with this, that is decidedly better than pinning versions).

KingZee:
Really great comments! @shadowspawn 

>     1. This will need to more package duplication in installs, but disk space is usually cheap.

This is very true. A fix to this 'issue' is non-trivial either as it would involve overriding dependencies of existing packages. Either way it's a compromise I feel like is necessary because this issue is quite important.

>     2. An upside is that fresh installs or updates do not include new compromised versions of transitive dependencies (which is the focus of this RFC). *But isn't there a corresponding downside that discovered vulnerabilities in existing versions will need to be manually updated and published by every package that depends on it, before you can get the suspect version out of your transitive dependencies?* My experience is that currently the main obstacle and delay in clearing `npm audit` warnings is waiting for packages with strict dependencies to be republished, and this change would make that far more common.

This is true, but there have been cases where a vulnerable package has had to be completely removed from npm rather than wait for an update, which causes longer downtime for packages depending on it and users. But the thing is that, this is a scenario that gets less and less likely on the long run. When a lot of packages start depending on exact versions, this implies that an update that introduces a vulnerability to a package will already not propagate through transitive dependencies, and if it does propagate to *some* packages, then it would mean those publishers are frequent updaters and would likely fix their dependencies in case a vulnerability is found. 

>     3. What is the new normal workflow for updating my package direct dependencies? Currently I run `npm outdated` and then `npm update` to get the "fairly safe" updates. Do you expect people to check each dependency and manually specify the new version with `npm install`? Or would there need to be some corresponding new features to ease the workflow?

The indirect(or not so indirect?) cause of this update is that you will now have to personally version your dependencies as to ease your workflow. You can still run `npm outdated` and see all the `latest` versions, as `wanted` will only display your current version. But if you feel like a package is safe enough, you can edit your `package.json` on installing, and set your own versions. Maybe to avoid having to access `package.json` every time, it would be really neat if npm were to introduce the opposite of the existing directive `--save-exact`, where you'd specify that you're okay with subscribing to minor/major versions. `npm install` already takes ranges, so such a feature could probably be implemented without a lot of work.

>     4. If it was decided locking versions was a good direction, would an alternative implementation be to make package-lock.json a published artefact? This would at least preserve the existing developer control over what versions are considered safe upgrades for direct maintenance, while still locking down the end install versions. (Then perhaps there could also be an option to ignore the package-lock in transitive dependencies for audit fixes...)

I feel like publishing package-lock with a package will override its current use-case. It's true that it's a great restriction in locking versions, but that is in fact a downside in my opinion. I don't support strict locking, what I'm proposing here is just to change the default behavior to exact versions, packages are very much invited to override this default behavior by actively managing their dependency versions. package-lock's usecase is extremely useful as it helps developers clone repositories with exact dependencies, but publishing it would make it very hard for package publishers and users to version their own projects.

@ljharb I am not sure what you mean. Quoting [npm update](https://docs.npmjs.com/cli/update.html) : 

> As of npm@2.6.1, the npm update will only inspect top-level packages. Prior versions of npm would also recursively inspect all dependencies. To get the old behavior, use npm --depth 9999 update.

I was under the assumption that as long as you're depending on an exact version, child semver directives will not be evaluated, unless I'm wrong. 
Hell, even assuming I'm wrong, this change will still bring to light the fact that a lot of package publishers care very little about managing their dependencies, which is the issue this tries to fix. `npm` is a dynamic evolving mutating ecosystem, dependencies are constantly changing, and assuming this change is implemented tomorrow, in a few months or a few years the outcome will be an inherent locking of transitive dependencies like we all want.

The summary is basically, I'm trying to shift the responsability every time a vulnerable package is found from "Well, what can we do, it's how npm works" to "It's *your* job as a package publisher". From a developer standpoint, I don't see the *issue* with locking dependencies: if you can manage your versioning, that is the desired goal, if you can't, or don't care enough to do so, npm does it for you and softly forces your package to never be compromised in the future for users.

ljharb:
I highly disagree with that statement; very few package publishers do not care about managing their dependencies. The problems this “fixes” are incredibly rare and, with a lockfile and a private registry for deployments, trivial to prevent.

Deduping is the power of npm; locking things throughout the graph hamstrings every user of npm.

mkay581:
@ljharb does have valid points, which is why I originally asked what discussions were had that led to this decision (which still hasnt been answered btw) :) Having a little more context around this change may give us the context we need to properly discuss it here.

ljharb:
Note that this isn’t a “decision”, it’s an RFC posted by what seems to be a non-npm collaborator - so we’d need someone from npm to weigh in.

KingZee:
> @ljharb does have valid points, which is why I originally asked what discussions were had that led to this decision (which still hasnt been answered btw) :) Having a little more info around this proposal may give us the context we need to properly discuss it here.

Sorry if I haven't answered you, I have linked in the PR 2 issues of different packages that have been found as vulnerable, one of them with clear intention of harm. You are welcome to look up other sources discussing both events or a few others, I just preferred not to link them here as most of them are quite opinionated.

@ljharb "not caring" might have been an over-statement, but a lot of packages do not change the caret-ed versioning in their dependencies. This is just my opinion but because that's the default behavior of npm, it most likely means the package maintainer didn't bother to change it. If a package *intentionally* chooses to subscribe to all minor updates of a package, then that is absolutely not a problem, but that's the point I'm trying to make here.

ljharb:
If you want intention, then there shouldn’t be a default. Otherwise, the best default is the caret, and not pinning.

isaacs:
We're not going to do this.  The downsides vastly outweigh any potential benefits.

The security analysis here is distorted by an availability heuristic.  Malware makes news, so it seems prevalent.  In reality, unpatched vulnerabilities are _far_ more likely to cause real security problems than malware. But "fixed a vuln, please update" is not as newsworthy, because the story carries less emotional valence.

Consider the effect when malware is discovered, vs when non-malicious vulnerabilities are discovered.  Malware in the registry is removed immediately.  In cases where it's only a single malicious version, applications may need to adjust their lockfiles, but they won't get re-owned on subsequent installs.  However, when a vulnerability is discovered and patched, typically the previous versions are left intact.  (They're not _malicious_, merely dangerous, in some scenarios, and you the human may know best for your app.)  However, if pinned, nothing will stop working, and there will be no easy way to know what can be safely upgraded, and what was pinned on purpose.

Pinning dependency versions means that a library is unlikely to benefit from patches that improve security.  In an application, it might not be so bad, because `npm update` could be modified to pull in updates explicitly or make it more convenient.  However, in a library, as this change spreads throughout the ecosystem, it would mean that your meta-deps effectively never update _either_.

**On balance, one of the worst things the npm CLI could do for application security would be to add any friction at all in the process of applying patch updates.**

Of course, there is nothing stopping _you_ from doing `npm config set save-exact true`.  But imposing this on the community as a whole would have far-reaching and disastrous long-term effects.

The change proposed runs counter to the stated goals of this RFC, and the npm CLI project.  I'm sorry, the cost is just much too high.

isaacs:
Another issue: if everyone pins their deps, then package trees will not be nearly as easy to dedupe.  That's a huge problem for the vast majority of npm users who are shipping their programs to web browsers.

And regarding the point that pinning deps would prevent malware from taking hold, (a) that's already done by the default use of a `package-lock.json` file, and thus (b) it would only  have any effect insofar as people haven't _already_ installed the malware at a pinned version.  Since there's no reason to assume that malware will always be discovered before anyone adds a new dependency (if we could do _that_ it wouldn't be a problem!), (b) is an invalid assumption.

What would happen is that people would still get the malware installed, and then they'd have to not  just run `npm update`, but a _recursive_ update in all the packages depending on that malicious version (which are now broken).

This would have the net impact of making malware clean-up much more costly.

PR #48, opened 2019-08-29
rfc: Hook script linking

lachlanhunt:


isaacs:
My main concern to investigate would be timing during installation.

Ie, if you `npm install`, and one of the packages being installed adds a `preinstall` hook, does it fire for the modules installed afterwards?  If it adds a `postinstall` hook, does it fire for itself?

Even if we're ok with whatever the current default behavior would be, I'd like it documented so we don't break folks later by changing some undefined behavior.

wesleytodd:
Since there can be only one file of each name (thus each of the predefined list of hook scripts, you would need some resolution logic to determine which to link wins or how to call them all (possibly changing the way that feature works today). Also, allowing any hook from any direct or transitive dep to be called for every installed dependency just by defining one in its `package.json` seems like an unintended (or at least unexpected) behavior.

Maybe I am not seeing the clear path forward for this RFC, but it seems like it is missing some large issues as it is defined today.

darcyclarke:
Closing this as hook scripts were removed in v7

PR #51, opened 2019-09-15
rfc: Expose arborist metadata to lifecycle scripts

vweevers:
Following the conversation in https://github.com/npm/rfcs/pull/38#issuecomment-529025921.

FYI @lovell @ralphtheninja

### DIFF
--- a/accepted/0000-arborist-lifecycle-scripts.md
+++ b/accepted/0000-arborist-lifecycle-scripts.md
@@ -1,0 +1,43 @@
+# Expose arborist metadata to lifecycle scripts
+
+## Summary
+
+When running any (?) lifecycle script for a package, insert the data for that package from `.arborist-metadata.json` into the environment.
+
+## Motivation
+
+For context, [`prebuild-install`](https://github.com/prebuild/prebuild-install) is a CLI that downloads prebuilt binaries for native modules and falls back to compiling the native module when no prebuilt binaries are available. It is typically used in the npm `install` lifecycle script, so that the prebuilt binary is available to a user after running `npm install`. In some cases, it is not desirable to download a prebuilt binary.
+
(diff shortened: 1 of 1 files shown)


isaacs:
This will be included in v7, and _may_ require a fall-back lockfile to support cases where the user installs with `--no-shrinkwrap` or deletes the shrinkwrap file.

ljharb:
When you say a "fallback lockfile" do you mean like, a temporary `npm install --package-lock --package-lock-only`? That'd be great (and it's what would make `npm audit` work for this scenario as well).

isaacs:
@ljharb The idea would be a copy (or hard-link) of the lockfile at `node_modules/.arborist-lock.json`.  It would not be authoritative during install (like a file at `./npm-shrinkwrap.json` or `./package-lock.json` would be), but it would give us a place to stash metadata whenever we change or read the package tree if the user is running with `--no-shrinkwrap` or `--no-save`.

That would also let us make things like `npm ls` and `npm fund` much faster, since we can be reasonably confident that bundled fallback lockfile reflects what's actually in the tree.

isaacs:
And yeah, it would make `npm audit` work even if there is no lockfile in the project root.

ljharb:
That sounds awesome, especially if `npm install --save` and friends iteratively updated it.

isaacs:
> That sounds awesome, especially if `npm install --save` and friends iteratively updated it.

The idea would be that it gets updated even when `--no-save` is set.  It'd be kept in sync 100% of the time, regardless of what we do (or don't do) to the package.json and package-lock.json.

isaacs:
I think the needs here are addressed by https://github.com/npm/rfcs/pull/90.

Basically, arborist will set `npm_package_from`, `npm_package_resolved`, and `npm_package_integrity`, for the package having its lifecycle script run.

I presume that will satisfy your needs?

vweevers:
Looks good, thanks!

I think the removal of `npm_config_argv` will also affect prebuild & friends, however. I'll check and then comment in #90.

PR #54, opened 2019-10-18
Add `funding` support & `npm fund` subcommand

darcyclarke:


accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,117 @@
...
+
+* Add a `funding` field to `package.json`
+  * supports a string URL or object with...
+  * keys that support arbitrary string identifiers
+    * ex. `"sponsor"`, `"sponsors"`, `"donations"`, `"bounties"`, `"contributors"`, `"patrons"` etc.
+  * values that represent a string URL or an array of URLs
+* Add notification at the end of output of package installation that references the number of packages with `funding` defined 
+  * ex. `23 packages are looking for funding. Run "npm fund" to find out more.`
+* Add `--no-fund` flag to opt-out of the funding notification when installing
+* Add `npm fund` subcommand which will: open the URL (string) **OR** list the funding references defined (object)
```
djsauble: The open URL scenario seems like something you'd do if an application has only _one_ fund-seeking dependency which has provided only a _single_ URL as its funding option. This seems like it will be a pretty rare case, especially as more packages start to incorporate the `funding` attribute.

Why bother opening the URL at all? It will happen sporadically at best.

ruyadorno:
thanks for documenting all that @darcyclarke 🙏

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,117 @@
...
+* Rationale: This is a straightforward and easily implemented solution to take the first step toward supporting Open Source Package Maintainence sustainability
+* Alternative: Do nothing
+* Alternative: Continue debating a larger, more complex schema /w correspondingly more complex tooling
+
+## Implementation
+
+* Add a `funding` field to `package.json`
+  * supports a string URL or object with...
+  * keys that support arbitrary string identifiers
+    * ex. `"sponsor"`, `"sponsors"`, `"donations"`, `"bounties"`, `"contributors"`, `"patrons"` etc.
```
feross: Do these string identifiers have a defined meaning? If not, then how does the CLI know which URL or URLs to display when running `npm fund`?

If the goal is to just do something quick, simple, and more useful than nothing, I'd recommend just starting with a string value that points to a URL with more information on funding, and drop support for an object with arbitrary keys. Then `npm fund` can just straightforwardly print out the funding URL for each package.

accepted/0017-add-funding-support.md
darcyclarke: @djsauble this is a good point and I will update the RFC based on that note. This was intended to follow a similar pattern to the `npm repo <pkg>` command which tries to open the correct URL.

accepted/0017-add-funding-support.md
darcyclarke: @feross I made a few changes to this RFC based on both @djsauble's & your feedback. I'm totally okay with de-scoping the work to only focus on a singular URL although I think the hope was to pull in some of the ideas behind the PM WG's more verbose [`backing` schema](https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/PACKAGE-SUPPORT.md#support-backing).

Hopefully the updated proposal for `npm fund <pkg>` makes it a bit more clear in terms of when/how we'd try to open a URL **versus** print out the information. 

Again, if we want to simplify this to only accommodate the first scenario, I'm all for that (as it would also make the `npmjs.com` changes that much simpler as well).

djsauble: One concern I have is that allowing a free-for-all with respect to these identifiers will make it harder to have consistency later on. We could perhaps come up with a set of "reserved" identifiers that people are welcome to use now with the understanding that they may have specific tooling behavior applied to them later.

But I also wouldn't mind dropping the arbitrary keys from v1 if the WG agrees.

lholmquist: >  We could perhaps come up with a set of "reserved" identifiers that people are welcome to use now with the understanding that they may have specific tooling behavior applied to them later.

I think this could become problematic.  It isn't always easy to take something away from users even if we are stating it ahead of time

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,137 @@
...
+  * values that represent a string URL or an array of URLs
+* Add notification at the end of output of package installation that references the number of packages with `funding` defined 
+  * ex. `23 packages are looking for funding. Run "npm fund" to find out more.`
+* Add `--no-fund` flag to opt-out of the funding notification when installing
+* Add `npm fund <pkg>` subcommand: 
+  * if a singular URL is defined for the specified package's `funding` field, it will try to open it using the `--browser` config param (similar to `npm repo <pkg>`)
```
djsauble: This is better, thanks.

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,137 @@
...
+
+* See [Prior Art](#prior-art)
+* Provide a means to **reference** existing backers or funding opportunities in `package.json`
+* Provide a means to **define the type** of backing or funding opportunities in `package.json`
+* Provide a means to **notify** package consumers of backers or funding opportunites with the cli
+* Provide a means to **view** dependency's backers or funding opportunites with the cli
+* Provide a means to **view** a package's backers or funding opportunites
+
+##  Rationale & Alternatives
+* Rationale: This is a straightforward and easily implemented solution to take the first step toward supporting Open Source Package Maintainence sustainability
```
djsauble: There's another rationale here, which is that building this functionality into the package management layer is the _only_ reliable way to _push_ funding notifications to the developers who use npm.

All other solutions rely on an opt-in model, which means they reach only a fraction of the desired audience.

darcyclarke: ✅ Added

accepted/0017-add-funding-support.md
ljharb: I would suggest - even if it’s more boilerplate - something like:
```json
"funding": {
  "bikeshed": {
  }
}
```
where every key of “bikeshed” is arbitrary and unrestricted, and thus serves as an ID for other things - tools, the “support” field the package management group is working on - to use to address the funding mechanism.

This keeps the design space open for any future additions to the funding object, at the (obviateable vía tooling) cost of adding an additional nesting to the funding object.

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Implementation
+
+* Add a `funding` field to `package.json`
+  * supports a string URL or object with...
+  * keys that support arbitrary string identifiers
+    * ex. `"sponsor"`, `"sponsors"`, `"donations"`, `"bounties"`, `"contributors"`, `"patrons"` etc.
+  * values that represent a string URL or an array of URLs
+* Add notification at the end of output of package installation that references the number of packages with `funding` defined 
+  * ex. `23 packages are looking for funding. Run "npm fund" to find out more.`
```
ljharb: This should probably differentiate between top-level and transitive deps somehow - it’s far more important to fund your direct deps (who may in turn fund their direct deps) than to support some arbitrarily deep transitive dependency, imo.

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+* Add `--no-fund` flag to opt-out of the funding notification when installing
```
ljharb: presumably also an npm config eng var/setting

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+  ...
+}
+```
+
+**Example of `npm fund <pkg>`:**
+```
+$ npm fund example-package 
+└─ example-package
+   ├─ sponsor
+   │  ├─ 0: https://github.com/users/my-account/sponsorship
```
ljharb: ideally the indices won’t be displayed here; surely it can be displayed as a list?

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+* [GitHub Sponsors](https://github.com/sponsors)
+* [License Zero](https://licensezero.com/)
+* [GitCoin](https://gitcoin.co/products)
+* [Feross: `thanks`](https://github.com/feross/thanks)
+* [Feross: `funding`](https://github.com/feross/funding)
+* [Indieweb: payment](https://indieweb.org/payment)
+* [microformats: `rel-payment`](http://microformats.org/wiki/rel-payment)
+* [Shields.io: Funding](https://shields.io/category/funding)
+* [ThanksApp: Donate Spec](https://github.com/ThanksApp/donate-spec)
+*  [Bevry: `sponsored`](https://github.com/bevry-archive/sponsored)
```
ljharb:
```suggestion
* [Bevry: `sponsored`](https://github.com/bevry-archive/sponsored)
```

accepted/0017-add-funding-support.md
ruyadorno: I disagree with this specific comment, nested dependencies already takes a big hit in terms of awareness and in many ways they can be as vital as the first-level ones to a given project

I would rather steer away from making an explicit differentiation between them for the initial implementation and see how the community reacts to that 😊

ljharb: @ruyadorno they should be visible, but transitive deps can be transparently refactored in and out all the time, so it's not the same thing as direct ones.

accepted/0017-add-funding-support.md
isaacs: I would suggest that, at the outset, we specify that there are exactly two fields: `url` and `type`, mirroring the `repository` field we have today.  `url` is mandatory, `type` is optional, and can often be inferred from the url.  (Eg, if it's at `patreon.com`, we know it's `"type": "patreon"` and can interpret it that way in `read-package-json`.)

The spec should state that additional fields MAY be added in the future, and so it is not safe to add them arbitrarily.  People still will, but then they have fair warning when we assign an official meaning to some of them in the future.

accepted/0017-add-funding-support.md
isaacs: I'd clarify this to say that a bare string value is interpreted as the value of the `url` field, and that we open the `url` string in a browser.

accepted/0017-add-funding-support.md
isaacs: All configs are in all places, so yes, this is implied.  (Adding CLI flags that are _not_ a first-class config values would be a much bigger RFC.)

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+Package maintainers want to clearly indentify how their software is currently, or could be in the future, supported monetarily.
+
+## Detailed Explanation
+
+* See [Prior Art](#prior-art)
+* Provide a means to **reference** existing backers or funding opportunities in `package.json`
+* Provide a means to **define the type** of backing or funding opportunities in `package.json`
+* Provide a means to **notify** package consumers of backers or funding opportunites with the cli
+* Provide a means to **view** dependency's backers or funding opportunites with the cli
+* Provide a means to **view** a package's backers or funding opportunites
```
isaacs: The "backers" term should be defined here, or removed.  To my interpretation, this would imply that we're supporting a way to list out the people supporting a package, which feels very out of scope at this stage.

accepted/0017-add-funding-support.md
isaacs: I agree with @ruyadorno that we should certainly include transitive dependencies by default.  We could indicate the difference in the UI, but they need to be included.

isaacs: That is, I'm saying, while we could indicate which ones are direct deps and which ones are transitive, we should endeavor to _not_ do so in a way that steers people towards only funding their direct dependencies.

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+  * supports a string URL or object with...
+  * keys that support arbitrary string identifiers
+    * ex. `"sponsor"`, `"sponsors"`, `"donations"`, `"bounties"`, `"contributors"`, `"patrons"` etc.
+  * values that represent a string URL or an array of URLs
+* Add notification at the end of output of package installation that references the number of packages with `funding` defined 
+  * ex. `23 packages are looking for funding. Run "npm fund" to find out more.`
+* Add `--no-fund` flag to opt-out of the funding notification when installing
+* Add `npm fund <pkg>` subcommand: 
+  * if a singular URL is defined for the specified package's `funding` field, it will try to open it using the `--browser` config param (similar to `npm repo <pkg>`)
+  * if an object is defined for the specified package's `funding` field, it will print out the tree of values
```
isaacs: I'd trim this out for now.  Feels like biting off too much at this point, and commits us to supporting whatever arbitrary stuff people put there.  We should be more opinionated about specifically which fields we treat specially, and what each are for.

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+  * if no package is specified, `npm` will try to print out a tree of all the funding references defined in the current project's installed dependencies
```
isaacs: `s/try to//`

accepted/0017-add-funding-support.md
```diff
@@ -0,0 +1,138 @@
...
+    "sponsors": "https://github.com/users/my-account/sponsorship#sponsors",
+    "contributors": "https://opencollective.com/my-account#section-contributors",
+    "patrons": [
+      "https://patrons-site-one.com/",
+      "https://patrons-site-two.com/",
+      "https://patrons-site-three.com/"
+    ]
+  }
+  ...
+}
```
isaacs: All of these other fields should be entire explicitly part of this spec, or deferred to a future RFC.

accepted/0017-add-funding-support.md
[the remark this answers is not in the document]
ljharb: That will help transitive deps not be so overwhelming, +1

accepted/0017-add-funding-support.md
ljharb: @isaacs why? people only have control over their direct deps. As a maintainer, if I swap out my package A's internals from dep B to dep C, and the consumers of A have been funding B (perhaps without realizing that B is there because A needs it), then shouldn't C be getting that funding instead?

In other words, imo it should be on the developers of each dep to fund their own deps - otherwise maintainers may be incentivized to fork/inline rather than have deps themselves. This doesn't seem like a desired consequence.

isaacs: @ljharb Because our dep chains are deep, and in practice, it doesn't trickle down efficiently.

There isn't an ideal solution, but imagine the situation where a module A needs funding, and is only a direct dep of B (which absolutely critically depends on it!), and B (which doesn't seek funding) is used by the 50 most popular npm packages.

If we only show the funding nag for direct deps, then we show B's nag to one person.  If we show it for all deps, we expose it to 5M devs.  If I were the maintainer of B, writing code that is downloaded billions of times each month and depended on by every Fortune 1000 company, it'd seem pretty unfair that my request for funding only goes out to one person.

The ideal solution, of course, is for npm to act as a broker, using the insight we have about downloads and dependencies to fairly direct funds where they're needed.  But that is a longer-term idea, and fraught with its own complexities.

isaacs: Of course, you can replace "doesn't show the nag" with "shows the nag but in a way that nudges you away from donating", and draw a conclusion that only differs in degree.  Metadeps are deps.

ljharb: @isaacs that's all fair, and i very much wouldn't want to *hide* transitive dep funding info - i'm mainly just saying it should be differentiated, so that users aren't confused about where the info comes from.

giuseppeg:
> npm just take my money and give it to the right people, based on what packages I'm using.

@isaacs I think that if npm goes for this option the impact will be huge. Please consider an alternative to mainstream donation-based models! The community doesn't need another "sponsors" program where only webpack, Babel and Sindre Sorhus like folks get funded because of their popularity (they deserve it but there are plenty of less known projects and folks that should get funding).

wesleytodd:
Agreed it would be good, but much better to leverage existing platforms which are working on this.  For example I think Open Collective is working on something called Back My Stack.  (cc. @piamancini who was pinged when we brought this up for Express).

brodybits:
> much better to leverage existing platforms which are working on this

My experience is that the "existing platforms" do not help all open source maintainers.

Open source sustainability seems to be a big, unsolved problem.

A big thank you to @darcyclarke, @isaacs, and many others who have been working hard on a solution that can help everyone (hopefully).

ruyadorno:
for everyone still following this thread, follow up PR in **npm/cli** is here: https://github.com/npm/cli/pull/273

isaacs:
For what it's worth, we have been chatting with the OC folks, and intend to continue to work closely with them where it makes sense.  Today a lot of people who'd use the `npm fund` feature are going to put open collective URLs in there, so I don't think we're working at cross purposes at all, and our intent is to stay aligned.

PR #68, opened 2019-11-19
RFC: Multiple Funding Sources

ljharb:
# What / Why

Allow the `funding` field, which currently can contain a string or an object, to contain an array of those things as well.

See the document for more details.

### DIFF
--- a/accepted/0018-multiple-funding-sources.md
+++ b/accepted/0018-multiple-funding-sources.md
@@ -1,0 +1,65 @@
+# RFC: Multiple Funding Sources
+
+## Summary
+
+Allow the `funding` field, which currently can contain a string or an object, to contain an array of those things as well.
+
+## Motivation
+
+ - One package author can have more than one source of funding - GitHub Sponsors and Tidelift, for example.
+ - One package can have multiple authors, each with their own source(s) of funding: two different GitHub Sponsors, for example.
+ - This mirrors how GitHub's own sponsor button works as well: a repo can have multiple funding sources, and any source whose model is not a single project typically can have an array of multiple users.
+
+Being forced to create a landing page - especially on an unknown domain, defeating any inference of the `type` from the URL - is a high burden for anyone in this scenario.
+
+## Detailed Explanation
+
+ - the `funding` field, which currently can have an X, will also be able to have an array of X.
+ - the `npm fund` output will display all URLs in the array.
+
+## Rationale and Alternatives
+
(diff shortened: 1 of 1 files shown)


ljharb:
@mikemimik i'd love to get this approved and into 6.14 (happy to do the implementation PR myself); will there be time, assuming it's approved in the meeting tomorrow?

mikemimik:
@ljharb provided it gets accepted tomorrow, of course we can get this in! If you've got time to implement, then that pull-request would be welcomed with open arms. 👍

ljharb:
This RFC has been updated, and I've filed https://github.com/npm/cli/pull/731 to implement the feature (under the presumption that this RFC will be accepted and merged in the next RFC call, absent any objections)

accepted/0018-multiple-funding-sources.md
```diff
@@ -0,0 +1,91 @@
...
+
+The `package.json` side is surely trivial; the difficulty will be in `npm fund` output. There will be a bit of UI design needed, and that will determine the implementation difficulty.
+
+The `--json` output of `npm fund` will also change when there's an array, but this should be trivial for consumers - change `process(fundingValue)` to `[].concat(fundingValue).map(process)`, or something equivalent. Since this sort of change was considered acceptable for https://github.com/npm/cli/pull/472, I presume it would be acceptable here as well.
+
+Expected code that will need to change is the same as in this PR: https://github.com/npm/cli/pull/472/files
+
+## Prior Art
+
+ - The `repository` field works in this same manner; it can be a string, an object, or an array of either.
```
tunnckoCore: Huh, definitely didn't know that! Cool. But in same time, what's the use case for array? And I'm curious how this will be shown in the npm page :D

ljharb: @tunnckoCore see the RFC; but i have 200+ packages, all of which could have both the same Github Sponsors URL for me personally, and a unique Tidelift URL for the package.

tunnckoCore: That's clear, i value that. I asked about the `repository` field, because didn't know it support an array.

ljharb: In that case this might be a mistake on my part :-) https://docs.npmjs.com/files/package.json#repository says it can be a string or an object; I was thinking of the "author" field (which is also string or object) vs the "contributors" field (which is an array of strings-or-objects).

tunnckoCore: Hahaa. Good catch, huh?

Cool, everything is clear now :)

PR #71, opened 2019-11-26
Verified account linking

djsauble:
# What / Why

Add a verified way to link your npm account with other accounts that you own.

OSS works because of its web of trust. When anyone can masquerade as anyone else, this web of trust is weakened. We need a way for someone linking their (for example) GitHub or Twitter account to prove ownership of those linked account.

See the RFC for more details.

### DIFF
--- a/accepted/0000-account-linking.md
+++ b/accepted/0000-account-linking.md
@@ -1,0 +1,66 @@
+# RFC: Account Linking
+
+## Summary
+
+Add a verified way to link your npm account with other accounts that you own.
+
+## Motivation
+
+Today, you can add your GitHub or Twitter username to your npm profile, which
+helps the community figure out who you are on other services. However, this is
+not a verified process, and thus it is possible to masquerade as someone you're
+not. By adding integrity to account linking, we solve this problem and also
+create the potential for authentication via third-party services.
+
+## Detailed Explanation
+
+We need a new subcommand in the CLI that will enable an authenticated user to
+link their npm account with an account on another service. To avoid vendor
+lock-in, we will not define a specific list of third-party service in the CLI,
+and will instead leave it to the upstream registry to figure out what services
+are available for linking and inform the CLI accordingly.
+
+As an additional security measure, we will also add nonce support to the CLI to
+avoid the possibility of replay attacks.
+
+## Rationale and Alternatives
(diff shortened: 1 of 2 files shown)


ljharb:
Would this be queryable from the outside, ie, if i know someone's github, i could trivially determine which npm account is linked to it?

ljharb:
Would this be queryable from the outside, ie, if i know someone's github, i could trivially determine which npm account is linked to it?

djsauble:
Potentially, yes! The registry will store that information. This RFC outlines a CLI-based interface for interacting with that information, which doesn't preclude making it available via a general-purpose API.

mikemimik:
Would these all just be OAuth handshakes? I know I've come across similar features on other platforms. If I make an account on npm, and want to link my github, I have to log into github to prove I own that account. Which is an OAuth flow right? Can we still accomplish this from the CLI?

ljharb:
The CLI can provide you a URL you need to visit to complete the oauth flow, which is a standard way to complete the oauth dance from the CLI.

ThisIsMissEm:
@djsauble have you looked into something like how keybase verifies identity? https://keybase.io/docs/proof_integration_guide

ljharb:
@nickmccurdy is this queryable via the api? also, can i link an npm org to a github org?

nickmccurdy:
> is this queryable via the api?

Yes: https://registry.npmjs.org/-/npm/v1/user

> also, can i link an npm org to a github org?

I can't confirm this since I'm not in any npm orgs, but I'm assuming no, since this is documented as an account level feature.

ljharb:
I'm not sure the RFC can be considered completed if it doesn't work for orgs?

nickmccurdy:
You could at least verify based on who published the package though, right?

ljharb:
if it was based on publishes it’d be in package.json; this is a momentary setting that isn’t connected to publishes.

nickmccurdy:
I mean you could programatically find the publishing user and then look up their profile via the api.

ljharb:
true - altho unfortunately the packument only preserves the npm username at the time of publish, not the github account they were linked with at the time, and i'm not sure the github api can tell me what historical permissions an account had

PR #73, opened 2019-12-05
RFC for `npm run` to traverse directory tree up to the root before failing

agorovyi:
# Let `npm run` traverse monorepo directory tree up to the root before failing

With this request I would like to be able to execute `npm run` command that will traverse the directory tree up to the root level if it can't find a required `/node_modules/.bin` with executable command in subdirectory where it was executed from.

A detailed explanation and the rationale is covered in the `.md` file.

### DIFF
--- a/accepted/0000-npm-run-traverse-directory-tree.md
+++ b/accepted/0000-npm-run-traverse-directory-tree.md
@@ -1,0 +1,106 @@
+# Let `npm run` traverse monorepo directory tree up to the root before failing
+
+## Summary
+
+Allow `npm run` command to traverse the directory tree up to the root level if it can't find a required `/node_modules/.bin` with executable command in subdirectory where it was executed from.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


isaacs:
Ah!  I had missed this PR, but I was working on the implementation for #90 today, saw the bit of code that sets up the `PATH` environ, and had a suspicion that this would be an issue.

In fact, rather than splitting on `node_modules`, it should append `node_modules/.bin` to _every_ parent path.  Eg, a module at `/x/y/z/node_modules/a/node_modules/b` would have a PATH containing:

```
/x/y/z/node_modules/a/node_modules/b/node_modules/.bin
/x/y/z/node_modules/a/node_modules/node_modules/.bin
/x/y/z/node_modules/a/node_modules/.bin
/x/y/z/node_modules/node_modules/.bin
/x/y/z/node_modules/.bin
/x/y/node_modules/.bin
/x/node_modules/.bin
/node_modules/.bin
```

I think it might be good to set a "project root" dir to not walk up past (ie, `/x/y/z` in this example), but getting that reliably is a little bit tricky, and I'm not sure it matters much.

So, I think npm v7 will satisfy you there by default.

As a corollary (which I'm not sure how or if to take advantage of?) is that we _could_  put metadeps in `./node_modules/node_modules/`, so they're accessible to anything in `node_modules`, but not to the root project.  That could prevent unlisted deps, which are a problem in some cases.

PR #81, opened 2019-12-20
remove --depth from npm update

isaacs:
see rfc


isaacs:
Ratified without objection (or even very much comment, really) in meeting #99.

PR #85, opened 2020-01-05
2fa code to copy instead only scan QR.

RunAge:
# What / Why
Display 2FA code to copy instead only QR code to scan. It will help to add TOTP/2FA to application when camera is broken.

## References
n/a


RunAge:
Ugh this repo is only for CLI

PR #92, opened 2020-01-22
RFC: Add staging workflow for CI and human interoperability

djsauble:
See the RFC

### DIFF
--- a/accepted/0000-staged-packages.md
+++ b/accepted/0000-staged-packages.md
@@ -1,0 +1,90 @@
+# RFC: Staged Packages
+
+## Summary
+
+Staging workflow which allows CI to publish new package versions, while allowing
+humans to be the final arbiters of general availability.
+
+## Motivation
+
+Anything published by CI is instantly available for general use, which isn't
+ideal when a human is required to be in the loop. And even when fully automated
+CI/CD is allowed, the inability to use 2FA with auth tokens is a security risk.
+
+## Detailed Explanation
+
+We need a minimum of four things to solve these problems:
+
+1. Staging area on the registry invisible to most `npm` commands
+2. Auth tokens that can be scoped to the staging area
+3. CLI flag which lets `npm` commands act on the staging area
+4. CLI command that lets users promote a staged package
+
+## Rationale and Alternatives
+
+One alternative is to defer to third-party repository managers for staging
+functionality. However, this requires that users set up and manage yet another
+tool, which most won't bother to do. In addition, even if users do acquire those
+tools, because staging commands aren't built into the npm CLI, they have to
+learn a new interface to perform staging operations, which adds unnecessary
+friction to the developer experience and again encourages the wrong sorts of
+publishing behaviors.
+
+Another alternative is to build this using our existing `dist-tag`
+functionality. The appeal is that it uses a mechanism we've already built. The
+downside is that we would have to actively ignore these tagged packages in the
+CLI and would be eating up versions as well, not ideal.
+
+Yet another alternative is to build a fancy release manager application,
+separate from the CLI. However, this is a larger project, and we can almost
+certainly implement a useful staging workflow based on tooling that already
+exists. Also, even if we start by adding this to the CLI, it doesn't mean we
+can't build a more complex application on top of the CLI later on.
+
+After considering these alternatives, we belive a staging workflow built into
+the npm CLI provides the shortest path to useful staging functionality that we
+can iterate upon. It will allow us to test these new concepts in a tool with
+which we and our community are both familiar.
+
+## Implementation
+
+* Add a `stagedVersions` object to the packument which lists all packages that
+  have been staged and not promoted.
+* A new `--stage` flag for the `npm publish` command that allows a package to be
+  published to staging.
+* A new `--stage` flag for the `npm install` command that allows a package to be
+  installed from staging.
+* A new `--stage` flag for the `npm view` command that allows us to retrive data
+  for staged packages.
+* A new `npm promote` command that allows a user to promote a staged package.
+* The ability to set a `staging` scope on auth tokens so we can grant read or
+  publish access to staging.
+* A new user permission that allows us to grant a user the ability to read or
+  promote packages in staging.
+  * These are separate from the existing read and publish permissions, since we
+    may want a user to have access to staging, but not production, and
+    vice-versa.
+
+## Prior Art
+
+* Pull Requests on GitHub allow humans to serve as the gatekeeper before new
+  code is merged into master. The "two +1s" rule helps reduce the risk of bad
+  code getting merged in, and having a CI system configured to build from master
+  means that approving a PR is conceptually similar to the gatekeeping
+  capabilities in the staging workflow we've defined above.
+* Staging in third-party repository managers, like Nexus Repository Manager,
+  also meets any gatekeeping requirements.
+  * In Nexus 2, this was implemented via dedicated staging repositories, where
+    packages could be published and then promoted to a production repository,
+    all via a nice staging UI.
+  * In Nexus 3, they shifted to an API-first approach, where new packages had
+    staging metadata set on them, and the API could enable CI and other tooling
+    to stage or promote packages by updating this metadata.
+
+## Unresolved Questions and Bikeshedding
+
+* Currently out of scope (but up for debate):
+  * Webhook support
+  * Metadata related to fitness of staged packages
+  * Multiple staged packages (current spec has a limit of one, successive stages
+    clobber the previous staged version).
--- a/accepted/0000-staged-packages.md
+++ b/accepted/0000-staged-packages.md
@@ -54,7 +54,7 @@ which we and our community are both familiar.
   published to staging.
 * A new `--stage` flag for the `npm install` command that allows a package to be
   installed from staging.
-* A new `--stage` flag for the `npm view` command that allows us to retrive data
+* A new `--stage` flag for the `npm view` command that allows us to retrieve data
   for staged packages.
 * A new `npm promote` command that allows a user to promote a staged package.
 * The ability to set a `staging` scope on auth tokens so we can grant read or


dominykas:
Sorry, maybe I missed something, I know there's the RFC meetings, but can't find the calendar for them - when is this going to be discussed? Is it better to wait for the meeting or post the questions now?

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Implementation
+
+* Add a `stagedVersions` object to the packument which lists all packages that
+  have been staged and not promoted.
+* A new `--stage` flag for the `npm publish` command that allows a package to be
+  published to staging.
+* A new `--stage` flag for the `npm install` command that allows a package to be
+  installed from staging.
+* A new `--stage` flag for the `npm view` command that allows us to retrive data
```
dominykas:
```suggestion
* A new `--stage` flag for the `npm view` command that allows us to retrieve data
```

isaacs:
The registry will support a data structure like this:

```json
{
  "stagedVersions": {
    "whatever": "other".
    "metadata": "we wanna",
    "put": "here",
    "versions": {
      "1.2.3": { "manifest": "data..." },
      "1.2.4": { "another": "staged manifest..." }
    }
  }
}
```

Discussion still to be had about what the `dist` field ought to look like in staged version manifests.

ljharb:
Ideally there'd be more information than just "version", "package.json", and "tarball" - git sha, in particular, would be helpful?

wesleytodd:
On the call we discussed the workflows around the tarballs and the package locks.  The concern was that if you stage a package and then a users installs from the staging it would mean their tarball location moved and then they would have to do an update of their package lock.

The reason I was concerned about this is because we talked about installing from the staged location, and I imagined that this would enable beta testers and other similar workflows.  This was a mistake.  Any workflow which relies on testing from the staging area should been done with pre-release versions and dist-tags.

I have changed my thoughts here and I think as this feature gets rolled out we should make sure it is clear that is not a supported use case.  I would like to propose two additions:

1. The `--stage` flag on install only works when installing a single package
2. Installing from `--stage` is not reflected in the `package.json` or `package-lock.json`

I think these would help prevent accidental usage of this feature in a way which I had originally imagined which had problematic DX.  You would still be able to run something like a CI test by running `$ npm i --stage pkg`, but it would not persist across installs or be locked accidentally.

ljharb:
I don’t think it should be possible for anyone but an owner to install a staged build; and ideally with the restrictions Wes outlined.

wesleytodd:
I thought about adding that, but I wonder if there is reason to let un-authed CI servers install it?  So I could set my publish CI pipeline to look for a staged version to test against but let it do so unauthed so I didn't need a CI token for that part.

ljharb:
If unauthed people can install a staged version, they can depend on it, and staging serves no purpose.

wesleytodd:
They can depend on it by going strongly out of their way if we add the package.json changes.  Seems fine to me if they would have to do `npm i --stage pkg`  every build separate from `npm i`.  But then this makes me thing we should for sure also specify that once promoted the `--stage` version tarball is no longer available.

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+can iterate upon. It will allow us to test these new concepts in a tool with
+which we and our community are both familiar.
+
+## Implementation
+
+* Add a `stagedVersions` object to the packument which lists all packages that
+  have been staged and not promoted.
+* A new `--stage` flag for the `npm publish` command that allows a package to be
+  published to staging.
+* A new `--stage` flag for the `npm install` command that allows a package to be
```
dominykas: It might be worth clarifying here that semver rules still apply (i.e. the staged version, if any, will only be installed if it matches the specifier in `package.json`)

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+  installed from staging.
+* A new `--stage` flag for the `npm view` command that allows us to retrive data
+  for staged packages.
+* A new `npm promote` command that allows a user to promote a staged package.
+* The ability to set a `staging` scope on auth tokens so we can grant read or
+  publish access to staging.
+* A new user permission that allows us to grant a user the ability to read or
```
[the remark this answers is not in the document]
dominykas: In the call, it was clarified that there'd be an option of a token with r/w permission without 2FA specifically for staging.

dominykas:
Some things we also discussed in the call, that should probably make it into the RFC (sorry for duplicating stuff already being discussed, just making sure we've got everything covered):

- for audit purposes, it should be possible to see both users - the one who staged the package and the one who promoted it
    - in the meeting we discussed that `npm_user` is the person who published, and a new field needs to be added for the person who promotes the package
- what happens with the lock files 
    - can they remain unchanged after the package is promoted? I think we discussed that the answer to this one is probably "yes", i.e. if the client detects that the version was staged, but no longer is, and the integrity matches, then there is no need to modify the lock file?
    - if a different tarball is staged at the same version, then install needs to fail
    - is there a need for a new flag in the lock file or should the fact the tarball is staged be detectable from the URL or smth?
    - if you have a staged package (either by an explicit flag or detected via URL) in the lock file - is it ok to install without the `--stage` flag?
- what happens with the tarball URLs
    - does the URL change after the package is promoted?
    - can the URLs be aliased?
    - are the tarballs at the staged URL mutable?
    - caching is also relevant in the context of this discussion (for people using various proxies and replicating registries)

---

Mike also suggested we post the use cases for this feature. For me, right now, the only use case I do care about a lot is to have an area where I can put a tarball without 2FA, and then promote that tarball with 2FA. I don't particularly care if that tarball is installable, TBH - and maybe that's also an option for the MVP, just so we have something to start playing with?

dominykas:
> Installing from --stage is not reflected in the package.json or package-lock.json

Not sure that's possible, though? If `package.json` is not updated, then it will retain the semver range which might not match the staged version? Would you have to update the `package.json` beforehand? This would then be broken completely as `npm i` would never work?

I like where you're going with the `--stage` flag being required and only working explicitly with the package name - in that case it does make sense to not update the `package-lock.json`, but I wonder if that might not create even more confusion?

Would a restriction of `--stage name` not introduce issues with deduplication and/or transient deps? Not that these can't be solved, just pointing out the possible implications.

I also like the idea of only allowing the authorized people to use the staged version - if this can be done with the initial version, maybe we don't have to solve _all_ of the tarball URL/package-lock problems, as the usage of the feature would be limited and could feed back into the ultimate decisions?

ljharb:
I don’t think a staged version should ever be allowed to exist in a lock file, regardless of who dan install it.

accepted/0000-staged-packages.md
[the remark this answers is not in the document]
isaacs: As this is 100% determined by the registry, there'll be a subsequent discussion/RFC around adding token permissions and TFA requirements other than login/write.

The CLI is agnostic to such things.  It just says "this is who I am" and "this is what I want to do", and it's up to the registry to decide whether to allow it, prompt for a TFA token, or reject it.

For the time being, given the status quo on the registry side, promoting a package requires the same permissions as publishing.  But there's no reason why it necessarily has to stay that way forever, and we all expect to change it, it's just outside the scope of this RFC.

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+  promote packages in staging.
+  * These are separate from the existing read and publish permissions, since we
```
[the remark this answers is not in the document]
isaacs: No, I don't think it is.

isaacs:
If you can install it, it belongs in a lockfile.  There's no reason why we should _not_ lock it once installed, as that can only result in build irregularities.  And if you _can't_ install it, why are we staging it?

Even if staging requires that someone go absurdly out of their way, and do `npm install https://registry.npmjs.org/foo/-/staged/some/kind/of/uuid/foo-1.2.3.tgz`, well, people will just do that.  Let's not make people jump through dumb hoops.  It's the equivalent of disabling copy/paste on a password field.

Requiring a login to install a staged public package is also not reasonable, and bloats the complexity for implementing this pretty significantly.

I'm not ready to throw the baby out with the bathwater here.  Let's not steer so far away from a hypothetical DX problem without at least thinking through how to fix it, and establishing how much of a problem it really is, and providing gentler guardrails to help people avoid it.

Here's what I propose:

- When installing, if `--include-staged` is not provided, then staged versions are never considered.  (This solves like 90% of the DX issue, because you have to opt in to using a staged package in the first place.)
- If `--include-staged` _is_ provided, then staged packages are still lower priority than published packages.  (This solves 90% of the rest; the chances of getting a staged package by accident are effectively nil, and a published version will always be used if possible.)
- When promoting, ensure that the `dist.tarball` path of the staged copy of the tarball continues to be valid and resolve to the final published URL of the tarball (possibly via a 301 redirect).  (This solves most of what's left; if you lock a staged package that then gets promoted, nothing breaks.)
- When overwriting a staged version, the tarball url changes, and the previous staged tarball URL responds with ideally a `410 Gone`, but _may_ respond with a `404 Not Found`.  (And here we have our tiny fraction of cases which will actually affect DX in a negative way.  But it's clear, you opted in, the error message says what to do, and you _probably_ want to take note of this anyway.)  A redirect is not appropriate in this case, because the integrity will be different, so anyone fetching from the previous location is going to get a harder to debug integrity error if we redirect it.

This is a lot simpler, lets the registry be in charge of what's what, and lets the CLI just do its current default behavior with the slight addition of checking `packument.stagedVersions.versions` for matches.  (Ie, just a change to [npm-pick-manifest](https://github.com/npm/npm-pick-manifest/pull/8/commits/5203bafa521ab8965c97df8d541d1d85cd50fabc), but no changes required to any other part of the system, at least for installation.)

The authorization is still just what it always is.  No special auth for special parts of the packument.  If you can publish, you can publish, if you can see it, you can install it.

The resulting data structure could look something like this:

```json
{
  "name": "the-world",
  "dist-tags": {
    "latest": "1.2.3"
  },
  "versions": {
    "1.2.3": {
      "name": "the-world",
      "description:": "is a stage",
      "version": "1.2.3",
      "_npmUser": "publishing-stager",
      "_npmUserPromoted": "promoting-stager",
      "dist": {
        "tarball": "https://registry.npmjs.org/the-world/-/the-world-1.2.3.tgz",
        "integrity": "sha512-deadbeefcafebadholymolywowsuchstage"
      }
    }
  },
  "stagedVersions": {
    "log": [
      "maybe some kind of bookkeeping here?  what versions staged at which time, etc."
    ],
    "versions": {
      "1.2.4": {
        "dist": {
          "tarball": "https://registry.npmjs.org/the-world/-/staged/sha512-content-hash-url-safe-base64/the-world-1.2.4.tgz",
          "integrity": "sha512-content-hash-normal-base64"
        }
      }
    }
  }
}
```

The registry could then have a 301 redirect set up from any `/:pkg/-/staged/:content-address/:file` to the appropriate file.  This wouldn't even have to be done explicitly for staged promotions; any file could be fetched this way, and it could be kept up to date with each file we store.  And, since the url is deterministically different from the final artifact url, we can have different caching semantics around it.  (Or not - the content-address is right in there, it's not like anything _else_ could ever be at that URL.)

isaacs:
If we wanted to go the extra mile for the user, Pacote could detect these 404 errors for tarballs, see that it's a staged tarball URL, re-fetch the packument and update the `resolved` value.

I start to worry a bit that that _might_ be too magical, but otoh, maybe it's exactly the right amount of magical.  In any event, we could explore that with this data and URL structure, and decide later whether it's worth doing.

isaacs:
The net net on this would be:

- Use a staged package: `npm install foo@2 --include-staged` (as long as `2.x` versions are all in staging, like if it's a beta or somethign, and 1.x is the published versions).
- When `foo@2` is promoted, then `npm install foo@2 --include-staged` will pick up the promoted version.
- If the staged `foo@2` depends on a staged `bar@3`, then ok, that'll Just Work, because you're in "include staged" mode.
- When the pkg is promoted and the resolved changes, the package lock doesn't break.  (It _might_ get slightly less canonical or cacheable, but then again, it might not.  Maybe we _want_ to have content-addresses in our tarball urls!)
- If the staged version is overwritten with a new staged version, then the build will break (unless Pacote is taught to just handle this, which may be worthwhile, and would be possible.)

I don't see any DX issues that justify restricting the utility of staged packages, but maybe I'm missing something.

accepted/0000-staged-packages.md
wesleytodd: Now that I am reading this, I feel like one other addition should be made here.  Install `--stage` would only work with a single package and a specific version. Example:

Good:
```
$ npm i --stage foo@1.0.0
```

Bad:
```
$ npm i --stage
$ npm i --stage foo
$ npm i --stage foo@^1.0.0
$ npm i --stage foo@latest
```

Again with the goal that installing from the stage be hard and a very explicit action.

ljharb:
To me, the entire point of the feature is to be able to avoid committing to anything published - and therefore usable - prior to promotion. I want to be able to, for example, have every master build in CI generate a staged version, but have them all *secret* - ie, impossible to even view for non-owners - until an owner selects one to promote.

I would prefer to see "testing a staged version" as a non-goal, since that's what prereleases are for - the value here for me is "i don't want to let anyone but a human do `npm publish`, but i want to be able to do all the prep steps for publishing separately, and without 2FA" (which i believe is also the use case for conventional-release?)

If 2FA is not required to stage something, and staged version can be installed, then that seems like a pretty large potential security hole for anybody using `--include-staged`?

isaacs:
In a case where you have a lot of packages that all work together, I'd really want to have 2 integration tests run before any publish is promoted from staging to release:

1. Run this module's tests against all published versions of the others in the set.
2. Run this module's tests against all staged versions of others in the set.

Otherwise, it's not really an integration test, because you're not testing the integration with current _and_ next.

As for using prereleases for this, yeah, you can effectively get all of this with prereleases.  But part of the goal of this feature is that it lets you minimize version churn (including prerelease version churn in git history).  This allows for something in between "npm install github:user/foo#next" and "npm install foo@next".  _Of course_ there's already ways to do all of this already; there's ways to do everything already (including publishing from CI while still having TFA protection).  They're just less convenient.  I don't see why making staged publishes less convenient is a value.  What harm comes from it?

If we make staged publishes effectively hidden, then we also make it intolerably friction-ful to have the registry decide to temporarily stage publishes that appear to be suspect.  If the staged copy is available behind a flag, then the user is temporarily inconvenienced, but not completely blocked in that scenario.

> If 2FA is not required to stage something, and staged version can be installed, then that seems like a pretty large potential security hole for anybody using --include-staged?

I don't think so.  (a) It's opt-in.  So, yes, you can shoot yourself in the foot by setting it in a .npmrc config or something, but you can also get into very similar trouble installing from a git repo.  And (b), staged package versions would always be deprioritized, so even when opting in, you'd only get the staged copies if they're the only thing that satisfies the dep.  (Allowing a monorepo to publish 100 packages to staging, verify that they all work together, fix whatever problems occur, and then promote them all together.)

isaacs:
Also, 2FA is not required to _publish_ today, so this isn't really much of a reduction in the security stance.  Requiring 2FA for all publishes would be a big step, and I'd get ready for torches and pitchforks as a result.

ljharb:
To clarify; I believe the desire (which i share) is for a package owner to require 2FA for promotion but not for staging, precisely so that it's convenient.

I guess I don't understand why "less version churn" is a value - https://twitter.com/izs/status/603936197941993476

dominykas:
Just to clarify:

1. Someone installs a package with `--include-staged`
2. It gets added to a lock file
3. _Someone else_ runs `npm install` for that lock file (the version is still staged)

Do they get an error (because the package is staged) or does everything proceed as if it's business as usual (because it's in the lock file and integrity matches)? Or do they get a confirmation step or at least a warning?

isaacs:
@dominykas With the approach I'm proposing, the tarball url would get in the lockfile (default behavior today, in other words).  The subsequent `npm install` user would get that staged version (if it's still available and has not been overridden), because fetching the tarball would be allowed.  (npm does not fetch the packument unnecessarily if it can get everything it needs from the lockfile, so it has no way of knowing that it's a staged version; it's just a pre-resolved tarball url.)

This is required to be able to run headless CI jobs against staged versions of things, which is a use case I'd like to be able to support.

isaacs:
> I guess I don't understand why "less version churn" is a value - https://twitter.com/izs/status/603936197941993476

Yes, as we all know, there's no shortage of integers.  But one of the use cases that has been brought up in reference to this feature is being able to more deterministically test and release projects that are composed of many different packages.  Prerelease tags aren't an ideal fit for that use case.

ljharb:
That use case seems somewhat in conflict between the use cases I described; perhaps we could make "it's installable" something that the stager can opt in to? like `npm stage --installable` or something?

wesleytodd:
> I would prefer to see "testing a staged version" as a non-goal

> As for using prereleases for this, yeah, you can effectively get all of this with prereleases.

This!  I was wrong to think this was a good idea, this is what `dist-tags` and pre-release versions are for.  The only exception to this is that a sanity check that the tarball the cli generated is what you expected (for example didn't accidentally `.npmignore` something your package needs).  But this is achievable if only an owner can see/install it.

> What harm comes from it?

The harm is that many other tools and workflows are coupled to today's behavior.  We need to make sure we don't outright break integrations and things like registry proxies.  Once end users can couple to staged versions it means they will start to have workflows which depend on it, and back-tracking later would be very difficult.  

> If we make staged publishes effectively hidden, then we also make it intolerably friction-ful to have the registry decide to temporarily stage publishes that appear to be suspect. If the staged copy is available behind a flag, then the user is temporarily inconvenienced, but not completely blocked in that scenario.

This is a use case I have not seen brought up here before.  Are we sure this should be a goal?  I like the idea in general, but I wonder if it could be served by an explicit feature for security warnings?  Especially with the "install any staged package" behavior, if we used this as a work around it would mean folks opening the flood gates just to get a sip of water.  For example, `package-a` has a auto-staged security warning which the user knows is invalid, so user does `npm i --included-staged` but didn't realize that `pacakge-b` also has one which is really a vulnerability.  I like the idea, but I think we need to think through the ramifications before including it as a goal in this feature.

> I guess I don't understand why "less version churn" is a value

Agreed, especially in pre-release versions.

>  "it's installable" something that the stager can opt in to? like npm stage --installable or something?

While this would get us out of the alignment bind we might be in today, I think this would be a mistake because it would let users opt into using this feature in a way which is less optimal than pre-release and `dist-tags`.

ThisIsMissEm:
@evocateur I believe this RFC may help with the "one shot" publish issue we had in lerna, where we switched to using a tmp dist-tag.

evocateur:
@ThisIsMissEm It's an interesting variation, to be sure. I think for the most part, `dist-tags` are still the way to go to ensure the unity of a given release. There is certainly tooling to improve in terms of _recovery_ from partial releases, especially the "promotion" of the temporary tag once all the (previously) failed packages have been successfully published.

wesleytodd:
> I believe this RFC may help with the "one shot" publish issue we had in lerna, where we switched to using a tmp dist-tag.

I think this is specifically a design goal, to support this use case, although it is not called out as such in the RFC.  Maybe it should be added?

isaacs:
While there are some potential footguns avoided by adding the restriction that only package owners can load staged packages, it adds significant complexity on both the server and client implementations, and (depending on how it's implemented) potentially adds a few even more concerning footguns.  I hope, in this message, to make it clear that is not feasible, so we can drop it from this discussion.

Let's say we do add the restriction that staged packages cannot be installed by non-owners.  This can be implemented (as far as I can tell) in 5 different broad strategies, all of them with profound challenges.

1. Strictly a client-side restriction.  The CLI just won't select packages for installation from the `stagedVersions` set unless the current user is in the `maintainers` list.
2. We exclude the `stagedVersions` object from the packument if the fetching user is not in the `maintainers` list.
3. We block the download of a staged tarball URL if the user is not in the maintainers list.
4. We use a different packument target entirely, and in order to download staged versions (even for inspection), one has to use a different packument endpoint.
5. Staged published artifacts do not exist in any packument at all, and some other out-of-band means would be implemented to find out about them.  (This is really a special case of 4, where "other packument target" might be a completely different non-packument endpoint.)

(1) is, of course, not a real restriction in any sort of security sense, but that might be fine.  Ie, a staged package isn't really securely _hidden_, let's say, just restricted by a handshake agreement that prevents the CLI from trying to fetch it unless the user is an owner.  Leaving aside the question of whether other npm clients would enforce the same restriction, the bigger challenge comes about when we consider that the `npm-pick-manifest` module would have to know, at install time, your npm username.  Without making a fetch to the registry, it doesn't know that.  We only store an opaque bearer token that tells the _server_ who you are.  By design, for security, your npmrc file does not contain your username.  In order to get the username from a bearer token, you have to hit the `/-/whoami` endpoint.  (This way, npmrc files containing formerly-valid logins don't leak any information if the token is revoked.)

Having to hit a registry endpoint to be able to even select a package version for installation is pretty crummy.  Authed requests are slower and more costly, so we try to minimize them as much as possible.  Any unscoped packages bypass that part of the registry stack entirely, and the auth stack for scoped packages has a lot of shortcuts to quickly jump from a bearer token and action to an answer about whether that's allowed.  What shows up in the `maintainers` list is a cached projection of the actual authorization set.  (Also, we'd have to always fetch in `fullMetadata` mode, because corgis don't list maintainers.)

The fact that it's opt-in would help with the perf _somewhat_, but there's still the question of maintenance burden and complexity, since npm-registry-fetch now depends on more than just the packument as a discrete object.  And once the staged package tarball url is in a lockfile, we'll have to track that it's staged (and, when doing that install, whether the user has the right to download that thing or not).  It makes my head spin trying to work out the edge cases in this.  If someone else wants to take it on in a fork or an alternative client, I'll be curious how it works out, but I'm not doing this in npm/cli.  Too hard.

(2) imposes similar complexity, but on the server side.  Remember that auth stack that we try to hit as little as possible?  Now we need to keep 3 forms of the packument, and check the auth state on every packument request.  I don't primarily work on that portion of the stack, but I'm confident that it would impose significant burden on us to do that, effectively making every package as costly to serve as a private package.  (Which, on each individual fetch, is not a huge cost, but multiplied by ~70Bn package downloads a month, adds up quick.)  So that's not gonna happen any time soon.

(3) suffers from essentially the same performance concern as (2), but on tarball urls instead of packument urls, which is potentially worse.  The mitigating factor here that might make it less infeasible is that the staged tarball URLs themselves could have a discernible pattern (which is good for other reasons anyway), so we'd save having to make the check on _every_ fetch.  However, this one has the "even worse footgun" I alluded to.  If you install something, and get a tarball url in your shrinkwrap, then someone else (who isn't a maintainer, or just isn't logged in) tries to download it, they get a 404.  This is exactly the DX issue that got us down this road to begin with.

(4) and (5) -- ie, create a net new API surface for staged publishes -- are probably the least infeasible from an operational point of view, but also will take the longest to spec and deliver, and frankly, might never actually happen.  Also, it's worth noting that "staged publishes that only a select few can install, from a completely different endpoint" is exactly what private registries provide.  And, if the new registry endpoint for staged publishes _isn't_ something you can `npm install` from (ie, option (5)), then it's pretty useless.  The number one way to inspect a published npm package is to install it and run tests against it.  (Ideally in CI, where a matrix of platform and engine versions can be tests in parallel.)

In all but (1), "staged publishes that only a select few can install" becomes a backdoor way to have a mix of private and public code at the same namespace entry, which can exist for both scoped and unscoped packages.  If we were ever to offer something like that, we'd charge for it, because that's difficult to implement, costly to operate, and primarily serves the interests of corporate users.  And (1) is a ton of work for me and my team that I'm not eager to have us do.  Assuming that the public registry makes staged versions of public packages available to all, nothing's stopping any _other_ registry implementation from doing this, of course.  I'd love to have some data (or at least, anecdata) about how that works out.  It might still not be a compelling enough case to get over the challenges at npm public registry scale, but it'd be _more_ compelling than _not_ having it, for sure.

I'm sorry.  I don't think that's going to happen.

The good news is, just reducing the priority of staged versions below that of published versions solves these problems exceedingly simply, and gets rid of most of the hazards.  And an implementation exists that supports it today, in the latest versions of `npm-pick-manifest`, `pacote`, and `@npmcli/arborist`.  It pretty much impossible to get a staged version by accident.  You have to both opt into considering staged versions, _and_ there has to be no published version that satisfies the requested range.  This prevents the "floodgates to get a sip of water" issue entirely, because those versions will never be chosen if there is any other option.  (It's like deprecation, but stronger.)

If "testing a staged version" is a non-goal, then I don't think this is worth doing, honestly.  Why would you ever stage something, if not to test it before committing to it being published?  Isn't that the whole point?

Here's what I'd like to focus on, which I see as the main remaining obstacles to this:

- The `dist.tarball` url of staged packages.  I like the content-address idea, could be long-lived even when the "real" artifact url is used moving forward.
- The handshake required to promote from staged to published.  I'm thinking, I don't wanna have a situation where a bad actor stages a new thing right as you're about to promote it.  The workflow has to be worked out there, though, because what you really want is to say "I want to promote this thing that I got 15 minutes ago", and you're unlikely to grab the integrity as you're installing it initially for inspection.  Maybe provide the artifact itself and we check that it matches?  What's the UX for "I'd like to start testing staged version XYZ, I'll promote it once I'm satisfied"?

ljharb:
You tested it before staging it, so no, imo that's not the whole point - the point is to separate "prepare the publishable artifact and securely transmit it" from "propagate it publicly, which requires extra auth".

isaacs:
> You tested it before staging it, so no, imo that's not the whole point - the point is to separate "prepare the publishable artifact and securely transmit it" from "propagate it publicly, which requires extra auth".

But you're explicitly saying that "transmitting" it is _not_ done securely.  It doesn't require TFA to stage a package.  (And, it's likely a token that far more people have access to.)  Then, later, I provide TFA and a more restricted access token to promote it from staged to published (the secure step).

So I'm supposed to just put my stamp of approval on something without inspecting it first?  That doesn't seem like a good idea.

isaacs:
> I believe this RFC may help with the "one shot" publish issue we had in lerna, where we switched to using a tmp dist-tag.

If I'm understanding you correctly, this is about the case where you have a bunch of packages you want to publish all together in a sync'ed version?

This would actually be slightly safer than using a temp dist-tag, because you wouldn't have the race condition period where mismatched versions are published and installable.

Granted, the version referenced by the `latest` dist-tag has long been prioritized, so if `1.2.3` is on `latest`, and you've soft-published `1.2.4` on the `tmp` dist-tag, then anyone fetching `1.2.x` will get `1.2.3` and not `1.2.4`.  And, if you're _not_ installing against `latest`, and it's a tmp dist-tag floating on top of `next` or something, then you won't have that coverage, but obviously way fewer people will be fetching that version range in the first place.

It does bring up an interesting point about the promotion flow, though.  Should it be somehow transactional across multiple packages and versions?  How much of a race condition is acceptable in this case?  Even if there is a window of mismatch, it's going to be much smaller than with actually uploading a bunch of tarballs, so maybe it's fine?

ljharb:
That's fair, and that's a reason why it's reasonable to download the tarball - but that doesn't have to mean that the only way to download the tarball is to `npm install` it, or that it has to be enmeshed in the lockfile/install/packument ball of complexity.

In other words, while I can concede that it is important to inspect the artifact before promoting it, i don't think the standard install flow - flag or not - is the appropriate place to do it.
----------------
In your example, what happens if i've staged 1.2.4 and then i end up staging 2.0.0 and publishing that, and never publishing any 1.x above 1.2.3? What happens to those depending on the staged v1.2.4?

wesleytodd:
I personally don't care so much about enforcing the installing user is a maintainer constraint.  That said...

> That's fair, and that's a reason why it's reasonable to download the tarball - but that doesn't have to mean that the only way to download the tarball is to npm install it, or that it has to be enmeshed in the lockfile/install/packument ball of complexity.

I think @ljharb is very correct here.  Avoiding all of the footguns and the complexity of lockfiles etc is a much better option.  Which goes back to my proposed additions:

> 1. The --stage flag on install only works when installing a single package
> 2. Installing from --stage is not reflected in the package.json or package-lock.json

Are these still viable additions?

isaacs:
> In your example, what happens if i've staged 1.2.4 and then i end up staging 2.0.0 and publishing that, and never publishing any 1.x above 1.2.3? What happens to those depending on the staged v1.2.4?

So, iiuc, the flow is:

- publish ..., 1.2.2, 1.2.3
- stage 1.2.4
- someone does `npm i --include-staged foo@1.2.4` (they'd have to specify a version; even with `--include-staged`, real published versions will be prioritized)
- stage 2.0.0
- prompte 2.0.0

In the current v7.0.0-beta with @npmcli/arborist and npm-pick-manifest status quo, the user who installed the staged 1.2.4 will have `https://registry.npmjs.org/foo/-/sha512-asdfasdfasdf/foo-1.2.4.tgz` in their package-lock.json, and they'll just keep on fetching it.

One potential issue to be noted here is that their _package.json_ will save `^1.2.4` in the dependencies object, so anyone installing _their_ package will also have to use `--include-staged` or get an `ETARGET` error.  I'm inclined to say this is a minor/acceptable footgun, since staged deps will be unlikely to be used in this case anyway.  (And it's somewhat easily worked around; we could even pretty easily add a warning printed when installing a staged dep to say that publishing this package won't work until the staged dep is promoted.)

One nice feature about this vs floating a git dep is that there's no added work to switch to the "real" version once it gets promoted.  The dep will just start working without the `--include-staged` flag.  (You'd still need to float a git dep if waiting on a PR or release, of course.)

> In other words, while I can concede that it is important to inspect the artifact before promoting it, i don't think the standard install flow - flag or not - is the appropriate place to do it.

I mean, technically speaking, if we give people access to the staged tarball, we are giving them access to publish it.  It feels a little overly paternalistic (not to mention more work for the CLI and registry) to require it to be a radically different flow.

I do think "install the package" is a common way that you'll want to go about inspecting a staged publish.  In my own experience with situations where I want to get someone to poke at a possible publish, I'm most often saying something like "Can you `npm i isaacs/foo#some-branch` and see if that fixes your problem?"  It's pretty inconvenient if I'd have to download the artifact, put it somewhere, then ask them to fetch it from there, etc.

> 1. The --stage flag on install only works when installing a single package

(I'd rephrase this from "only works when installing a single package" to "only works for the packages explicitly listed] in command line positional arguments".  I'd assume that if I do `npm i foo@1.2.3 bar@2.3.4 --include-staged` it should give me staged versions of both foo and bar.)

This is a tricky one.  It's not hard to _do_, necessarily (we'd just drop that from the options when resolving for any package that isn't in the `explicitRequests` list), but it does cut off the capability to test multiple staged deps all together, and makes the package.json issue worse.

What if I do actually require something that's only in a staged version?  And I want to test _that_ thing I'm doing in some other app?

For example, let's say that nyc stages a v16.0.0 release.  It has some new features, and it's a big release, so I want to take some time and run a battery of analysis and tests as I'm developing the next major of tap that depends on it, and I want to do that in parallel with the nyc stage acceptance process.

It'd be nice if I could install the staged version, and explicitly save it as a dep.  Then, do my work, get it where I want it, but I want to play test it in an app that needs some of that new functionality.  So, I go over to that app, and run `npm i tap@16.0.0 --include-staged`.  Should it be unable to fetch the staged nyc version?  If so, we're asking people to think about metadeps, and it's a long-standing fundamental design principle that npm _doesn't_ ask people to do that as a general rule.

> 2. Installing from --stage is not reflected in the package.json or package-lock.json

This is much easier.  We could say that `--include-staged` implies `--save=false` by default.  I would want to split these two concerns apart, though, just for clarity.

Saving to the package.json automatically, yeah, that's a footgun.  Because, then if I publish my package, and it's _not_ staged, you still would have to do `--include-staged` to get my dep, which might not be what I want.  (I should still be _able_ to explicitly save it, by specifying `--save` on the command line.)

Not saving to the package-lock.json seems like a much bigger step to me.  If the URL is accessible, and includes the content-address, then there's really no harm in letting it be there, and I'd be worried about quietly getting into a state where the lockfile doesn't reflect the files on disk, so devs are working on different systems without realizing it, the integrity is harder to inspect, etc.  If it turns out that a staged publish _is_ compromised, checked-in lockfiles are a pretty good way to detect the extent of the damage (just based on what we've seen with tracking down malware impact in published packages).

coreyfarrell (commented):
@evocateur if this goes though do you see this as something which could be worked into the lerna workflow?  One of the major problems I have with monorepo's is that it is difficult to test changes to monorepo packages from outside the monorepo (cannot npm install from git).  Currently this requires publishing to a `next` tag which can be undesirable if you're not actually ready to publish.

CC @bcoe since this might be of interest to your release bot.

EDIT: apparently I should have paid better attention, I went straight to the RFC file and initially missed the discussion outside comments that were made on the files.  Sorry if any of my comments seem to ignore previous discussions, that wasn't on purpose.

coreyfarrell:
> If unauthed people can install a staged version, they can depend on it, and staging serves no purpose.

I think it would be smart for `npm install` to produce a warning if package.json referenced a staged package.  If staged packages cannot be `npm install`'d then it would not be useful for me (though I acknowledge others would find it useful for publish bots).

evocateur:
@coreyfarrell Yes, if this is implemented I'm sure Lerna would be taught to follow suit eventually. The current status quo (dist-tags) has rough edges, but there are bigger fish to fry (from Lerna's perspective) in terms of automated CI publishing, which is currently very, very painful.

evocateur:
> It does bring up an interesting point about the promotion flow, though. Should it be somehow transactional across multiple packages and versions? How much of a race condition is acceptable in this case? Even if there is a window of mismatch, it's going to be much smaller than with actually uploading a bunch of tarballs, so maybe it's fine?

@isaacs The ideal case would indeed be transactional across _n_ packages at explicit versions. I agree race conditions aren't a big concern in this case (even when it's a less-ideal "iterate over the list of packages+versions and make individual API calls to promote out of staging" approach).

wesleytodd:
> (I'd rephrase this from "only works when installing a single package" to "only works for the packages explicitly listed] in command line positional arguments".

Yeah better phrasing and behavior!

> For example, let's say that nyc stages a v16.0.0 release. It has some new features, and it's a big release, so I want to take some time and run a battery of analysis and tests as I'm developing the next major of tap that depends on it, and I want to do that in parallel with the nyc stage acceptance process.

Wouldn't you just have to explicitly add that to your list?  `npm i --include-staged tap nyc`?


>  If the URL is accessible, and includes the content-address, then there's really no harm in letting it be there

> One potential issue to be noted here is that their package.json will save ^1.2.4 in the dependencies object, so anyone installing their package will also have to use --include-staged or get an ETARGET error.

I think this is the harm and why I don't think it should go in the `package.json` or `package-lock.json`.  This is going to happen to everyone who uses this, they will accidentally push it and break everyone else working on the project.  Think about a monorepo at a company with 100 devs working in it.

> checked-in lockfiles are a pretty good way to detect the extent of the damage (just based on what we've seen with tracking down malware impact in published packages).

I see your point here, but I disagree on the impact.  If a malicious package is staged and someone installs it, they are already compromised and their `package-lock.json` is not trustworthy.  This is specifically why I think we should make it hard for folks to install staged versions in any workflow other than a CI test before promote.  

> What would npm i --stage foo@1.0.0 do to package.json?
> I think it would be smart for npm install to produce a warning if package.json referenced a staged package.
> If staged packages cannot be npm install'd then it would not be useful for me 

@coreyfarrell Would it be acceptable with what I proposed?  It is only installable via explicit command but did not modify `package.json` or `package-lock.json`?

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+certainly implement a useful staging workflow based on tooling that already
+exists. Also, even if we start by adding this to the CLI, it doesn't mean we
+can't build a more complex application on top of the CLI later on.
+
+After considering these alternatives, we belive a staging workflow built into
+the npm CLI provides the shortest path to useful staging functionality that we
+can iterate upon. It will allow us to test these new concepts in a tool with
+which we and our community are both familiar.
+
+## Implementation
```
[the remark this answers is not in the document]
isaacs: It wouldn't be too hard to just make `npm unpublish` allowed for any version that is staged.  (Ie, `unpublish` would remove from both `versions` and `stagedVersions.versions`, and the registry would allow removal from `stagedVersions` without any restriction on dependents or publish time.)

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+Anything published by CI is instantly available for general use, which isn't
+ideal when a human is required to be in the loop. And even when fully automated
+CI/CD is allowed, the inability to use 2FA with auth tokens is a security risk.
+
+## Detailed Explanation
+
+We need a minimum of four things to solve these problems:
+
+1. Staging area on the registry invisible to most `npm` commands
+2. Auth tokens that can be scoped to the staging area
```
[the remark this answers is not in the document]
isaacs: In some of the conversations that happened before the initial RFC (but which I don't see reflected in the RFC here) the assumption was that either a specific token could be restricted to only doing staged publishes in general, or a user/team could be granted access to only stage in a given package.

The complexity around that part is a bit orthogonal to the rest of this, however, and so for now, I think it's best to leave it aside.  For the purpose of limiting the scope of _this_ RFC, we can assume that users who are maintainers can both stage, publish, and promote, since it's up to the server to allow or reject any given action based on ACLs and token metadata.

isaacs:
> > For example, let's say that nyc stages a v16.0.0 release....
> Wouldn't you just have to explicitly add that to your list?  `npm i --include-staged tap nyc`?

That imposes an extra burden on the user, though.  Remember, they'd have to do `npm i --include-staged tap@16.0.0 nyc@16.0.0`, and they might not even be aware that nyc is a thing.  Also, what if I also staged new versions of tcompare, tap-yaml, treport, and libtap, because we're doing this all in a monorepo, and I've just staged everything for the next major release?  Now for me to test out the new tap in any other project, I have to list out its whole dep family, or else the install will fail.

This might seem a bit extreme, but I do actually maintain some things with dozens of interdependent dependencies, and I don't think it's all that unusual.  Consider if lodash or babel wanted to do the same thing.  For internal testing, this is taken care of by keeping deps together in a monorepo or using git repo deps, but doing integration testing with the staged flagship project would likely entail a lot of headaches.

We are currently going through this exact scenario with the npm v7 beta today.  There's been 4 major version production-published releases of npm-registry-fetch, just because each time we find some new issue, it means we have to semver-major the whole world all over again.

isaacs:
> @isaacs The ideal case would indeed be transactional across n packages at explicit versions. I agree race conditions aren't a big concern in this case (even when it's a less-ideal "iterate over the list of packages+versions and make individual API calls to promote out of staging" approach).

@evocateur That's good to hear!  Transactional promotion of multiple packages would definitely impose significant technical challenges on the registry side, since it would mean that promoting packages could not be a stateless operation on a single packument.  (Right now, publishes get thrown in a queue and processed by a worker that can update files and CDN caches, and we can scale it horizontally very easily when traffic spikes, because no one publish depends on any other, and they're each idempotent as long as they happen in order within a given package.)

isaacs:
On saving...

### package.json

To recap, the argument against saving to `package.json` by default is that it can lead to a situation where a package gets published, but can only be _installed_ by opting into the staged version of the dep.  (Either by explicitly listing the dep, which you don't care about, or by passing `--include-staged` on the command line to allow any staged versions.)

This is somewhat compelling to me, as well, but I'd still like to explore some of the other effects of this, and just for thoroughness, bring up some ideas for how it might be addressed even _if_ it's saved (including because the user specified `--save`, which as an explicit user opt-in, should definitely be respected).

### package-lock.json

I'm very concerned about not saving something to the shrinkwrap that we intentionally install, as that creates whole new footguns and areas for inconsistency, and is not straightforward or elegant from an implementation point of view.  (Which ime means: we'll regret it in a year when we are still dealing with the bug factory.)

- Arborist gets the `resolved` and `integrity` values from the package-lock.json.  So we'll have to make sure we store that somewhere so that we know what is (or ought to be) in node_modules.  (It will be stored in `node_modules/.package-lock.json`, which addresses this point somewhat.)
- The contract of the `package-lock.json` is that it allows devs working on a project to ensure that they're all working on the same state at the same time.  If the user decides not to use a package-lock at all, that's one thing.  But if there's a package-lock there, and we don't keep it in sync, that feels like we're violating that contract.
- It makes it harder to do ongoing work in a branch against the not-yet-actually-published code, getting back to the cascading semver-majors in small modules problem [described above](https://github.com/npm/rfcs/pull/92#issuecomment-592330946).
- If you install a staged package, and then install anything else, it'll prune or replace the staged package.  (Because it's present, but not in the ideal tree!)

### saving in package-lock.json but not package.json

If we save a staged dep install in package-lock but _not_ in package.json, then we can quickly run into a case where the two are out of sync, resulting in an ideal tree that doesn't satisfy the deps listed in the root package.json.  We could conceivably work around this, but it'll be hacky at best.

One approach might be to put `staged: true` (or some other explicit indication) in the package-lock.json entry, so that we know that whatever version it is, is provisional, and should be accepted (or warned about), but not treated as an invalid dependency to be fixed.

I still hate this, because data inconsistency always leads to problems, but I hate it a lot _less_ than just not saving to package-lock at all.

### other ideas

Zooming out a bit, the _problem_ statement is "I do not want users of my package to have to opt into the staged version of one of my package's deps in order to install my package".

We're all jumping from there to "I do not want to publish a package that depends on a staged version of a dep", but I think that might be prematurely assuming an implementation.

For example: as a package publisher, I really _want_ to stage versions of npm-registry-fetch, make-fetch-happen, pacote, and cacache.  From there, I want to pull them all into staged versions of libnpmpublish, npm-profile, etc., and pull those into the npm/cli project to work on while we iron out all the kinks.

#### bad idea: block publishes if a dep is staged

On the registry, do not allow a publish or promotion if a package's dependency ranges can only be satisfied by including staged versions.

Problems with this:

- publish workflow gets significantly slower and more complicated
- if you stage a set of cyclical deps, you'll have to UNpublish one/some of them, then re-publish (skipping staging) to break any cycles, before promoting the others.  That sucks and is not obvious.

#### maybe better idea: just lean on a warning

So here's a thought: we save in both places, just like normal.  But when we publish (or even stage), we check to see if anything in the tree is a staged version of a dep, and do this:

```
$ npm publish

Warning! The dependency at:
  node_modules/@foo/bar
is a staged publish that has not been promoted.

In order to install this project (@baz/beep), your users
will need to run with `--install-staged` set.

(Silence this warning by running `npm promote @foo/bar@1.2.3`)

Proceed?  (N/y)
```

dominykas:
> bad idea: block publishes if a dep is staged

What if publish can be unblocked with npm publish --include-staged? Same flag, works. both ways?

Which kind of leads me to another concern - should it be disallowed to use --include-staged via .npmrc?

wesleytodd:
> That imposes an extra burden on the user, though.

Yes, intentionally.  If you are testing with staged versions you should have to be very explicit about it IMO.  And if they want a staged version of any package they should have to be aware of it and opt-in explicitly.  Otherwise we introduce a whole new class of "it worked on my machine" issues when folks implicitly test with staged versions.

> Consider if lodash or babel wanted to do the same thing.

I think a simple `npm i --include-staged $(npm workspace ls)`, `npm workspace i --include-staged` or similar command could be constructed.  I think this is for sure a use case we should consider, but there are other ways to support it directly than making `npm i --include-staged` work on a whole tree.

> including because the user specified --save, which as an explicit user opt-in, should definitely be respected

This I can agree on, if the user explicitly opts-in it should be respected.  In this case I would see the `package.json` and `package-lock.json` being updated accordingly.

> We're all jumping from there to "I do not want to publish a package that depends on a staged version of a dep"

While I am worried about publishing, I am much more worried about pushing it to a git repo.  In an application, this would break all of your teammates and builds.  This is specifically why I am pushing against this, **for application developers workflows**.

> Which kind of leads me to another concern - should it be disallowed to use --include-staged via .npmrc?

Yes!  This would be a huge debugging issue if folks started having it in `.npmrc`.

ruyadorno:
Note from OpenRFC call:
- This topic needs a [deep dive](https://github.com/npm/rfcs/issues/105)

darcyclarke:
I've created an agenda issue [here](https://github.com/npm/rfcs/issues/116). Feel free to queue up ideas for implementation details/topics we'd like to cover in the "Deep Dive" call next week.

Also, definitely let us know if you don't think you'll be able to make the *Wednesday 2pm ET* day/time.

isaacs:
Ideas from deep-dive, which need to be incorporated into the RFC.

- autoexpire after a month/week/days/hours/??

    This can be entirely a server-driven policy, of course.  But the RFC should consider the merits of a longer/shorter/configurable limit.  Takeaway from discussion was that 1 day (or less) is optimal to prevent a staged version from effectively being a forever dependency (and abuses that could circumvent around the "one version, one publish" rule).  However, less than a week makes it tricky to get downstream OSS users to try out a prerelease build.

- `--include-staged=foo` to allow staged versions of `foo` and all deps down that path of the dependency graph.  This was a compromise to the "must be installable" use cases, and the "must not be accidentally installable" concerns.

    As a possible enhancement, if no staged versions were resolved as a result of the config, it should warn that it was unnecessary.  (This is a bit tricky, since it means tracking which versions were resolved to staged versions, which npm-pick-manifest doesn't expose, and isn't particularly easy to get at in an elegant way, so I might push back on that suggestion, at least for the time being.)

- Seems like we're settled on the content-address style tarball URL, since it's something that can be stored in a lockfile without (a) breaking once the package is promoted, or (b) requiring that it not be permanent and cacheable.  (CA urls can also be enabled for all packages relatively easily, so it'd keep on working once the tarball is promoted.)  Open question whether this should be a 301 redirect, or just another URL referencing the same content.

    When a new artifact is staged at the same version as a previously staged artifact, the overwritten tarball should be deleted from the registry and removed from cache.

- We didn't get into the "security detection" use case of staged publishes, but that should also be added to the RFC.

    Essentially, this would allow a registry to run a set of tests against an incoming (non-staged) publish, and if it detects something fishy (private keys, malware, obfuscated code, bitcoin miners, what have you), it could choose to stage the version rather than publishing it, and push a notice telling the user what's up and how to install it.  Then, after some time period, once it's vetted by a human (or some other more advanced system), it could be automatically promoted on the users' behalf.  (Or deleted if actually a problem, along with other remediations as appropriate.)

    I don't think it changes any of the concerns discussed already, but it should be called out specifically.  (And, if there's no way to install staged versions, then it's pretty much impossible to do this kind of automated quarantining in a way that isn't massively disruptive.)

ThisIsMissEm:
Quickly replying on:

> Open question whether this should be a 301 redirect, or just another URL referencing the same content.

I think it'd make sense to have a single URL for tarballs, however, if possible add a response header of something like "see other version". This keeps with content addressing, without breaking too much. 

I do think the tarball name should be the content address, and the folder the package name, followed by the version, so: `/foo/1.2.5/<content address>.tar.gz`, but I feel like that's probably a larger discussion for another RFC. 

I think it'd also be a really good idea to adopt the IPFS content addressing scheme, if we're going to do that. As this would allow for future p2p installs of packages using content addressing, so only packument is needed from npm.

arslan-raza-143 (approved):

endless61 (approved):

nateghia219 (commented):
Update systems to help you get

Bonkers10 (approved):

TokensDL:
djsauble:staged-packages

x157660503 (approved):
- [ ]

Hirram (approved):

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+* Staging in third-party repository managers, like Nexus Repository Manager,
+  also meets any gatekeeping requirements.
+  * In Nexus 2, this was implemented via dedicated staging repositories, where
+    packages could be published and then promoted to a production repository,
+    all via a nice staging UI.
+  * In Nexus 3, they shifted to an API-first approach, where new packages had
+    staging metadata set on them, and the API could enable CI and other tooling
+    to stage or promote packages by updating this metadata.
+
+## Unresolved Questions and Bikeshedding
```
Romeroal200, line 84: E

dominykas:
Is there any chance this is going to happen? Should we just close it to avoid having it as a bot spamhole?

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
...
+
+* Currently out of scope (but up for debate):
+  * Webhook support
+  * Metadata related to fitness of staged packages
+  * Multiple staged packages (current spec has a limit of one, successive stages
+    clobber the previous staged version).
```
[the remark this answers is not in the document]
moduller: npm unpublish

<details>
<summary>Details</summary>

```
npm unpublish

```
`npm unpublish`
`npm unpublish *`
`npm unpublish stagedVersions
</details>

moduller (approved):
`npm unpublish`

kelset:
cc @darcyclarke 👆 so many bots spamming in this RFC 😥

Xcalibur53 (approved):
[]()

Metlina3:
> RFC'ye bakın

113ll3 (approved):
Djsaudible:staged-packages

accepted/0000-staged-packages.md
113ll3, line 1: Accepted/0000-staged-packages.md

113ll3 (commented):
Hi

accepted/0000-staged-packages.md
```diff
@@ -0,0 +1,90 @@
+# RFC: Staged Packages
```
Blg21, line 1: @djsauble

Blg21 (approved):

0137248506:
> RFC'ye bakın

Nmmqkwed,mwm,ewqmroamq. Oewozoq

Thapeloza (approved):

Abdulqadirasif:
[Uploading Connected_Devices_Best_Practices.pdf…]()

RTapio9912 (commented):
Hey RHT9912

accepted/0000-staged-packages.md
RTapio9912, line 1: Review comment

RTapio9912 (commented):
Review

Amrfawzy9:
> Ideally there'd be more information than just "version", "package.json", and "tarball" - git sha, in particular, would be helpful?

{
  "stagedVersions": {
    "whatever": "other",
    "metadata": "we wanna",
    "put": "here",
    "versions": {
      "1.2.3": { "manifest": "data..." },
      "1.2.4": { "another": "staged manifest..." }
    }
  }
}

0xsarawut (approved):
accepted/0000-staged-packages.md

CyberNFT (commented):
+

sin-strokes (commented):
[]()

accepted/0000-staged-packages.md
hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmaol.com

accepted/0000-staged-packages.md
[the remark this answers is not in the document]
hoangnhan7991-boop36: djsauble:staged-packages
hoangnhan7991@gmail.com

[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

accepted/0000-staged-packages.md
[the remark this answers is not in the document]
hoangnhan7991-boop36: hoangnhan7991@gmail.com

accepted/0000-staged-packages.md
hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7901@gmail.com

hoangnhan7991-boop36: hoangnhan7991@hmail

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991@gmail.com

hoangnhan7991-boop36: hoangnhan7991-boop36

hoangnhan7991-boop36:
> In a case where you have a lot of packages that all work together, I'd really want to have 2 integration tests run before any publish is promoted from staging to release:
> 
>     1. Run this module's tests against all published versions of the others in the set.
> 
>     2. Run this module's tests against all staged versions of others in the set.
> 
> 
> Otherwise, it's not really an integration test, because you're not testing the integration with current _and_ next.
> 
> As for using prereleases for this, yeah, you can effectively get all of this with prereleases. But part of the goal of this feature is that it lets you minimize version churn (including prerelease version churn in git history). This allows for something in between "npm install github:user/foo#next" and "npm install foo@next". _Of course_ there's already ways to do all of this already; there's ways to do everything already (including publishing from CI while still having TFA protection). They're just less convenient. I don't see why making staged publishes less convenient is a value. What harm comes from it?
> 
> If we make staged publishes effectively hidden, then we also make it intolerably friction-ful to have the registry decide to temporarily stage publishes that appear to be suspect. If the staged copy is available behind a flag, then the user is temporarily inconvenienced, but not completely blocked in that scenario.
> 
> > If 2FA is not required to stage something, and staged version can be installed, then that seems like a pretty large potential security hole for anybody using --include-staged?
> 
> I don't think so. (a) It's opt-in. So, yes, you can shoot yourself in the foot by setting it in a .npmrc config or something, but you can also get into very similar trouble installing from a git repo. And (b), staged package versions would always be deprioritized, so even when opting in, you'd only get the staged copies if they're the only thing that satisfies the dep. (Allowing a monorepo to publish 100 packages to staging, verify that they all work together, fix whatever problems occur, and then promote them all together.)

https://github.com/npm/rfcs/pull/92#issuecomment-587163036

oliviamazzotta47-lab (approved):

PR #96, opened 2020-01-31
RFC: Add publish confirmation prompt

ruyadorno:
`npm publish` should have a confirmation prompt in order to provide a chance for users to review the package contents, similar to how 2FA-enabled users are able to review the contents prior to inputting their OTP.

[See RFC](https://github.com/npm/rfcs/blob/8129a02baa03d1bd4f0f9f9ee98d3f7de0b4e0b6/accepted/0000-publish-prompt.md)

/cc @mikemimik

### DIFF
--- a/accepted/0000-publish-prompt.md
+++ b/accepted/0000-publish-prompt.md
@@ -1,4 +1,4 @@
-# Publish prompt
+# Publish confirmation prompt
 
 ## Summary
 
@@ -18,7 +18,7 @@ A broader population of package authors can benefit from this UX if we formalize
 
 ### Feedback from the discussion around this RFC
 
-- Should be implemented behind an opt-in/experimental flag for npm@7
+- Should be implemented behind an opt-in flag for npm@7
 - Non-TTY environments should skip the prompt.
 
 ## Alternatives
@@ -27,10 +27,10 @@ A broader population of package authors can benefit from this UX if we formalize
 
 ## Implementation
 
-Prompts the user for confirmation when using the experimental flag:
+Prompts the user for confirmation when using the opt-in flag:
 
 ```sh
-$ npm publish --experimental-publish-confirmation
+$ npm publish --publish-confirmation
 
 npm notice
 npm notice 📦  disparity@3.0.0
@@ -60,7 +60,7 @@ Do you wish to proceed? [y/N]
 The user can also provide a `--yes` option that can skip the prompt and keep current behavior:
 
 ```
-$ npm publish --experimental-publish-confirmation --yes
+$ npm publish --publish-confirmation --yes
 ```
 
 ## Prior Art
--- a/accepted/0000-publish-prompt.md
+++ b/accepted/0000-publish-prompt.md
@@ -16,17 +16,21 @@ Currently "publish" has a very nice **unintended UX** for users with 2FA enabled
 
 A broader population of package authors can benefit from this UX if we formalize it. Allowing for review of contents and cancelation of package publishing.
 
+### Feedback from the discussion around this RFC
+
+- Should be implemented behind an opt-in/experimental flag for npm@7
+- Non-TTY environments should skip the prompt.
+
 ## Alternatives
 
-1. Not change it, just keep the current behavior?
-2. Implement it behind an _opt-in_ config option
+1. Not implement it, keep the current behavior.
 
 ## Implementation
 
-This is a breaking change from the current `npm publish` behavior, it would prompt the user for confirmation:
+Prompts the user for confirmation when using the experimental flag:
 
 ```sh
-$ npm publish
+$ npm publish --experimental-publish-confirmation
 
 npm notice
 npm notice 📦  disparity@3.0.0
@@ -56,13 +60,9 @@ Do you wish to proceed? [y/N]
 The user can also provide a `--yes` option that can skip the prompt and keep current behavior:
 
 ```
-$ npm publish --yes
+$ npm publish --experimental-publish-confirmation --yes
 ```
 
 ## Prior Art
 
-Currently `npm publish` will automatically publish the tarball to the registry in case the user does not have OTP enabled.
-
-## Unresolved Questions and Bikeshedding
-
-- Could be tricky to handle publishing of multiple packages in the context of a **workspace** environment.
+Currently `npm publish` will automatically publish the tarball to the registry without any confirmation prompt in case the user does not have OTP enabled.
--- a/accepted/0000-publish-prompt.md
+++ b/accepted/0000-publish-prompt.md
@@ -1,0 +1,67 @@
+# Publish prompt
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


0018-publish-prompt.md
```diff
@@ -0,0 +1,67 @@
...
+
+It would be very nice to expand that "review" 
+
+## Alternatives
+
+Not change it, just keep the current behavior?
+
+## Implementation
+
+This is a breaking change from the current `npm publish` behavior, it would prompt the user for confirmation:
```
ljharb: perhaps a good first step would be, provide an option to enable the prompt, and then change the default in a future major?

isaacs: We could just make the default be `--yes` when `process.stdout.isTTY === false`.  Then it'll keep on working for build scripts and CI, but prompt everyone on the CLI.

juanpicado: Do you have plans to align with other package managers?  Would be great to have a consistent behavior, I think this is a great idea.

arcanis: I agree, I think we would do this as well in Yarn. cc @zkochan

zkochan: Seems fine for single-package scenario but I am not sure this would work during a publish in a monorepo. I doubt that someone would review the content of every published package if dozens are published.

zkochan: What if it would only show the prompt if the list of files that are published differs from the list of files in the previously published version?

ljharb: Not all monorepos publish multiple packages at once; in my experience many use independent versioning at this point.

arcanis: Printing a file diff if possible could be interesting 🤔

Something else to consider is that workflows publishing multiple packages (which do exist, for example this is what we do on Yarn, and I think Babel and Jest do the same) might be more likely to just use `--yes` (which would be handy, but would kinda negate this RFC altogether).

0018-publish-prompt.md
```diff
@@ -0,0 +1,67 @@
...
+npm notice filename:      disparity-3.0.0.tgz
+npm notice package size:  5.2 kB
+npm notice unpacked size: 13.3 kB
+npm notice shasum:        4344ee202484ab134227913a3af6f4a0ae5f0a59
+npm notice integrity:     sha512-NAItmPQyt6dya[...]m5N3kfPPJYj0w==
+npm notice total files:   7
+npm notice
+
+This operation will publish your package to the npm registry.
+Do you wish to proceed? [Y/n]
```
juanpicado: I wonder what is your thought to have `N` by default?  A double accidental press enter will publish the package anyway, it happened to me couple of times when terminal get stuck for some reason.

ljharb: +1, defaults should be non-destructive

0018-publish-prompt.md
ljharb: multiple packages is common; "bumping all package versions at once even if they haven't changed, to keep all versions the same" is the original lerna behavior that babel, and many others, have thankfully abandoned :-)

ruyadorno:
Notes from the OpenRFC call:
- Let's make this an opt-in for npm@7 and then move it to opt-out in a next major release
- 2FA awareness could/should be addressed in different features / RFCs

0018-publish-prompt.md
evocateur: Clarifying Lerna behavior for future reference:

> "bumping all package versions at once even if they haven't changed, to keep all versions the same"

This isn't actually what Lerna does in the default "fixed" versioning scheme. It applies a single version to any package (and its local dependents) that has changed since the previous release (annotated git tag).

> Not all monorepos publish multiple packages at once; in my experience many use independent versioning at this point.

"Independent" versioning doesn't change anything about how Lerna detects changes between releases or bumps dependents of changed packages, it merely allows one to choose versions per-package instead of globally.

0018-publish-prompt.md
ruyadorno: agreed, will update 😊

ruyadorno:
Updated contents of the RFC based on feedback collected during the OpenRFC call and the comments left here 😊 should be ready for ratification if no contention point is raised

accepted/0000-publish-prompt.md
```diff
@@ -0,0 +1,68 @@
...
+
+Breaking the prompt into a two-step operation allows for a validation of contents prior to uploading the package.
+
+## Detailed Explanation and Rationale
+
+`npm publish` should ask for a confirmation prompt prior to uploading the package tarball.
+
+Currently "publish" has a very nice **unintended UX** for users with 2FA enabled, in which it stops the publish process in order to ask for the OTP code, allowing for a review of the file contents (and even cancelling the process altogether) prior to uploading the tarball file.
+
+A broader population of package authors can benefit from this UX if we formalize it. Allowing for review of contents and cancelation of package publishing.
```
ljharb:
```suggestion
A broader population of package authors can benefit from this UX if we formalize it. Allowing for review of contents and cancellation of package publishing.
```

accepted/0000-publish-prompt.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Alternatives
+
+1. Not change it, just keep the current behavior?
+2. Implement it behind an _opt-in_ config option
+
+## Implementation
+
+This is a breaking change from the current `npm publish` behavior, it would prompt the user for confirmation:
```
ljharb: presumably in a non-TTY, it would not prompt, making it not a breaking change for automated processes, only for the human workflow?

ruyadorno: documented in the RFC 👍

ruyadorno:
cleaned up the RFC and documented all feedback collected 👍 should be good to go

eridal:
Is it possible to make the files listed in *Tarball Contents* be sorted alphabetically?

ruyadorno:
@eridal that does sounds a bit outside of the scope of this RFC but that said, it would def be nice to have! 👍

ruyadorno:
Updated with feedback from the [OpenRFC call](https://github.com/npm/rfcs/issues/111) removing "experimental" from the name of the option/flag.

eridal:
@ruyadorno thanks for your answer

Does that happens here? https://github.com/npm/libnpmpack/blob/ebe5d4790f61f93c7045dd0341f61681ed520899/utils/tar.js#L18

I'd like to attempt to submit a separate PR to address that feature, but that module seems to be currently worked on

isaacs:
@eridal Yes, that's currently being worked on by @claudiahdz.  Publish and pack will both use the same code path for displaying the tarball contents, and presumably we'd leave that untouched by this RFC.  Sorting should be straightforward, I'd think, but I recommend you post an issue on https://github.com/npm/libnpmpack, just so that there's a thing to track.

ruyadorno:
Landed in https://github.com/npm/rfcs/commit/0ad3bdc5597fad6d735eee8c15ded4d68bd2e0a9

PR #97, opened 2020-02-04
Added rfc

restarian:
Added template 0000-robert_steckroth.md as a feature request for the npmjs website.

:waning_crescent_moon: _i also like planets -- Peter Griffin laugh: eh, eh, eh_


darcyclarke:
@restarian appreciate this contribution! We actually have a backlogged ticket for this on our end already. That said, I might ratify this and then we can move it to "implemented" once that finally gets rolled out.

PR #102, opened 2020-02-11
RFC for removing figgy-pudding from the CLI dep stack

isaacs:
[See RFC](https://github.com/npm/rfcs/blob/isaacs/internal-de-figgy-pudding/accepted/0000-npm-option-handling.md)

### DIFF
--- a/accepted/0000-npm-option-handling.md
+++ b/accepted/0000-npm-option-handling.md
@@ -1,0 +1,190 @@
+# Handling of npm config options in the CLI and dep stack
+
+NB: this is an internal RFC related to the code structure and patterns of
+the CLI and its dependencies.  No user-facing changes are contemplated by
+this proposal.
+
+## Summary
+
+- Replace usage of `figgy-pudding` with plain JavaScript option objects.
+- Document and define defaults for configs used by each module in the
+  stack.
+- Clone options objects if they may be mutated in the future.
+- Pass options objects intact to components further down the stack.
+- Canonicalize option names to match consistent `camelCase` format.
+
+## Motivation
+
+The goal of FiggyPudding was to get the npm CLI and its associated
+dependencies away from having a shared `npmconf` god-object with `.get()`
+and `.set()` methods, which could be mutated at any time, and contain any
+arbitrary fields.
+
+However, the use of FiggyPudding has led in practice to added complexity
+and an even _more_ entrenched reliance on a god-object pattern, albeit one
+that is even more opinionated and harder to use than `npmconf` was.
+
+In one example where this has proven particularly challenging, `npm ci`
+does not get the full set of config values, and thus it does not properly
+respect config values that would be handled if the user ran `npm install`.
+Since _all_ package tree generation and reification will be handled by an
+external dependency in npm v7 (ie, `@npmcli/arborist`), this will become an
+even bigger issue.
+
+## Detailed Explanation
+
+In all of the CLI's dependencies, remove `figgy-pudding`, and replace with
+an options object that uses the canonical names for npm config values.
+
+In the npm CLI, create an options object with these canonically named
+fields, applying defaults and resolving aliases, as appropriate, using the
+`npm.config` object as its source of truth.
+
+### Canonical Naming
+
+Use the `camelCase` forms of all config fields.  For example, if the user
+runs `npm install --prefer-online`, then the config options object would
+contain `{ preferOnline: true }`.
+
+Thus, the npm CLI's dependencies should use `options.preferOnline` instead
+of `options['prefer-online']`.
+
+## Rationale and Alternatives
+
+It's worth noting the problem that figgy-pudding aims to solve and the value
+it currently provides, and ways that we will continue to solve that problem
+and gain that value.
+
+1. A figgy-pudding config object is a self-documenting list of the options
+   that a given module uses, with defaults, canonical name, and so on.
+2. Figgy-pudding allows a module to have a _different_ set of respected
+   configs than it provides to the modules deeper in the stack, because
+   they all proxy to the same underlying data object.
+3. Figgy-pudding prevents any accidental mutation of the options provided
+   to a module.
+4. Use of figgy-pudding avoids tight coupling to an `npmconf` object with
+   `get` and `set` methods.
+
+The response to these points is:
+
+1. It is better to document the list of options that a module uses _in its
+   documentation_, and make it clear in the code what their default values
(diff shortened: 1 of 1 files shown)


claudiahdz:
Internally, we also use the following as our base-figgy-pudding config:
```
    baseConfig = NpmConfig(npm.config, {
      cache: path.join(npm.config.get('cache'), '_cacache'),
      color: !!npm.color,
      dirPacker: pack.packGitDep,
      hashAlgorithm: 'sha1',
      includeDeprecated: false,
      log,
      'npm-session': npmSession,
      'project-scope': npm.projectScope,
      refer: npm.referer,
      dmode: npm.modes.exec,
      fmode: npm.modes.file,
      umask: npm.modes.umask,
      npmVersion: npm.version,
      tmp: npm.tmp,
      Promise: BB
    })
```
We should clean that up too and make it clear which files/commands are actually in need of said config.

accepted/0000-npm-option-handling.md
```diff
@@ -0,0 +1,190 @@
...
+    ```
+
+    and document the `someFoo` field in the package `README.md`.
+
+2. When a module in the CLI dependency stack needs to call out to one lower
+   down (for example, arborist calling pacote, which in turn calls
+   npm-registry-fetch, which then calls make-fetch-happen, etc.), we can
+   just pass the options object to the deeper dep.
+
+    It's _fine_, really.
```
evocateur: Until someone forgets to clone and mutates the config object passed down. I've seen that anti-pattern far too many times inside npm-owned code (with no equivalent to eslint [no-param-reassign](https://eslint.org/docs/rules/no-param-reassign#props) enabled) to be especially confident of this.

_Especially_ when considering third-party consumers of said npm-owned libraries.

evocateur: Also, leaving aside the parameter hygiene issue, you now have to document (or at least be _aware_?) of config that isn't relevant in the context of a given module. I'm less confident folks will apply this "just pass around a bag of options" pattern 100% consistently across 100% of the stack without some help, either via static analysis, really really good memories, or some other (hopefully) automated process.

isaacs: So, figgy-pudding doesn't actually solve most of those problems.  Config hygiene has still been a huge problem.

There is no One Weird Trick for this.  It comes down to:

1. good docs that outline which configs are used
2. tests with 100% coverage
3. removing configs that are no longer relevant, and refactoring them when they don't make sense (see: only/also/production/optional/dev flags)
4. setting good defaults close to where they matter.

Third-party consumers of npm-owned libraries generally just passed in POJOs anyway, which would get figgy-ified.  The problem is that once a module is using figgy-pudding, everything _it_ uses must _also_ use figgy-pudding, and it has to be careful not to accidentally exclude an option that one of its deps will end up using, so every level of the stack effectively has to know about every other.  It's not encapsulation, it's the opposite of that.  And no one can debug its peculiar restrictions.

We can explore doing something clever with config management later, but for now, _nothing_, and just having good tests, is far better than the _something_ we have.

isaacs: > Until someone forgets to clone and mutates the config object passed down.

The recent move to 100% test coverage in all new code and for all dependencies has caught a ton of these types of problems, with far less hazardous side effects or bizarre restrictions imposed by figgy-pudding.

ljharb: Is there a reason not to Object.freeze it, and easily force the cloning?

isaacs: > Is there a reason not to Object.freeze it, and easily force the cloning?

I don't see why not.

evocateur: A frozen options object goes a long way toward mitigating my concerns (many of which, yes, no program can fix entirely).

isaacs:
On the documentation angle, the plan that we've adopted as a team is:

1. Update deps that are using non-canonical option names.  (Only offenders turn out to be npm-registry-fetch, which uses css-case names instead of camelCase, and npm-pick-manifest, which uses the cute `enjoyBy` instead of the canonical `before`.)  This should fix the issues we're having trying to migrate to pacote 10, which is not fp enabled.
2. Add the flat options object to the cli, start passing that instead of a fp object.  (This actually works just fine for the fp-enabled deps.  They just pass a plain object to fp instead passing an fp proxy, and it uses fp from there onwards.)
3. Bump major and convert npm-registry-fetch, ssri, npm-pick-manifest, cacache to use options object instead of fp.
4. Bump major, convert from fp to pojo opts, and update deps in all the `libnpm*` and `npm-profile` packages.
5. Pull all the updated deps into the CLI.

At every point, when updating a dep to use pojo options instead of fp:

1. Update code to set defaults in an intelligible way.  (This varies, depending on how many defaults it sets, what it uses them for, etc.)
2. Document options that are relevant to this module.
3. If the options object is mutated, document this.  (Exceedingly rare, but eg SSRI and pacote depend on this for updating integrity, size, etc.)
4. If the options object is mutated _and not cloned_, then _definitely_ document this.  (I haven't found any cases where this isn't a bug, so I'd be surprised if anything makes it through this refactor with this property intact, but leaving open the possibility that there's a use case we've failed to consider.)
5. Document which modules this package's options object is sent to.  (Eg, Arborist sends its options object to pacote, pacote sends it to SSRI, cacache, and npm-registry-fetch, etc.)  So, while it might not be feasible or maintenance-cost-effective to document _every_ option that _any_ dep might need and keep it up to date, at least we can send the reader in the right direction to compile that list for themselves when they need it.
6. 100% test coverage.  Always and everywhere.  If you're ripping up the ground anyway, may as well put down something less grotty in its place.

accepted/0000-npm-option-handling.md
```diff
@@ -0,0 +1,190 @@
...
+Since _all_ package tree generation and reification will be handled by an
+external dependency in npm v7 (ie, `@npmcli/arborist`), this will become an
+even bigger issue.
+
+## Detailed Explanation
+
+In all of the CLI's dependencies, remove `figgy-pudding`, and replace with
+an options object that uses the canonical names for npm config values.
+
+In the npm CLI, create n options object with these canonically named
```
darcyclarke: Looks like a typo slipped through (ie. `...create "an" options...` unless you mean `n` in terms of *Big O* notation

isaacs: "an" == "n" where `n === 1`.  Which it does in this case.  (I'll fix the typo, thanks :)

PR #103, opened 2020-02-14
RFC: Add npm workspaces

ruyadorno:
[See RFC](https://github.com/npm/rfcs/blob/26e8ac6ee176943d6522d5d057fab05e37655e1c/accepted/0000-workspaces.md)

### DIFF
--- a/accepted/0000-workspaces.md
+++ b/accepted/0000-workspaces.md
@@ -12,10 +12,9 @@ The name “workspaces” is already established in the community with both Yarn
 
 ## Detailed Explanation
 
-After sourcing feedback from the community, there are 3 major implementations/changes required in the **npm cli** in order to provide the feature set that would enable a better management of nested packages.
+After sourcing feedback from the community, there are 2 major implementations/changes required in the **npm cli** in order to provide the feature set that would enable a better management of nested packages.
 - Make the **npm cli** **workspace**-aware.
 - Install: In a **npm workspaces** setup users expect to be able to install all nested packages and perform the associated lifecycle scripts from the **Top-level workspace**, it should also be aware of **workspaces** that have a **dependency** on one another and **symlink** them appropriately.
-- Run some npm subcommands in the context of each **workspace**.
 
 The set of features identified in this document are the ones that are essential to an initial [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) of the **npm workspaces** support. The community should expect further development of this feature based on the feedback we collected and documented at the end of this RFC.
 
@@ -85,55 +84,6 @@ $ npm install
 ```
 
 For the initial **workspaces** implementation, we're going to stick with **arborist**'s default algorithm that privileges **hoisting packages** but will place packages at nested `node_modules` whenever necessary.
-
-### 3. Run commands across all child packages
-
-Make **npm cli** subcommands **npm workspaces**-aware, so that running a command tries to run it within all **workspaces** as long as a **workspaces configuration** field is properly defined in `package.json`.
-
-Only a subset of commands are to be supported:
-
-- `fund` List funding info for all **workspaces**
-- `ls` List all packages including **workspaces**
-- `outdated` List outdated **dependencies** including **workspaces** and its **dependencies**
-- `run-script` Run arbitrary **scripts** in all **workspaces**, skip any **workspace** that does not have a targetting **script**
-- `rebuild` Rebuild all **workspaces**
-- `restart`
-- `start`
-- `stop`
-- `test` Run test **scripts** in all **workspaces**
-- `update` Updates a **dependency** across the entire installation tree, including **workspaces**
-- `view` View registry info, also including **workspaces**
-
-A new **npm cli** configuration value should be introduced in order to allow for filtering out a subset of the **workspaces** in which to run these commands. e.g: `--filter`
-
-#### Test example:
-
-```
-├── package.json { "name": "foo", "workspaces": ["dep-a", "dep-b"] }
-├── dep-a
-│   └── package.json { "version": "1.0.0" }
-└── dep-b
-    └── package.json { "version": "1.3.1" }
-
-$ npm test
-
-> foo@1.0.0 test /Users/username/foo
-> echo "Error: no test specified" && exit 1
-
-Error: no test specified
-npm ERR! Test failed.  See above for more details.
-
-> dep-a@1.0.0 test /Users/username/foo/dep-a
-> done
-
-> dep-b@1.3.1 test /Users/username/foo/dep-b
-> done
-```
-
-
-### 4. Publishing workspaces
-
-A **Top-level workspace** package may not be published to the registry and for all publishing purposes having a valid `"workspaces"` entry in a `package.json` is going to be the equivalent of `"private": true`.
 
 
 ## Examples
@@ -340,5 +290,6 @@ During the discussions around this RFC it was brought up to our attention that a
 
 ## Unresolved Questions and Bikeshedding
 
-- Support for non-nested file structure workspaces (ex: adjacent directories)
 - Should we support an opt-out config? e.g: `workspacesEnabled`
+- Should we prevent **publish** of the **Top-level workspace** package? Previous note on it:
+  - A **Top-level workspace** package may not be published to the registry and for all publishing purposes having a valid `"workspaces"` entry in a `package.json` is going to be the equivalent of `"private": true`.
--- a/accepted/0000-workspaces.md
+++ b/accepted/0000-workspaces.md
@@ -57,7 +57,7 @@ We're following the lead of Yarn in supporting the `workspaces` `package.json` f
 
 ### 2. Installing dependencies across child packages
 
-Change `npm install` (arborist) behavior to make it install every nested package by default once a valid configuration is defined.
+Change `npm install` ([arborist](https://www.npmjs.com/package/@npmcli/arborist)) behavior to make it install every nested package by default once a valid configuration is defined.
 
 Arborist should also be aware of all nested workspaces and correctly link to an internal workspace should it match the required semver version of an expected dependency anywhere in the installing tree. e.g:
 
--- a/accepted/0000-workspaces.md
+++ b/accepted/0000-workspaces.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Add a set of features to the npm cli that provide support to managing multiple packages from within a singular top-level, root package.
+Add a set of features to the **npm cli** that provide support to managing multiple packages from within a singular top-level, root package.
 
 ## Motivation
 
@@ -12,21 +12,24 @@ The name “workspaces” is already established in the community with both Yarn
 
 ## Detailed Explanation
 
-After sourcing feedback from the community, there are 2 major implementations/changes required in the cli in order to provide the feature set that would enable a better management of nested packages.
-- Install: In a workspaces setup users expect to be able to install all nested packages and perform the associated lifecycle scripts from the root-level, it should also be aware of sibling packages that depend on one another and link them appropriately
-- Run some npm subcommands in the context of each nested workspace
+After sourcing feedback from the community, there are 3 major implementations/changes required in the **npm cli** in order to provide the feature set that would enable a better management of nested packages.
+- Make the **npm cli** **workspace**-aware.
+- Install: In a **npm workspaces** setup users expect to be able to install all nested packages and perform the associated lifecycle scripts from the **Top-level workspace**, it should also be aware of **workspaces** that have a **dependency** on one another and **symlink** them appropriately.
+- Run some npm subcommands in the context of each **workspace**.
+
+The set of features identified in this document are the ones that are essential to an initial [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) of the **npm workspaces** support. The community should expect further development of this feature based on the feedback we collected and documented at the end of this RFC.
 
 ## Rationale and Alternatives
 
-First and foremost there’s the alternative of leaving the problem set for userland to solve, there’s already the very popular project Lerna that provides some of these features.
+First and foremost there’s the alternative of leaving the problem set for userland to solve, there’s already the very popular project [Lerna](https://github.com/lerna/lerna) that provides some of these features.
 
 Also available is the alternative of supporting only the install (or bootstrap as Lerna names it) aspect of this proposal, following a less feature-rich approach but one that would still enable the basic goal of improving the user experience of managing multiple child packages but from all the feedback collected during the research phase of this RFC, this alternative is much less desirable to the community of maintainers involved.
 
 ## Implementation
 
-### 1. Make npm workspace-aware
+### 1. Workspaces configuration: Making the npm cli workspace-aware
 
-We're following the lead of Yarn in supporting the `workspaces` `package.json` field which defines a list of paths, each of these paths may be a workspace itself but it also support globs.
+We're following the lead of Yarn in supporting the `workspaces` `package.json` property which defines a list of paths, each of these paths may point to the location of a **workspace** in the file system but it also support **globs**.
 
 `package.json` example:
 
@@ -54,12 +57,14 @@ We're following the lead of Yarn in supporting the `workspaces` `package.json` f
 }
 ```
 
+The **npm cli** will read from the paths and **globs** defined in this **workspaces configuration** and look for valid `package.json` files in order to create a list of packages that will be treated as **workspaces**.
+
 
 ### 2. Installing dependencies across child packages
 
-Change `npm install` ([arborist](https://www.npmjs.com/package/@npmcli/arborist)) behavior to make it install every nested package by default once a valid configuration is defined.
+Change `npm install` ([arborist](https://www.npmjs.com/package/@npmcli/arborist)) behavior to make it properly install **dependencies** for every **workspace** defined in the **workspaces configuration** described above.
 
-Arborist should also be aware of all nested workspaces and correctly link to an internal workspace should it match the required semver version of an expected dependency anywhere in the installing tree. e.g:
+**Arborist** should also be aware of all **workspaces** in order to correctly link to another internal **workspace** should it match the required semver version of an expected **dependency** anywhere in the installing tree. e.g:
 
 ```
 // Given this package.json structure:
@@ -76,34 +81,30 @@ $ npm install
 │   ├── dep-a -> ./dep-a
 │   └── dep-b -> ./dep-b
 ├── dep-a
-│   └── node_modules
-│       └── dep-b -> ../../../node_modules/dep-b
 └── dep-b
 ```
 
-NOTE: The final implementation of the symlinking structure might differ from the examples laid out here (which are trying to be as illustrative as possible). Arborist might end up resolving all symlinks up to the actual place instead of creating a chain of symlinks.
+For the initial **workspaces** implementation, we're going to stick with **arborist**'s default algorithm that privileges **hoisting packages** but will place packages at nested `node_modules` whenever necessary.
 
 ### 3. Run commands across all child packages
 
-Make npm commands workspace-aware, so that running a command tries to run it within nested packages as long as a workspaces field is defined in `package.json`.
+Make **npm cli** subcommands **npm workspaces**-aware, so that running a command tries to run it within all **workspaces** as long as a **workspaces configuration** field is properly defined in `package.json`.
 
-Only a subset of npm commands are to be supported:
+Only a subset of commands are to be supported:
 
-- `fund` List funding info for all packages
-- `ls` List all packages
-- `outdated` List outdated deps for all packages
-- `publish` Publishes all packages
-- `run-script` Run arbitrary scripts for all packages
-- `rebuild` Rebuild all packages
+- `fund` List funding info for all workspaces
+- `ls` List all packages including workspaces
+- `outdated` List outdated deps including workspaces and its deps
+- `run-script` Run arbitrary scripts in all workspaces
+- `rebuild` Rebuild all workspaces
 - `restart`
 - `start`
 - `stop`
-- `test` Run tests in all packages
-- `update` Update a dependency in all packages
-- `version` Bump versions for all packages
-- `view` View registry info for all packages
+- `test` Run tests in all workspaces
+- `update` Updates a dependency across the entire installation tree, including workspaces
+- `view` View registry info but also including workspaces
 
-A new config value should be introduced in order to allow for filtering out a subset of the packages in which to run these commands. e.g: `--filter`
+A new **npm cli** configuration value should be introduced in order to allow for filtering out a subset of the **workspaces** in which to run these commands. e.g: `--filter`
 
 #### Test example:
 
@@ -132,14 +133,14 @@ npm ERR! Test failed.  See above for more details.
 
 ### 4. Publishing workspaces
 
-A workspace may not be published to the registry and for all publishing purposes having a valid `"workspace"` entry in a `package.json` is going to be the equivalent of `"private": true`.
+A **Top-level workspace** package may not be published to the registry and for all publishing purposes having a valid `"workspaces"` entry in a `package.json` is going to be the equivalent of `"private": true`.
 
 
 ## Examples
 
 ### 1. Simplistic example expanding on symlinking structure and `package-lock` file shape.
 
-Given a workspaces setup with the following contents:
+Given a **npm workspaces** setup with the following contents:
 
 ```
 $ cat ./package.json
@@ -185,6 +186,18 @@ $ cat ./packages/worskpace-b/package.json
         "react": "^16.x.x"
     }
 }
+
+$ cat ./packages/workspace-c/package.json
+{
+    "name": "workspace-c",
+    "version": "1.0.0",
+    "peerDependencies": {
+        "react": "^16.x.x"
+    },
+    "dependencies": {
+        "workspace-b": "^1.0.0"
+    }
+}
 ```
 
 Will result in the following symlinking structure:
@@ -198,25 +211,21 @@ $ tree
 │   ├── libnpmutil -> ./core/libnpmutil
 │   ├── workspace-a -> ./packages/workspace-a
 │   ├── workspace-b -> ./packages/workspace-b
+│   ├── workspace-c -> ./packages/workspace-c
 │   └── react
 ├── core
 │   └── libnpmutil
-│       ├── package-lock.json
-│       └── node_modules
-│           └── lodash -> ../../../node_modules/lodash
 └── packages
     ├── workspace-a
-    │   ├── package-lock.json
-    │   └── node_modules
-    │       ├── react -> ../../../node_modules/react
-    │       └── worspace-b -> ../../../node_modules/workspace-b
-    └─ worspace-b
-        ├── package-lock.json
-        └── node_modules
-            └── react -> ../../../node_modules/react
+    ├── workspace-b
+    └─ worspace-c
+        └── node_modules
+            └── workspace-b@1.0.0
 ```
 
 And the following `package-lock.json` files:
+
+NOTE: The following lockfile is for illustration purpose only and its final shape might differ.
 
 ```
 $ cat ./package-lock.json
@@ -256,13 +265,21 @@ $ cat ./package-lock.json
       "resolved": "packages/workspace-b",
       "link": true
     },
+    "node_modules/workspace-c": {
+      "resolved": "packages/workspace-c",
+      "link": true
+    },
     "packages/workspace-a": {
       "name": "workspace-a",
       "version": "1.7.3"
     },
     "packages/workspace-b": {
       "name": "workspace-b",
-      "version": "2.1.1"
+      "version": "1.0.0"
+    },
+    "packages/workspace-c": {
+      "name": "workspace-c",
+      "version": "1.0.0"
     }
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


isaacs:
> the root is a package that's published

I'm really perplexed about a workspace root that's also a package getting published.  How does that work when installing a workspace root as a dependency?  Do we just strip off the `workspaces` declaration when installing, or something?  Do the workspace packages get listed as dependencies?

ljharb:
presumably you’d want to do that anyways if you didn’t require private: true for workspaces to work.

see https://github.com/benmosher/eslint-plugin-import for that particular example.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Detailed Explanation
+
+After sourcing feedback from the community, there are 2 major implementations/changes required in the cli in order to provide the feature set that would enable a better management of nested packages.
+- Install: In a workspaces setup users expect to be able to install all nested packages and perform the associated lifecycle scripts from the root-level, it should also be aware of sibling packages that depend on one another and link them appropriately
+- Run some npm subcommands in the context of each nested workspace
+
+## Rationale and Alternatives
+
+First and foremost there’s the alternative of leaving the problem set for userland to solve, there’s already the very popular project Lerna that provides all of these features.
```
wesleytodd:
```suggestion
First and foremost there’s the alternative of leaving the problem set for userland to solve, there’s already the very popular project Lerna that provides some of these features.
```

I think Lerna is a great tool when you are specifically working in a monorepo.  I hope that we can think beyond repo setup in how users work on projects.  This one is a bit of a nit-pick, so don't put too much weight on the requested change, but I don't want us to fall into the trap of thinking the community has provided everything folks need.  We can and should strive for more than we have today.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+    "workspaces": [
+        "packages/*"
+    ]
+}
+```
+
+
+### 2. Installing dependencies across child packages
+
+Change `npm install` (arborist) behavior to make it install every nested package by default once a valid configuration is defined.
```
wesleytodd:
```suggestion
Change `npm install` ([arborist](https://www.npmjs.com/package/@npmcli/arborist)) behavior to make it install every nested package by default once a valid configuration is defined.
```

For any folks unaware (most likely most people).

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+> done
+
+> dep-b@1.3.1 test /Users/username/foo/dep-b
+> done
+```
+
+
+### 4. Publishing workspaces
+
+A workspace may not be published to the registry and for all publishing purposes having a valid `"workspace"` entry in a `package.json` is going to be the equivalent of `"private": true`.
```
wesleytodd: This seems like a needless restriction and blocks some interesting use cases. I can imagine a developer experience which involves teams publishing workspaces as a mechanism for sharing a setup.  You could just specify that installing a published package with a workspaces key will do nothing, and then let userland tooling extend this if it wants.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+## Prior Art
+
+- [Lerna](https://github.com/lerna/lerna)
+- [Yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)
+- [Pnpm workspaces](https://pnpm.js.org/en/workspaces)
+
+## Unresolved Questions and Bikeshedding
+
+- Should we add a Workspace class (subclass of Node) in Arborist?
+- For this initial implementation there's no intention of adding a more ellaborate version/publish subcommand/workflow that would allow for bumping versions of nested packages + updating dependency references elsewhere across a workspace (similar to `lerna --independent` publish workflow).
```
wesleytodd:
```suggestion
- For this initial implementation there's no intention of adding a more ellaborate version/publish subcommand/workflow that would allow for bumping versions of nested packages + updating dependency references elsewhere across a workspace (similar to `lerna --independent` publish workflow).
- Support for non-nested file structure workspaces (ex: adjacent directories)
```

I have a bunch of opinions here to share and want to take the time to fully write them up.  But I wanted to add this to the questions in case other folks had ideas here to share.

accepted/0000-workspaces.md
wesleytodd: @isaacs 

> Do we just strip off the workspaces declaration when installing

I would think for a use case like sharing a setup you would not strip it, only ignore it.

> Do the workspace packages get listed as dependencies?

No, as with other workspace implementations you only get it when developing locally with it as the top level package.  So to publish a package it must define all that it requires when installing from the registry.

wesleytodd:
> multirepo vs monorepo

I would really love for us to get past this focus on "monorepo" vs "multirepo".  Can we just talk about Multi-Project support?  The repo part is a vestige of our node module resolution mechanism being tied to the filesystem structure and a repo being the only reliable way of ensuring file structures across systems.  In the future we will get a lot more value in decoupling this all from the filesystem, and building a workspace implementation by copying Yarn v1 is already behind the state of the art.

There are a bunch of examples of decoupling the file system from the module resolution mechanism: Yarn PnP, tink and [Import Maps](https://github.com/nodejs/modules/issues/477).  I think we should strongly think about how multi-project looks when we have a future where the directory structure does not dictate where modules live on the filesystem.

isaacs:
> There are a bunch of examples of decoupling the file system from the module resolution mechanism: Yarn PnP, tink and Import Maps.

Agreed! And they're all outside the scope of this rfc ;)

Our intent is to use a fallback fs implementation (is, tink), so we still have to figure out what the file system will look like. If we then virtualize it, we still need a way to unwind it when requested and need to know what to virtualize. So, all this gross disk stuff needs to get worked out regardless. Also, files on disk has its advantages.

accepted/0000-workspaces.md
isaacs: So if it's not a top of tree module (root or link target, not a deep in node_modules) then the workspaces object is irrelevant? That could work I guess.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+
+First and foremost there’s the alternative of leaving the problem set for userland to solve, there’s already the very popular project Lerna that provides all of these features.
+
+Also available is the alternative of supporting only the install (or bootstrap as Lerna names it) aspect of this proposal, following a less feature-rich approach but one that would still enable the basic goal of improving the user experience of managing multiple child packages but from all the feedback collected during the research phase of this RFC, this alternative is much less desirable to the community of maintainers involved.
+
+## Implementation
+
+### 1. Make npm workspace-aware
+
+We're following the lead of Yarn in supporting the `workspaces` `package.json` field which defines a list of paths, each of these paths may be a workspace itself but it also support globs.
```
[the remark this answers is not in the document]
isaacs: I suppose if we're going that far, may as well. Make it work somewhat like files:[...] in package.json.

isaacs:
> something custom (eg, eslint-plugin-import)
> the root is a package that's published
subpackages also exist that are published (and npmignored by the root)

What happens if they're not npmignored? I think maybe we should treat them just like normal package contents, included by default.

wesleytodd:
@isaacs your points are a bit unclear to me.  I am not sure what a "fallback fs" means here, and since this whole proposal is centered around the existing disk layout based methodology then how is that not applicable?

As for "files on dis has its advantages", I don't disagree, but could you elaborate on how that applies specifically to this workspace implementation?

isaacs:
> @isaacs your points are a bit unclear to me. I am not sure what a "fallback fs" means here, and since this whole proposal is centered around the existing disk layout based methodology then how is that not applicable?

Fallback fs is the approach used by tink. We model out the file system, and instead of providing an alternative module loader, like pnp does, we virtualization the fs module in node, such that any ENOENT in a node_modules folder falls back to the appropriate package contents in a single global cache. 

On advantage of this approach is that you can always unwind to the "real" location, and no changes are required in the vast majority of use cases. 

My point is that, even with such a virtualized system, we still need to decide the shape of it as a file system. 

The biggest advantage of files on disk (or a sufficiently convincing fallback fs) is that it works trivially with the other tools, including editing to float patches, webpack and other transpilers, etc.

wesleytodd:
> Fallback fs is the approach used by tink

Ah, I hadn't looked at that part of what it was doing. I am not opposed to this approach but I would like to know why you think that is better than a loader as the runtime portion?  It seems like this reaches out of the package installer's domain and is also more heavy handed.

> On advantage of this approach is that you can always unwind to the "real" location, and no changes are required in the vast majority of use cases.

It seems like this is possible from any of the approaches we might take here, right?

> we still need to decide the shape of it as a file system.

I guess since the "shape of it as a filesystem" maps directly to the shape of the optimized dep tree?  Because if you choose something like an Import Map over a virtualized fs that relationship doesn't really matter anymore.

> The biggest advantage of files on disk (or a sufficiently convincing fallback fs) is that it works trivially with the other tools, including editing to float patches, webpack and other transpilers, etc.

This is a great point and is the thing I have always been worries about with tink and PnP.  Are those other approaches finding a solution to this?  I do think we should have an answer for this as it relates to workspaces so the design of that doesn't have to change too much in relation.

accepted/0000-workspaces.md
ruyadorno: @wesleytodd this is looking interesting 👍 assuming there's no technical limitations that  might prevent it - I'm onboard with the suggestions

> You could just specify that installing a published package with a workspaces key will do nothing, and then let userland tooling extend this if it wants.

just to clarify, in this case there's no filtering out mechanism at the moment of publishing a top-level workspace 🤔 meaning all the contents of nested packages within it get published along with that top-level package?

ruyadorno:
@ljharb having the final implementation being flexible enough to support these use cases is my `#1` priority 😁 that said, I have a few questions regarding some specific examples:

>   * subpackages depended on by other subpackages are able to be linked for dev/tests, but _also_ able to be unlinked, so the explicit versions referenced are tested

by "able to be linked for dev/tests" are you suggesting anything more complex than just the symlinking generated after `npm install` that is currently described currently in the RFC?

"_also_ able to be unlinked, so the explicit versions referenced are tested" this sounds to me like something that might be outside of the scope of an initial workspaces implementation but to be clear, in the current proposal the way to _unlink_ a subpackage (or have it be fetched from the registry instead) would be to define a semver range that is not satisfied by the `package.json` version that is currently sitting in the file system in its dependents, but again all this is handled by arborist meaning that if you want to do things for dev/tests only it would imply a more complex workflow in which you have to manually edit versions, run `npm install` and test things again - does this makes sense? did I understood correctly your points in this item? 😊 let me know

>   * subtle peer dep conflicts are surfaced, ideally based on the published refs _and_ based on linked refs

This item seems to be outside of the scope of this initial work 🤔 at least in its current form - how could we better handle that scenario? Can you (or anyone else really) provide some examples/ideas?

>   * subpackages also exist that are published (and npmignored by the root)

Are these manually `npmignored`?

---

Feel free to suggest changes to the document if you want to see any specific wording in the final ratified version.

And thanks for the list! 😄 it provides a great initial checklist for all test cases we'll want to make sure to have in arborist.

ljharb:
@ruyadorno 
> are you suggesting anything more complex

the current main flaw with linking is that every peer dep has to be separately linked too; these would need to be solved.

> did I understood correctly your points in this item?

What I mean is, there's two modes that are important: development (where links are ideal) and CI/production (where i want to test what *consumers* will get when they install). In other words, if subpackage A depends on subpackage B@2.0.0, i'd expect in dev that B would be the current working directory, but that when "unlinked", or in prod, that it'd be the published v2.0.0.

> This item seems to be outside of the scope of this initial work 🤔 at least in its current form - how could we better handle that scenario? Can you (or anyone else really) provide some examples/ideas?

Whether it's in scope or not, I'd consider it a blocker. Issues around peer deps are wildly multiplied in a workspaces scenario - for example, enzyme has a handful of "react adapters", each of which requires a different, conflicting version of react. In airbnb's javascript style guide, there's two eslint config packages (the main one, and the base), the main requires the base, and both the main and the base peer depend on eslint as well as a number of eslint plugins, some overlapping. Basically, my concern is that a workspaces development model makes it much easier to obscure issues around dependency conflicts, and I think it's highly critical that the tooling surface these.

> Are these manually npmignored?

Yes, I think it's totally fine to leave "what's in the tarball" management to the user, but it might be nice to warn the user if one published package would end up containing another.

evocateur:
> the root is a package that's published

FWIW, I really don't like this pattern. The root manifest of a multi-package repository should be the place you coordinate workspaces, test/dev dependencies, and other bits of "global" configuration. Making the root manifest _also_ a published package is not gaining enough benefits to outweigh the drawbacks (confusing, arbitrary ignores) in my opinion.

accepted/0000-workspaces.md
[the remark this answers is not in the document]
evocateur: Beware globstars (`**`) diving into nested `node_modules` when it shouldn't...

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+    ]
+}
+```
+
+
+### 2. Installing dependencies across child packages
+
+Change `npm install` ([arborist](https://www.npmjs.com/package/@npmcli/arborist)) behavior to make it install every nested package by default once a valid configuration is defined.
+
+Arborist should also be aware of all nested workspaces and correctly link to an internal workspace should it match the required semver version of an expected dependency anywhere in the installing tree. e.g:
```
evocateur: Will Arborist also support the [`workspace:` specifier](https://pnpm.js.org/en/workspaces#workspace-ranges-workspace) supported by Yarn v2 and pnpm?

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+│   └── node_modules
+│       └── dep-b -> ../../../node_modules/dep-b
+└── dep-b
+```
+
+NOTE: The final implementation of the symlinking structure might differ from the examples laid out here (which are trying to be as illustrative as possible). Arborist might end up resolving all symlinks up to the actual place instead of creating a chain of symlinks.
+
+### 3. Run commands across all child packages
+
+Make npm commands workspace-aware, so that running a command tries to run it within nested packages as long as a workspaces field is defined in `package.json`.
```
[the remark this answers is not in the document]
evocateur: Outside of workspace-aware `npm run`, I don't think npm needs to provide a generic `exec` command. As long as `npm ls` continues to provide parseable output (no reason to expect otherwise), I would consider that "good enough" for core.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+- `outdated` List outdated deps for all packages
+- `publish` Publishes all packages
+- `run-script` Run arbitrary scripts for all packages
+- `rebuild` Rebuild all packages
+- `restart`
+- `start`
+- `stop`
+- `test` Run tests in all packages
+- `update` Update a dependency in all packages
+- `version` Bump versions for all packages
```
evocateur: Presumably a workspace-aware `npm version` would **not** include such fancy git-related magic that say, `lerna version` does? Because I can guarantee you a gigantic bug-farm, otherwise.

accepted/0000-workspaces.md
evocateur: I think it's the right call to make a `"workspaces"` field in package.json equivalent to `"private": true`. "Sharing configuration" (for an identical workspace, what?) should be a separate concern.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+
+```
+$ cat ./package-lock.json
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
```
evocateur: Can we name this something more significant, like ... `_rootManifest`? An empty string as a key feels like high potential for long-term regret.

accepted/0000-workspaces.md
evocateur: Nota bene: `--independent` has nothing to do whatsoever with Lerna's cross-dependency change detection, it just lets you choose the version bump per-package instead of per-monorepo.

And yes, npm's initial implementation of workspaces should stay far, _far_ away from this certifiable bug farm.

accepted/0000-workspaces.md
wesleytodd: > meaning all the contents of nested packages within it get published along with that top-level package?

My Yeah you would use the normal methods for specifying which files make it into the tarball if you wanted different behavior.

> Sharing configuration" (for an identical workspace, what?) should be a separate concern.

Just wondering, why do you think this?  To me this clause of "workspaces means private" makes it a concern here.  I am fine having that as a separate conversation as long as this clause does not make it into the implementation.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+    "workspace-a": {
+      "version": "file:packages/workspace-a"
+    },
+    "workspace-b": {
+      "version": "file:packages/workspace-b"
+    }
+  }
+}
+
+$ cat ./packages/workspace-a/package-lock.json
```
evocateur: If the root lockfile already serializes all of this information, why are the nested lockfiles necessary? Or am I missing something? (I'm probably missing something)

Or, looking closer, I see that the nested lockfiles _are_ specifying their "local" dependencies etc. Why can't that information be hoisted to the root lockfile's `packages` subtree?

accepted/0000-workspaces.md
wesleytodd: This change might be missleading.  You cannot "add a line" without "editing the line before or after.  My recommendation was just to add the open question:

```
- Support for non-nested file structure workspaces (ex: adjacent directories)
```

accepted/0000-workspaces.md
evocateur: @wesleytodd The manifest that coordinates the workspaces, as I've commented elsewhere, should not be conflated with a manifest that specifies a publishable package. The fact that Lerna _can_ do that is an accident of history, and does not represent what I believe to be a robust pattern for managing multi-package repositories.

accepted/0000-workspaces.md
wesleytodd: Yeah I would agree that this kind of logic should be kept out of lower level tools like npm.  The problem here is that the line is fuzzy, because folks do want sane version bumping and tagging behavior.  Is there a good middle-ground you think we might get to @evocateur?

accepted/0000-workspaces.md
wesleytodd: I would also like to hear in general what the npm team thinks about these kind of additions?  This is a great way to make whole package ecosystems which have strict tool version dependence.  I would love it if we could avoid this (although yarn is already introducing these) and having a general understanding of where we all want to be in the future would be great so we can have productive conversation about these individual additions as they come up.

accepted/0000-workspaces.md
wesleytodd: Publishing it is a distribution mechanism, and the registry is the distribution mechanism in our ecosystem. If folks find novel ways to use this as both a distribution mechanism and a tool to build better developer experience, GREAT!  I don't see this needless restriction gaining us anything at this time.

While it might not not help your monorepo-centric tool doesn't mean that it cannot help other types of setups.  Do you think that your worries hold up when you move outside of a monorepo?  Do you have suggestions for supporting multi project workspaces outside of a monorepo?

My single biggest concern for this implementation, that it not be tightly coupled to the monorepo setup.  If npm just follows yarn and lerna on this I will continue *not* to be able to use them for my main use cases.  Most OSS projects are a collection of projects in their own repos.  Some popular ones outgrew the tooling (a separate problem I think we can solve), but the vast majority have not.  If we do not make this a top priority for workspaces in npm I think we will have missed an opportunity to appeal to a much larger audience.

wesleytodd:
Just to hoist this topic to the top level thread:

I strongly think we need to think both about monoreps (well trodden ground today) as well as multirepo. 

Most projects I work on is multirepo.  We have great reasons for this and we will not be changing.  Today our tooling provides almost no support for this, and so I have many DX workarounds. To be clear, I am not against monorepos. I helped move my last companies codebase into one, it is great for many types of projects.  But not all.   We can do better for this majority of package authors, and we should.

ljharb:
@evocateur re https://github.com/npm/rfcs/pull/103#issuecomment-587578206
> FWIW, I really don't like this pattern

ftr neither do i, but the best way for repos that use it to migrate to a better pattern is if they have reliable tooling available that supports both patterns - like npm.

accepted/0000-workspaces.md
ruyadorno: I'll go ahead and accept this item since it's mostly an "Unresolved question" still

accepted/0000-workspaces.md
ruyadorno: > Will Arborist also support the `workspace:` specifier supported by Yarn v2 and pnpm?

no, not in this initial implementation at least - of course we need to keep sourcing feedback from the community, specially in the form of the foundation WGs (just to say that we're never shutting the door to new ideas and if the community wants any new feature outside of the scope of this RFC in the future it can always make its way into a future version)

accepted/0000-workspaces.md
ruyadorno: > Presumably a workspace-aware `npm version` would **not** include such fancy git-related magic that say, `lerna version` does? Because I can guarantee you a gigantic bug-farm, otherwise.

yep 👍 100% the idea is to avoid all that crazyness, the `version` support would just be the equivalent of running `npm version <target>` inside each of the packages

accepted/0000-workspaces.md
ruyadorno: oh this is outside of the scope of this RFC, this is just a basic `package-lock` v2 template - maybe it's worth poking the team at https://github.com/npm/arborist if you have strong feelings about it 😊

accepted/0000-workspaces.md
wesleytodd: With many commits or just one? Same question for the tags.  I think this might already be getting into the craziness which @evocateur was worried about.

accepted/0000-workspaces.md
isaacs: Lockfiles are for machines.  The key in the `packages` section of a v2 or v3 lockfile is the `/`-normalized relative path from the root to the node in question.  There's a valid point as to whether it ought to be in the lockfile at all, since it's duplicative with `package.json`, but if it _is_ included, then `""` is the most portable and sensible key for it, imo.  Having to constantly do `node.location || '_rootManifest'` (and then transpose it back, as well) imposes a fair bit of added complexity in the metadata handling logic.

accepted/0000-workspaces.md
isaacs: The thinking here is that you might define a workspace with a bunch of other packages (as separate git working dirs), and you'd still want those other projects to have a lockfile included for users who check them out independently.

accepted/0000-workspaces.md
ruyadorno: just one commit/tag for all repos involved (mind the RFC also mentions the possibility of filtering a subset of packages) 😅 but IMO if we realize `version` and `publish` are too overwhelming to support in this initial implementation I'll be in favor of dropping support to it altogether - mind that getting the _installing_ part of the workspaces right is the core of this RFC

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,480 @@
...
+
+We're following the lead of Yarn in supporting the `workspaces` `package.json` field which defines a list of paths, each of these paths may be a workspace itself but it also support globs.
+
+`package.json` example:
+
+```
+{
+    "name": "workspace-example",
+    "version": "1.0.0",
+    "workspaces": {
```
ruyadorno: should we also support the singular form `workspace` ?

ljharb: oof, let's please not repeat "bundleDependencies" vs "bundledDependencies" :-)

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+- `rebuild` Rebuild all packages
+- `restart`
+- `start`
+- `stop`
+- `test` Run tests in all packages
+- `update` Update a dependency in all packages
+- `version` Bump versions for all packages
+- `view` View registry info for all packages
+
+A new config value should be introduced in order to allow for filtering out a subset of the packages in which to run these commands. e.g: `--filter`
```
danielleadams: Is this flag for listing subdirectories that are excluded or subdirs that are included in the command?

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+
+Make npm commands workspace-aware, so that running a command tries to run it within nested packages as long as a workspaces field is defined in `package.json`.
+
+Only a subset of npm commands are to be supported:
+
+- `fund` List funding info for all packages
+- `ls` List all packages
+- `outdated` List outdated deps for all packages
+- `publish` Publishes all packages
+- `run-script` Run arbitrary scripts for all packages
```
danielleadams: so if I wanted to run a script (ie. `build`) for a subdirectory called `ui`, I would run (for example) `npm run-script build --filter ui`?

danielleadams: Also, do all the packages/subdirectories have to contain the same script in their `package.json`? If one does not have the script, will it fail or does it skip that subdir?

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+
+- [Lerna](https://github.com/lerna/lerna)
+- [Yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)
+- [Pnpm workspaces](https://pnpm.js.org/en/workspaces)
+
+## Unresolved Questions and Bikeshedding
+
+- Should we add a Workspace class (subclass of Node) in Arborist?
+- For this initial implementation there's no intention of adding a more ellaborate version/publish subcommand/workflow that would allow for bumping versions of nested packages + updating dependency references elsewhere across a workspace (similar to `lerna --independent` publish workflow).
+- Support for non-nested file structure workspaces (ex: adjacent directories)
```
danielleadams: I'm curious what this would look like or if it's worth bringing into the discussion. From what I've seen, a common file structure is something like:

```
├── package.json
├── server
│  ├── package.json
├── front-end
│  ├── package.json
```
Usually there are scripts included at the root `package.json` and the dependencies are listed at the nested `package.json`s with their own scripts, such as install and build. Do you see the root `package.json` being the "main" directory? Also, do you think there will be any issues with this structure?

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+└── dep-b
+```
+
+NOTE: The final implementation of the symlinking structure might differ from the examples laid out here (which are trying to be as illustrative as possible). Arborist might end up resolving all symlinks up to the actual place instead of creating a chain of symlinks.
+
+### 3. Run commands across all child packages
+
+Make npm commands workspace-aware, so that running a command tries to run it within nested packages as long as a workspaces field is defined in `package.json`.
+
+Only a subset of npm commands are to be supported:
```
danielleadams: Is the idea to have the npm commands be run against all or subsets of packages? I ask because I'm curious if it's worth it to have commands that are run against just one workspace (ie. `npm workspace package-a run build` or something). Is this what `--filter` is supposed to cover?

accepted/0000-workspaces.md
evocateur: The precedent in Lerna is to skip workspaces that don't contain the targeted lifecycle script.

evocateur: With `lerna run`, the filter you pass is the package name (or glob), _not_ the directory name.

accepted/0000-workspaces.md
evocateur: `--filter` is the way envisioned to reduce the workspaces targeted by the command. By default, all configured workspaces are targeted by the workspace-aware `npm` subcommands.

accepted/0000-workspaces.md
ruyadorno: agreed, will document it there 😊

accepted/0000-workspaces.md
ruyadorno: this is the blurriest part of the current proposal and that's on purpose since I couldn't come up with a good enough API design, ideally I would like for it to be very simplistic and intuitive - that said, my hope is that we'll get there during the time of this RFC 😊 

some prior art in this specific domain:
- [lerna filter-options](https://www.npmjs.com/package/@lerna/filter-options)
- [yarn v2 foreach include/exclude](https://yarnpkg.com/cli/workspaces/foreach)
- [pnpm recursive --filter](https://pnpm.js.org/en/cli/recursive#filter-lt-package_selector)

accepted/0000-workspaces.md
bajtos: From my experience maintaining https://github.com/strongloop/loopback-next, filtering using package names is a bit of PITA when compared to filtering using directory names. Shells like Bash can autocomplete directory names for me, but not package names.

Compare typing directories with auto-completion:

```
lerna run --scope ex<TAB>amples/to<TAB>do
```

to typing package names with no auto-completion:

```
lerna run --scope @loopback/example-todo
```

It would be great if `npm` could support both flavors and/or provide shell auto-completion (as it already does in other places - see https://docs.npmjs.com/cli/completion).

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+│   └── package.json { "dependencies": { "dep-b": "^1.0.0" } }
+└── dep-b
+    └── package.json { "version": "1.3.1" }
+
+$ npm install
+
+// Results in this symlinking structure:
+├── node_modules
+│   ├── dep-a -> ./dep-a
+│   └── dep-b -> ./dep-b
```
bajtos: IIUC, this is proposing to link all sub-packages to root-level `node_modules` folder. I find such arrangement as rather problematic, because it allows sub-packages to require other sub-packages that are not listed in the dependencies. Such code will work at runtime when executed inside the monorepo (e.g. when running tests of the package), but will fail once the problematic package is installed from the registry (i.e. when consuming the package in 3rd-party projects).

As an example, consider the example presented in the RFC document, with an additional `dep-c` package with no dependencies and a file `dep-c/index.js` that's calling `require('dep-a')`.

Now when you run `node -e "require('dep-c')"` from monorepo root, all is good. When you run the same command from a project depending on `dep-c`, it will fail with `Error: Cannot find module 'dep-a'`.

In `lerna`, workspace sub-packages are never linked to root-level `node_modules`, the symlinks are always created between the packages only.

```
├── node_modules
├── dep-a
│   └── node_modules
│       └── dep-b -> ../../dep-b
├── dep-b
└── dep-c
```

As a result, `node -e "require('dep-c')"` executed from monorepo root fails.

I understand that hoisting all sub-packages in a shared top-level directory may have benefits. In that case I am proposing to use a different directory name than `node_modules`, e.g. `.workspace_modules`.

```
├── .workspace_modules
│   ├── dep-a -> ./dep-a
│   └── dep-b -> ./dep-b
├── dep-a
│   └── node_modules
│       └── dep-b -> ../../../.workspace_modules/dep-b
└── dep-b
```

accepted/0000-workspaces.md
bajtos: IMO, a single tag makes it difficult to look up code for a specific version when the multi-repo uses independent versioning strategy.

Using the example from the top of this RFC, let's say `dep-a` will be released in version `1.2.3` and `dep-b` will be released as `2.1.0`. How do you propose to build a tag name for such release? How can a person looking for code of `dep-b@2.1.0` find which git tag to check out?

ljharb: I strongly feel that independent versioning should be the default (and ideally, required) - an option that encourages a package to bump a version unnecessarily, only to make it match another package's version, has not been good for users (when babel did it, for example)

ruyadorno: tbh as of now I'm really leaning on the side of just removing `version` and `publish` completely from the scope of this initial workspaces implementation, hopefully we can have a final decision during the deep dive call next week and I'll come back to document it here 😊

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+> done
+
+> dep-b@1.3.1 test /Users/username/foo/dep-b
+> done
+```
+
+
+### 4. Publishing workspaces
+
+A workspace may not be published to the registry and for all publishing purposes having a valid `"workspace"` entry in a `package.json` is going to be the equivalent of `"private": true`.
```
raymondfeng: Please fix the typo: valid `"workspace"` entry => valid `"workspaces"` entry

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+## Examples
+
+### 1. Simplistic example expanding on symlinking structure and `package-lock` file shape.
+
+Given a workspaces setup with the following contents:
```
raymondfeng: It seems weird to call `a workspaces`.

accepted/0000-workspaces.md
raymondfeng: Isn't a singular `workspace` representing a collection of packages? Calling each nested package a workspace seems to be misleading.

accepted/0000-workspaces.md
ruyadorno: following up on this discussion, we had a quick brainstorm session in which we ended up seeing more benefits adopting [Yarn's approach](https://classic.yarnpkg.com/en/docs/cli/workspace): `npm workspace <workspace_name> <command>` OR `npm ws <workspace_name> <command>` to which I think we could also innovate a bit more and something fancier in the likes of: `npm :<workspace_name> <command>`, so for example running `test` in a specific **workspace** named `a` with these different ideas:

- `npm workspace a test`
- `npm ws a test`
- `npm :a test`

let's follow up with the discussion on [tomorrow's Open RFC call](https://github.com/npm/rfcs/issues/111)

accepted/0000-workspaces.md
ruyadorno: thanks @raymondfeng I did a round of cleaning up the RFC in which I defined a **Dictionary** section and reused the exact same words everywhere (to which people can refer at the dictionary section of the RFC) in order to make the entire conversation more straightforward

to answer to your question, the collection of packages are named **workspaces** (so each package of this collection is referred to as a **workspace**), the root-level package in which you define the configurations for working with **npm workspaces** is the **Top-level workspace**

from all the documentation I read across Yarn and pnpm, this terminology seems to reflect better the relationship the community is already using (feel free to correct me if I'm wrong)

accepted/0000-workspaces.md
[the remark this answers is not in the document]
ruyadorno: I commented about it in a couple of places: [here](https://github.com/npm/rfcs/pull/103#discussion_r383477860) and [here](https://github.com/npm/rfcs/pull/103#discussion_r393835251)

bottom line here is, remember to come to the [OpenRFC call tomorrow](https://github.com/npm/rfcs/issues/111) 😄

accepted/0000-workspaces.md
ruyadorno: as announced/discussed in the [OpenRFC Deep Dive call](https://github.com/npm/rfcs/issues/106), `version` and `publish` are not going to be supported in this initial **npm workspaces** implementation.

accepted/0000-workspaces.md
ruyadorno: as discussed in the [OpenRFC Deep Dive call](https://github.com/npm/rfcs/issues/106) this is no longer going to be the case - RFC have been updated to reflect this change.

isaacs:
Work is progressing on this over here: https://github.com/npm/arborist/pull/50

accepted/0000-workspaces.md
ruyadorno: @danielleadams let's move the discussion to the follow up PR #117 😊 

it elaborates a bit more on filtering **workspaces** and propose some solutions such as introducing a way to group them

accepted/0000-workspaces.md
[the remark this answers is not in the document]
ruyadorno: documented 😊

accepted/0000-workspaces.md
ruyadorno: let's move the conversation to #117 it evolves a bit more the ideas we gathered around it and I'd love to have some feedback there 😊

accepted/0000-workspaces.md
ruyadorno: ...in that call it was decided to completely remove the support to running sub commands from the scope of this RFC and open a follow up instead: #117

accepted/0000-workspaces.md
ruyadorno: Note: For the sake of making the entire conversation around **workspaces** more clear we went ahead and defined a **Dictionary** section that you can see [over here](https://github.com/npm/rfcs/blob/8a5080c881a771d25638d66e5150e4ea8e39c7a2/accepted/0000-workspaces.md#dictionary) - it defines the "main directory" as being the **Top-level workspace**. That said, any script in the **Top-level workspace** can be run as usual using `npm run` BUT it will still be necessary to use some sort of userland module such as [Lerna](https://lerna.js.org/) to help out with running `scripts` across nested **workspaces**.

With regards to the provided example, I think this is exactly the kind of structure we're aiming to support so def I see no issues 😊 as I mentioned in previous comments, the nested **workspaces** `run-scripts` support and other **npm commands** are going to be tackled in a follow up implementation defined at #117

lmcarreiro:
From those three use cases that @ljharb pointed out at the beginning, the third one is the most important:

> 3. monorepo
> - basically nothing is published, everything is private: true
> - the rest is the same as in a monorepo, except that everything's always linked because there's no published versions

The first two are already supported by third party tools, like Yarn, Lerna and Microsoft Rush.

Breaking a private big project into a bunch of small private pieces is way more common than breaking it into public packages. And yet we don't have a simple solution to such a simple problem. People don't do this in Javascript projects because there is no simple way to do that. So in any big project with a single package.json, every new installed dependency are available on the entire source code, which is a nightmare from the point of view of software architecture.

During the install, if some dependency is a workspace, it should be handled differently if it is a local dev install `npm install` or a production install `npm install --production`, and if it is private or not:
- **On a dev machine, without `--production`:** just create symbolic links an it will work fine.
- **On a server, with `--production`:** have to take a look at the local package.json file
   - `private: false`: just fetch the dependency from the registry
   - `private: true`: pack the dependency and use the resulting `.tgz`, and if this dependency depends on other private workspaces, this should be done recursivelly

Yarn v2 introduced the `workspace:` protocol. It creates a link on a dev install, but on prod install it always tries to fetch from the registry, even when the dependency has `private: true`.

remorses:
Is someone working on this? A lot of people would migrate from yarn as soon as this feature is implemented

ruyadorno:
@lmcarreiro that sounds like an interesting idea! I'll bring that up during the next OpenRFC call but as work on this RFC has been finalized it might be better to follow up with it in #117 

@remorses that'd be me 😬 you can follow up progress in the link @isaacs shared above or just follow up v7 progress in **npm/arborist** in general 😊

ruyadorno:
I noted the PR doesn't contain all the historic info on the ratification process, so just for the sake of documenting these details I'm going to leave these references here:

- This RFC was discussed in length during the [OpenRFC Deep Dive call of March 10](https://github.com/npm/rfcs/issues/106) and the [OpenRFC meeting of April 1st](https://github.com/npm/rfcs/issues/122) (and previous ones)
- RFC was ratified at the [OpenRFC meeting of April 15](https://github.com/npm/rfcs/issues/130)

smably:
I'm a little late to the party here, but reading the RFC it seems that there will be a single `package-lock.json` in the root. Is there a migration path for projects currently using Lerna where there is one `package-lock.json` file per leaf package? I.e., will there be a way to maintain the existing locked versions when introducing npm workspaces to a project managed by Lerna?

ruyadorno:
@smably that's a very good question I would love to hear @evocateur thoughts about it, since I believe there might be tooling out there from the Lerna community around it from the time Yarn v1 first shipped workspaces.

That said the idea is that once **npm@7** is shipped, Lerna+npm users will now have the option to migrate to a [useWorkspaces bootstrap setup](https://github.com/lerna/lerna/tree/master/commands/bootstrap#--use-workspaces) similar to the way Lerna+Yarn users were already familiar with - migrating a project to that kind of setup might be non-trivial though, the mismatching `package-lock.json` problem you just described is a side-effect of a much larger change which is that now you only have a single installation tree and that can lead to problems like mismatching peer dependencies (the one I saw most often having tried a dozen different projects).

smably:
@ruyadorno Makes sense. I don't think there's a way to merge multiple Yarn lockfiles because Yarn's lockfile format maps each semver range directly to a resolved version, so it's not possible to maintain different resolutions for the same semver range in different workspaces. E.g., one of my (pre-workspaces) packages could have in its `yarn.lock`
```
lodash@^4.17.5:
  version "4.17.5"
  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.5.tgz#99a92d65c0272debe8c96b6057bc8fbfa3bed511"
  integrity sha512-svL3uiZf1RwhH+cWrfZn3A4+U58wbP0tGVTLQPbjplZxZ8ROD9VLuNgsRniTlLe7OlSqR79RUehXgpBW/s0IQw==
```
and another could have
```
lodash@^4.17.5:
  version "4.17.15"
  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.15.tgz#b447f6670a0455bbfeedd11392eff330ea097548"
  integrity sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A==
```
and there'd be no way to represent both of those with a single `yarn.lock` in the root.

Whereas in this RFC, if I'm reading the `package-lock` example correctly, it appears that each workspace has its own nested `dependencies` tree, so it would be feasible to merge multiple lockfiles and maintain the original resolutions within each workspace.

ruyadorno:
> Whereas in this RFC, if I'm reading the `package-lock` example correctly, it appears that each workspace has its own nested `dependencies` tree, so it would be feasible to merge multiple lockfiles and maintain the original resolutions within each workspace.

Right right, the npm lockfile format preserves information about nested dedupped modules which makes it _theoretically_ possible to have them merged but you would still need some manual intervention in order to handle `peerDependencies` and any top-level conflicts - either way that would be def outside of the scope of the work we're doing in the npm cli now 😊 awesome tooling for someone in the community to pick up though 😉 😉

accepted/0000-workspaces.md
zkochan: @ruyadorno pnpm is using "workspace" as a set of projects. I don't understand why Yarn decided to use "workspace" as a synonym for package (or project).

I also did a small poll. Most of pnpm users believe that a workspace should be a set of projects: https://twitter.com/ZoltanKochan/status/1214223464363675649

Also, pnpm stores the package globs in a separate file called `pnpm-workspace.yaml`, which defines the root of the workspace. It is much easier to find the root of the workspace by looking for `pnpm-workspace.yaml` vs looking for a package.json with a workspace field. A pnpm workspace may even be without a package.json in the root.

zkochan: Seems like even @isaacs agreed with the definition used by pnpm: https://twitter.com/izs/status/1214305689977253888

PR #114, opened 2020-03-18
RFC: Expand list of ignored files

ruyadorno:
TLDR; Expand the default list of files that should not end up in published tarball with reasonable patterns used by the JS community.

[See RFC](https://github.com/ruyadorno/rfcs/blob/5ee1fe0e763cd7d829c86fe82cf306ed838db22e/accepted/0000-expand-list-of-packlist-ignored-files.md)


ruyadorno:
discussion started here: https://github.com/npm/npm-packlist/issues/48 and the @npm/cli-team decided it would be nice to throw a proper RFC to discuss this a bit more with the community.

accepted/0000-expand-list-of-packlist-ignored-files.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Motivation
+
+A large population of npm users are concerned about package sizes and with the advent of the file explorer now available on [npmjs.com](https://www.npmjs.com/) we can now see a number of common files that are very intrinsic to the JS community that we could start ignoring from package bundles without too much friction to the larger ecosystem.
+
+## Detailed Explanation
+
+Expand the current list of ignored files to also ignore by default:
+
+- `.editorconfig` common plugins
```
ljharb: This will cause npm test to fail, when it checks the editorconfig (as many of my packages do)

accepted/0000-expand-list-of-packlist-ignored-files.md
```diff
@@ -0,0 +1,37 @@
...
+- `.gitattributes` and/or more git things
+- `.idea/` (or other editors similar configs/store/etc)
+- `.travis.yml`, `.github/` (and/or more ci services)
+- `.yo-rc.json` template/boilerplate related files
+
```
wesleytodd, line 20: `.env` files would be another good addition.

ruyadorno:
Action items from the [OpenRFC call](https://github.com/npm/rfcs/issues/122):
- Restrict this RFC to only items that can be potentially harmful to the ecosystem (high potential of leaking secrets, etc)
- Let's create a new **warning** list that would prompt publishers at publish time

ruyadorno:
It would be awesome if folks can contribute to this by suggesting potential files that they notice have high potential to be harmful if published by mistake to the registry 😊

accepted/0000-expand-list-of-packlist-ignored-files.md
ljharb: this file is still in the list, but it's been marked as resolved

ljharb:
What about also ignoring `*.sublime-project`, `bower.json`, `component.json`?

ruyadorno:
from @bnb in OpenRFC call:

> We could borrow items from: https://github.com/github/gitignore/blob/master/Node.gitignore

accepted/0000-expand-list-of-packlist-ignored-files.md
```diff
@@ -0,0 +1,37 @@
...
+A large population of npm users are concerned about package sizes and with the advent of the file explorer now available on [npmjs.com](https://www.npmjs.com/) we can now see a number of common files that are very intrinsic to the JS community that we could start ignoring from package bundles without too much friction to the larger ecosystem.
+
+## Detailed Explanation
+
+Expand the current list of ignored files to also ignore by default:
+
+- `.editorconfig` common plugins
+- `.gitattributes` and/or more git things
+- `.idea/` (or other editors similar configs/store/etc)
+- `.travis.yml`, `.github/` (and/or more ci services)
```
rwjblue, line 18: FWIW, removing `.travis.yml`/`.github/` files would be a breaking change for some packages that use the CI configuration in order to provide users with useful messaging about supported platforms (where the supported platforms are defined as platforms that are under test). 

An example would be [this code in ember-cli](https://github.com/ember-cli/ember-cli/blob/v3.18.0/lib/utilities/platform-checker.js), which is used to emit warnings like:

> Node ${process.version} is no longer supported by Ember CLI. We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.

Or:

> Node ${process.version} is not tested against Ember CLI on your platform. We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.

ljharb: wow, TIL people do this.

altho in this case, https://github.com/ember-cli/ember-cli/blob/v3.18.0/.npmignore#L18 would ensure ember-cli is not broken by this change.

rwjblue: > wow, TIL people do this.

Ya, our goal was to ensure we could use a single source of truth for this guidance. We could remove it I suppose, but it has come in helpful for things like new Node releases that are untested with their release.

> altho in this case, ember-cli/ember-cli@v3.18.0/.npmignore#L18 would ensure ember-cli is not broken by this change.

Ya, good point. As long as the `.npmignore` "wins" (including negation) this won't be an issue. Thank you for looking at that!

accepted/0000-expand-list-of-packlist-ignored-files.md
```diff
@@ -0,0 +1,37 @@
...
+## Rationale and Alternatives
+
+Avoiding bundling undesirable files is something we already do today, the idea is only to make it more useful by including some other common files in the JavaScript ecosystem. That said, possible alternatives are:
+
+- Status quo, do not alter the current existing [list of ignored files](https://github.com/npm/npm-packlist/blob/master/index.js#L38).
+- More alternatives?
+
+## Implementation
+
+Add some more entries to the already existing [list of ignored files in packlist](https://github.com/npm/npm-packlist/blob/master/index.js#L38) and make sure we have tests asserting it behaves the way we intend.
```
vweevers, line 32:
```suggestion
Add some more entries to the already existing [list of ignored files in packlist](https://github.com/npm/npm-packlist/blob/main/index.js#L38) and make sure we have tests asserting it behaves the way we intend.
```

accepted/0000-expand-list-of-packlist-ignored-files.md
```diff
@@ -0,0 +1,37 @@
...
+
+Let's expand the default list of ignored files in [packlist](https://github.com/npm/npm-packlist).
+
+## Motivation
+
+A large population of npm users are concerned about package sizes and with the advent of the file explorer now available on [npmjs.com](https://www.npmjs.com/) we can now see a number of common files that are very intrinsic to the JS community that we could start ignoring from package bundles without too much friction to the larger ecosystem.
+
+## Detailed Explanation
+
+Expand the current list of ignored files to also ignore by default:
```
vweevers, line 13: Do the items listed here mean e.g. `/.editorconfig` or `**/.editorconfig`? The latter could cause breakage in generators / initializers that have template files in subdirectories, for example https://github.com/pkgjs/create/tree/main/templates

webstech:
Is there any incentive to move forward with this?  I was shocked to find how many dependabot and workflow files I had in various repos today.  Yarn already excludes `.github` during pack/publish so there is precedent.  Internally there seems to be an established [list](https://github.com/npm/cli/blob/7a5fd01647766352b6116e3144a82527ed0866e2/scripts/bundle-and-gitignore-deps.js) of well known files to blacklist.  I found 100+ flavours of dot filenames in local `node_modules` directories.  It may be more code but a warning to suggest using `.npmignore` or `package.files` could help with migration.  I understand any warnings may not be seen in automation runs - just trying to get some of this to happen.

ljharb:
Please do not include `.editorconfig` in this list; I run a tool in my packages that lints against it, and i rely on that file being present.

PR #117, opened 2020-03-19
RFC: npm workspaces - Working with workspaces

ruyadorno:
Adds a workflow to working with subcommands across workspaces.

[See RFC](https://github.com/npm/rfcs/blob/767407bc8b55c3afdf0cd0c7e042c2beb06e15af/accepted/0000-workspaces-part2.md)

### DIFF
--- a/accepted/0000-workspaces-part2.md
+++ b/accepted/0000-workspaces-part2.md
@@ -124,7 +124,7 @@ It's possible to define a **groups** property within your **workspaces configura
     "version": "1.0.0",
     "workspaces": {
         "groups": {
-            "core": ["core/*"], // accepts fs location
+            "core": ["core/*"], // accepts globs
             "plugins": ["lorem", "ipsum"], // also accepts workspaces names
             "common": ["util", "helpers"]
         },
@@ -157,6 +157,12 @@ Further example, install a **peer dependency** across a group of packages named 
 $ npm workspace core install react@16 --save-peer
 $ npm ws core install react@16 --save-peer
 $ npm :core install react@16 --save-peer
+```
+
+It should also be possible to define groups as cli arguments, so that the following example will be equivalent to the previous examples:
+
+```
+$ npm workspace core install react@16 --save-peer --workspace-group core=core/*
 ```
 
 ## Prior Art
@@ -193,4 +199,13 @@ During the discussions around this RFC it was brought up to our attention that a
 
 ## Unresolved Questions and Bikeshedding
 
-TBD
+- Filter syntax:
+    - `npm workspace foo test`
+    - `npm ws foo test`
+    - `npm :foo test`
+    - `npm --workspace=foo test`
+    - `npm -w=foo test`
+- Groups as a cli argument: `--workspace-group`
+    - how to define the `key=value` nature of it?
+    - is there any arg currently in the cli using key/value pairs?
+    - maybe it's not worth having it?
--- a/accepted/0000-workspaces-part2.md
+++ b/accepted/0000-workspaces-part2.md
@@ -1,52 +1,52 @@
-# npm workspaces: Running commands
+# npm Workspaces: Running commands
 
 ## Summary
 
-Introduces basic support to running **npm commands** across nested **workspaces**.
+Introduces basic support for running **npm commands** across a project's **workspaces**.
 
 ## Motivation
 
-The need for running **npm commands** across nested **workspaces** was sourced from the community during the elaboration of the [original **npm workspaces** RFC](https://github.com/npm/rfcs/pull/103).
+The need for running **npm commands** across a project's **workspaces** was surfaced by the community during the discourse spun out of the [initial **npm workspaces** RFC](https://github.com/npm/rfcs/pull/103).
 
 ## Detailed Explanation
 
-Following up from the original **npm workspaces** RFC are the following command-related changes:
+Following up on the original **npm workspaces** RFC are the following command-related additions/changes:
 
-- A subset of the standard **npm commands** needs to be made aware of **npm workspaces**
+- A subset of the standard **npm commands** need to be made aware of **npm workspaces**
 - It's desired to have a standard way to filter out **workspaces** in which to run specific **npm commands**
 
 ## Rationale and Alternatives
 
-The ability to run **npm commands** across defined **workspaces** is essential to successfully manage a **npm workspaces** workflow.
+The ability to run **npm commands** across a project's **workspaces** is essential to effecient management of, & enabling, complex developer workflows.
 
-Following are possible alternatives to provide this functionality:
+The following are possible alternatives to unlock this functionality:
 
 - Do not implement further support to manage running commands across **workspaces**, keep the **npm workspaces** set of features to a minimum
-- Defer to other tools from the ecosystem such as Lerna to solve the problem of running top-level commands across **workspaces**
-- Implement the proposed features of this RFC as a standalone package / separated tool from the **npm cli**
+- Defer to other tools from the ecosystem - such as **Lerna** - to solve the problem of running top-level commands across **workspaces**
+- Implement the proposed features of this RFC as a standalone package apart from the **npm cli**
 
 ## Implementation
 
-### 1. Run commands across all child packages
+### 1. Run commands across all child workspaces
 
-Create a new **npm cli** subcommand named **workspaces** (aliased to **ws**) that can route any of the supported subcommands to run in the context of the configured **workspaces** as long as a **workspaces configuration** field is properly defined in `package.
+Create a new **npm cli** config, `--workspaces` (aliased to `--ws`), that can route any of the supported subcommands to run in the context of the configured **workspaces** as long as a **workspaces configuration** field is properly defined in `package.json`
 
-We identified 5 different categories of subcommands based on how they're expected to work:
+There are currently five distinct categories of _"workspace-awareness"_ that existing subcommands will belong to:
 
-#### 1. Commands that just need context from package.json
+#### 1. Commands that _just_ need context from `package.json`
 
-Commands that, from an user point of view, are the equivalent of: `cd <workspace-name> && npm <cmd>`.
+Commands that, from a user's point of view, are the equivalent of: `cd <workspace-name> && npm <cmd>`.
 
 - `docs`
 - `doctor`
-- `diff` Package diff in the context of specific **workspaces**
-- `dist-tag` List dist-tags for specific **workspaces**
-- `pack` Run pack in the context of specific **workspaces**
-- `publish` Run publish in the context of specific **workspaces**
+- `diff`
+- `dist-tag`
+- `pack`
+- `publish`
 - `repo`
 - `set-script`
 - `unpublish`
-- `view` View registry info, also including **workspaces**
+- `view`
 
 #### 2. Custom implementation
 
@@ -56,13 +56,13 @@ General class of helper commands that load from the installed tree to produce so
 
 - `audit`
 - `explain`
-- `fund` List funding info for all **workspaces**
-- `ls` List all packages including **workspaces**
-- `outdated` List outdated **dependencies** including **workspaces** and its **dependencies**
+- `fund`
+- `ls`
+- `outdated`
 
 ##### 2.2. Modify installed dependency tree
 
-The set of commands that will modify an install tree (from an implementer point of view, these are just [arborist.reify](https://github.com/npm/arborist/) proxies).
+The set of commands that will modify an install tree (from an implementation stand-point, these are just [arborist.reify](https://github.com/npm/arborist/) proxies).
 
 - `ci`
 - `dedupe|find-dupes`
@@ -70,23 +70,22 @@ The set of commands that will modify an install tree (from an implementer point 
 - `install-test`
 - `install`
 - `link`
-- `rebuild` Rebuild all **workspaces**
-- `update` Updates a **dependency** across the entire installation tree, including **workspaces**
-- `uninstall`
+- `rebuild`
 - `update`
+- `uninstall`
 
 ##### 2.3. Other
 
-A command that needs a special/custom workspace-aware implementation outside of the context of reading/writing to the install tree (using **Arborist**).
+Commands that need a special/custom _"workspace-aware"_ implementation outside of the context of reading/writing to the install tree.
 
-- `exec` Run exec in the context of specific **workspaces**
-- `init` Initialize a new **workspace**
-- `run-script|restart|start|stop|test` Run arbitrary **scripts** in all **workspaces**, skip any **workspace** that does not have a targetting **script**
-- `version` Run version in the context of specific **workspaces**, needs tweaked commit msg, tag
+- `exec`
+- `init`
+- `run-script|restart|start|stop|test`
+- `version`
 
 #### 3. Unsupported
 
-This category of **npm cli** subcommand is completely unrelated to anything that the current working directory could affect. All registry helper/management types of commands fall into this category and it's a best UX to just exit with an error code in order to let the end user aware that trying to run these in the context of workspaces don't have any effect.
+This category of **npm cli** commands is completely unrelated to anything that the current working directory could affect. This includes a number of registry-specific helper/management operations which don't make sense/have no effect within the context of a workspace. Trying to run these commands with a workspace flag/config will exit with a descriptive error code.
 
 - `adduser|login`
 - `bin`
@@ -126,7 +125,7 @@ Running tests across all configured **workspaces**:
 └── dep-b
     └── package.json { "version": "1.3.1" }
 
-$ npm ws test
+$ npm test --workspaces
 
 > foo@1.0.0 test /Users/username/foo
 > echo "Error: no test specified" && exit 1
@@ -145,11 +144,10 @@ npm ERR! Test failed.  See above for more details.
 
 Filter is done via named argument (`--workspace`, short: `-w`) and here are some of the reasons why we decided to go that route:
 - [Lerna filters](https://www.npmjs.com/package/@lerna/filter-options) were the starting point but early in the discussion some other considerations were brought forward, specially the many pitfalls of supporting globs in the variety of supported shells and operational systems.
-- npm configs do have some baggage, configs might apply to to other cli commands and can also possibly be defined project or system-wide so the most sure way to avoid these pitfalls is for configs to only be used in the context of a single **npm cli** subcommand.
 
 ### 3. Examples
 
-Given the results of all this preliminar work, the preferred way to run a command in the context of a single workspace is to use the named `--workspace` argument or its `-w` short alias, e.g:
+Given the results of all this preliminar investigation, the preferred way to run a command in the context of a single workspace is to use the named `--workspace` argument or its `-w` short alias, e.g:
 
 In a project with the following structure:
 ```
@@ -165,20 +163,20 @@ In a project with the following structure:
 You can run tests for the `foo` workspace from the root of your project, with the following syntax:
 
 ```
-npm ws test --workspace=foo
+npm test --workspace=foo
 ```
 
 
 #### Install dependency example:
 
-Add `tap` as a **dependency** of all your configured **workspaces**:
+Add `tap` as a **dependency** for all of your configured **workspaces**:
 
 ```sh
-npm ws install tap
+npm install tap --workspaces
 ```
 
 
-Note: **Globs** are not supported as a valid `--workspace` argument value, the proposed alternative is to use `npm ws <cmd> --workspace=<dir>` in which `<dir>` is a folder containing multiple workspaces.
+> Note: **Globs** are not supported as a valid `--workspace` argument value, the proposed alternative is to use `npm <cmd> --workspace=<dir>` in which `<dir>` is a folder containing multiple workspaces.
 
 ## Prior Art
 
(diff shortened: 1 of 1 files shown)


accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,188 @@
...
+
+> dep-b@1.3.1 test /Users/username/foo/dep-b
+> done
+```
+
+**NOTE:** `publish` and `version` are very special cases in a **npm workspaces** setup and are outside of the scope of this RFC. These features will live in userland in the form of popular modules such as [Lerna](https://lerna.js.org/) for the foreseeable future.
+
+### 2. Filter a subset of workspaces
+
+Rather than defining a strict UX, the current proposal is to adhere to a `workspace` command prefix in the form of either: `npm workspace <workspace_name> <command>` OR `npm ws <workspace_name> <command>` OR an even shorter shortcut using a symbol `npm :<workspace_name> <command>`, for example running `test` in a specific **workspace** named `foo` with these different ideas:
```
wesleytodd, line 75: I know I am a bit late to this, sorry backed up notifications.  But I am not a huge fan of adding special syntax for this when chaining some commands together can do the same.  I think you could do this with a `npm worksapce-list` command and `grep`/`xargs` right?  Seems like it is just more complexity for something with an already workable solution.

ljharb: That's not available on every node-supported platform, and using pipes wouldn't be able to easily use npm configuration to define what a project is.

wesleytodd: Sure, but I am sure there is viable alternatives, right?  Can you give an example for what you mean on "use configuration to define what a project is" means?

ljharb: Sure - maybe I want to configure workspaces to ignore projects that start with `example-*`. If I was forced to use things like grep, i'd have to duplicate that config in my piped commands.

wesleytodd: Conversation resolved via RFC call

ruyadorno: 👍 thanks @ljharb and @wesleytodd

ruyadorno:
Action items from OpenRFC call:
- Document backstory on why we want to use positional arguments vs named config in the RFC
- Maybe also document in some form the initial discussion from the [OpenRFC call](https://hackmd.io/urLMaymHRQ2Z94oEYTmBbg)

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+Following up with the previous configuration example, it would be possible to run tests across all **workspaces** from the `plugins` group, as following:
+
+```
+// one of these:
+$ npm workspace plugins test
+$ npm ws plugins test
+$ npm :plugins test
```
wesleytodd, line 137: I am not a huge fan on this syntax.  It is mysterious and I would prefer either `workspace` or `ws`.

evocateur: On the other hand, I use `npm i` and `npm t` several dozen times a day, on average. Shorthands have their benefits (and they appear completely additive in this case).

ljharb: a long of "workspace" and a short of "ws" makes sense to me; making the short version do something different does not.

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,196 @@
...
+> dep-b@1.3.1 test /Users/username/foo/dep-b
+> done
+```
+
+**NOTE:** `publish` and `version` are very special cases in a **npm workspaces** setup and are outside of the scope of this RFC. These features will live in userland in the form of popular modules such as [Lerna](https://lerna.js.org/) for the foreseeable future.
+
+### 2. Filter a subset of workspaces
+
+Filter is done via positional argument and here are some of the reasons why we decided to go that route:
+- [Lerna filters](https://www.npmjs.com/package/@lerna/filter-options) were the starting point but early in the discussion some other considerations were brought forward
```
evocateur, line 76: yeah, don't copy those please, i rue the `--scope` flag.

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,196 @@
...
+#### Previously
+
+- [npm workspaces](https://github.com/npm/rfcs/blob/de8d71c0453f5cf443d3ef2f47e313f12dd6aaf9/accepted/0000-workspaces.md)
+
+#### Filtering a subset of workspaces
+
+- [lerna filter-options](https://www.npmjs.com/package/@lerna/filter-options)
+- [Yarn v1 workspace](https://classic.yarnpkg.com/en/docs/cli/workspace)
+- [Yarn v2 foreach include/exclude](https://yarnpkg.com/cli/workspaces/foreach)
+- [pnpm recursive --filter](https://pnpm.js.org/en/cli/recursive#filter-lt-package_selector)
```
evocateur, line 173: Looks like this moved to https://pnpm.js.org/en/filtering (I really like pnpm's approach, tbh)

zkochan: In the past, we used to also support passing the filters after `--`. So for instance, you could do `pnpm update -- foo bar` would update dependencies in `foo` and `bar`. I think that was pretty convinient but it caused confusion because some commands treat `--` differently.

I had some ideas [here](https://github.com/pnpm/pnpm/issues/2012) to make a unified syntax but in the end we decided to deprecate `--` and only allow passing the filters through one or more filter flags.

The filter flag can go before the command name, so you can run `pnpm --filter foo update`

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,196 @@
...
+- defining a subset of workspaces in a `.npmrc` file might be just a footgun leading users into confusion
+- Yarn v1 uses positional arguments for filter in the form of `yarn workspace <workspace_name> <cmd>`, so there's precedent
+
+Leading this RFC to a proposal using positional arguments. Rather than defining a strict UX, the current proposal is to adhere to a `workspace` command prefix in the form of either: `npm workspace <workspace_name> <command>` OR `npm ws <workspace_name> <command>` OR an even shorter shortcut using a symbol `npm :<workspace_name> <command>`, for example running `test` in a specific **workspace** named `foo` with these different ideas:
+
+- `npm workspace foo test`
+- `npm ws foo test`
+- `npm :foo test`
+
+Leaving space to gather feedback from the community on whether to introduce one of this proposed commands or even introducing some novel interface.
```
zkochan, line 87: I don't know if there are other examples of such syntax outside of Yarn. For me, a flag would look more usual. It can be something short like `npm -w foo test`. If you want to run tests in foo and bar, you could do either `npm -w foo -w bar test` or `npm -w foo,bar test`

accepted/0000-workspaces-part2.md
ljharb: can you elaborate on why? (do you mean the functionality, or the API?)

accepted/0000-workspaces-part2.md
ruyadorno: I really like using a shorter alias such as `-w` but the reason to steer away from named configs has more to do with the [specific **npm** behavior](https://github.com/npm/rfcs/pull/117/files#diff-828bb29c16ce81ff3a4108b6e8dab945R78)

zkochan: can't npm treat the `--workspace` flag differently? Not reading it from the rc file?

accepted/0000-workspaces-part2.md
ruyadorno: thanks for sharing that bit of context @zkochan 😊 

I see **pnpm** `--filter` also supports `<directory>` (which I suppose maybe also `globs` along with that?) what I proposed initially in this RFC is to avoid these in favor of a separate "aliases" OR "groups" definition that can also be defined in `package.json` for ease to share/reuse while also decoupling that categorization from the fs layout

zkochan: Directories must start with `./` or be inside `{}`

For instance, `--filter ./packages` or `--filter {packages}`. Also, they don't support globs. They select all packages inside the specified directories.

zkochan: also, you may combine selectors. Like you may filter by all packages that start with `babel-` and are inside some dir: `--filter babel-*{packages/subdir}`

accepted/0000-workspaces-part2.md
ruyadorno: this is related to the conversation going on at https://github.com/npm/rfcs/pull/133#issuecomment-621397885 and honestly we might just spin off an RFC/discussion around a new config management system for the cli

accepted/0000-workspaces-part2.md
evocateur: 83% the API surface (too many flags, bad names), 29% the prickliness of "raw" globs in a CLI environment (when do you need to quote them, which accidental evaluation caused unintended behavior, et cetera).

wesleytodd:
One thing not mentioned in the RFC is what happens when a command is run in a sub-package.  The assumption is that it ignores the workspace and just runs in the current package, I confirmed this with @ruyadorno via slack, but it is not explicit in this doc.  Maybe it is not necessary to add, but it was context I was missing.

ruyadorno:
Notes from the OpenRFC call:

- make sure groups/alias also have an alternative way to be declared as cli args

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,211 @@
...
+- Filter syntax:
+    - `npm workspace foo test`
+    - `npm ws foo test`
+    - `npm :foo test`
+    - `npm --workspace=foo test`
+    - `npm -w=foo test`
+- Groups as a cli argument: `--workspace-group`
+    - how to define the `key=value` nature of it?
+    - is there any arg currently in the cli using key/value pairs?
+    - maybe it's not worth having it?
```
ruyadorno, line 211: @evocateur I'd really appreciate your input on this one, this idea was surfaced during the life of this RFC but I'm afraid it might be an invite to all the category of problems you listed with `--scope`

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,196 @@
...
+
+**Note:** Alternative ways of executing each of the described features are defined along with their implementation details below.
+
+## Implementation
+
+### 1. Run commands across all child packages
+
+Make **npm cli** subcommands **npm workspaces**-aware, so that running a command tries to run it within all **workspaces** as long as a **workspaces configuration** field is properly defined in `package.json`.
+
+Only a subset of commands are to be supported:
```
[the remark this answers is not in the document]
ruyadorno: yeah, I think what we need here is a good definition of what is a package-related command vs all the other administrative kind of commands

it makes sense to suppport `--workspaces=foo` in commands such as `npm test`, `npm ls` but it makes no sense in admin types of commands such as `npm adduser`, `npm token`, etc

so that was my goal here with the "subset of commands" definition 😊 I'm def open to reword that!

accepted/0000-workspaces-part2.md
ruyadorno: thanks @zkochan we ended up going with that option! 😄

ruyadorno:
RFC updated to refer to named config choice for filter, e.g: `npm test --workspace=foo`

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,221 @@
...
+### 2. Filter a subset of workspaces
+
+Filter is done via named argument (`--workspace`, short: `-w`) and here are some of the reasons why we decided to go that route:
+- [Lerna filters](https://www.npmjs.com/package/@lerna/filter-options) were the starting point but early in the discussion some other considerations were brought forward
+- npm configs do have some baggage, configs might apply to to other cli commands and can also possibly be defined project or system-wide - that said, there are things we can do to avoid these pitfalls (such as making sure `npm_config_workspace` env varibable never gets set in `run-script`)
+- During the many discussions around this RFC a few strong points for named arguments stand out:
+  - Allow us to keep `npm workspace` for other command usages
+  - npm users are more familiar with usage of named configs, example: `npm test --workspace=foo` is very similar to `npm test --prefix=./packages/foo` - if you think on it in terms of "what workspace am I in" again, a named config makes more sense in npm
+  - [Organized a poll](https://github.com/npm/rfcs/issues/160) in which named arguments were the leading option, here are the results for reference:
```
mikestopcontinues, line 81: This data shouldn't be included. The sample size is too small and more than likely non-representative.

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,221 @@
...
+It's possible to define a **groups** property within your **workspaces configuration** that will allow defining named sets of **workspaces** that can be later on used to run commands, e.g:
+
+```
+{
+    "name": "workspace-example",
+    "version": "1.0.0",
+    "workspaces": {
+        "groups": {
+            "core": ["core/*"], // accepts globs
+            "plugins": ["lorem", "ipsum"], // also accepts workspaces names
```
mikestopcontinues, line 147: Does workspace name === subpackage name? There might be some ambiguity between directory structure and package names, so maybe, everywhere it's specified, locations should be relative paths (`./packages/loren`) and package names regular strings? This would go for CLI args as well as aliases. Either should be able to use globs.

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,221 @@
...
+#### Install dependency example:
+
+Add `tap` as a **dependency** of a **workspace** named `foo`:
+
+```sh
+npm install tap --workspace=foo
+```
+
+
+Note: **Globs** are not supported as a valid `--workspace` argument value, the alternative is to use `npm <cmd> --workspace=<alias>` in which the `alias` can be defined beforehand in your **workspace** configuration as detailed in the following up section.
```
mikestopcontinues, line 118: What about comma-separated values? EG `-w foo,bar`

ljharb: That's not very idiomatic/common for CLI args, in my experience.

bnb: yeah, I've not really seen that before and would not expect it.

accepted/0000-workspaces-part2.md
mikestopcontinues: If you look at the other bullet points, this section is exclusively an argument for the chosen filter style. In that context, the survey data is very misleading. Standards RFC trolls (like us, I guess) are far from average devs, and having only 35% of only 56 votes suggests that popular opinion is far from settled. On the other hand, this proposal has a lot of momentum and I think it's well-designed, so I don't see it making much difference either way.

accepted/0000-workspaces-part2.md
mikestopcontinues: As long as we've got aliases, this isn't an issue.

accepted/0000-workspaces-part2.md
mikestopcontinues: I think it's better dropped as well. There's plenty of `package.json` config you can't change via the CLI. Aliases should be a little hard to change anyway, since there can be big consequences to accidentally changing them. In the context of a deploy, it can take down a site or break a published package.

ljharb: Aliases shouldn’t be able to be changed, but absolutely from the cli i need to be able to scope to subpackages on the fly. Very few things in package.json that *npm* looks at are not configurable from the cli.

accepted/0000-workspaces-part2.md
ruyadorno: thanks for the review @mikestopcontinues - I do agree with @bnb in which I want to keep the track record of the process but I think you have a valid point in which maybe I should just reword this phrase a little to emphasize that this was not a critical factor in determining the final syntax.

accepted/0000-workspaces-part2.md
ruyadorno: the way [nopt](https://www.npmjs.com/package/nopt) works is that you specify multiple times that same item (and I don't have any intention to change from the default behavior here) so assuming we want to support multiple **workspaces** declarations the following should be valid: `npm <cmd> -w foo -w bar`

but @mikestopcontinues I think you got the idea of the design here in which I want to promote/steer usage of pre-defined aliases instead

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,221 @@
...
+### 1. Run commands across all child packages
+
+Make **npm cli** subcommands **npm workspaces**-aware, so that running a command tries to run it within all **workspaces** as long as a **workspaces configuration** field is properly defined in `package.json`.
+
+Only a subset of commands are to be supported:
+
+- `fund` List funding info for all **workspaces**
+- `ls` List all packages including **workspaces**
+- `outdated` List outdated **dependencies** including **workspaces** and its **dependencies**
+- `run-script` Run arbitrary **scripts** in all **workspaces**, skip any **workspace** that does not have a targetting **script**
```
jamiebuilds, line 38: I would highly suggest a number of different options to configure the parallelization of `run`

- `--parallel` spawn scripts across different packages in parallel processes (default to [physical cores](https://www.npmjs.com/package/physical-cpu-count))
- `--parallel=<count>` spawn `<count>` processes (as much as possible)
- `--parallel-nodes` split workload automatically across [CI Nodes](https://github.com/jamiebuilds/ci-parallel-vars)
- `--parallel-order=dependency-graph` limit parallelization by workspace cross-dependency graph, so that scripts only run on packages once all other packages have been run ([see `task-graph-runner` for example](https://github.com/jamiebuilds/task-graph-runner))
- `--parallel-order=any` Run scripts on packages in any order

Workspace-based monorepos have a huge need to optimize their different scripts and these options go a huge way to solve for different use cases.

jamiebuilds, line 38: `npm exec` would be another awesome addition for workspaces. Oftentimes workspace monorepos need to write scripts that run on every package. Needing to do stuff like this is super common:

```sh
npm exec -- $(pwd)/scripts/update-readme-toc.sh
```

This would evaluate the script `/path/to/project/scripts/update-readme-toc.sh` on every workspace directory as a separate process with `/workspace/node_modules/.bin:/node_modules/.bin` in the `PATH`

ruyadorno: this was brought up once again during today's OpenRFC call 😄 I'll make sure to add it to the RFC but it will probably live in `npm workspaces exec` (`npm ws exec`) since in **npm7** we already have a `npm exec` top-level command.

accepted/0000-workspaces-part2.md
```diff
@@ -0,0 +1,221 @@
...
+- `rebuild` Rebuild all **workspaces**
+- `restart`
+- `start`
+- `stop`
+- `test` Run test **scripts** in all **workspaces**
```
jasonwilliams, line 43: My expectation would be that `build` would be included here considering how much that is used in various projects. Is here a reason you’ve gone for `rebuild` but not `build`?

wesleytodd: `rebuild` is [command build into npm](https://docs.npmjs.com/cli/v6/commands/npm-rebuild).  I think what you are referring to is when users have a script called `build` in their package.json.  Right?  In which case I think you would run `npm workspace run build` (I think that is the right command, don't have all the new ones by memory yet).

jasonwilliams: Oh ok thanks for clearing that up

Christian24:
So, I will try again to phrase what I have said in the RFC call: 

Imagine a monorepo, which has `webpack` in the root and then in subsequent packages `webpack` is also a dependency. Now if you link the packages together, you do end up with multiple instances of `webpack`. So, if `webpack` or any other library does some kind of `instanceof` check it will most probably return false. 

It is discussed in detail in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/631#issuecomment-775257654 (there is even a reproduction with lerna). 

That being said I suppose @ljharb's suggestion of sharing `peerDependencies` should solve that, but I am not 100%, since the correct use of `peerDependencies` is still a bit nebulous to me :)

ljharb:
@Christian24 typically webpack, a dev dep, isn't a peer dep of workspace projects - but if you could elect to have it "shared" *as if* it were a peer dep, i think that would solve it.

Christian24:
I guess all I am essentially looking for is a fancy workspace dedupe. :)

jdrydn:
> I guess all I am essentially looking for is a fancy workspace dedupe.

Wouldn't you just omit `webpack` as a devDependency in `workspace` directories? Calls to `require('webpack')` would resolve to the top-level `node_modules` directory?

jsg2021:
> I guess all I am essentially looking for is a fancy workspace dedupe. :) 

It's already deduping? The root package.json just needs the `workspaces: []` key... all the leaf projects will currently hoist everything that doesn't conflict up to the root.

ljharb:
@jrylan that only works if every package in the workspace uses the same version of webpack. if some use v5 and some use v4, your only current option is duplication.

accepted/0000-workspaces-part2.md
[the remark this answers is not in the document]
ruyadorno: revamped all this session and categorized **all** top-level commands to make sure the support matrix is extra clear 😊

ruyadorno:
heads up, workspaces support went out in **npm@7.7.0** for the first couple of commands: `exec` and `run-script`

glen-84:
What is the status of `npm install <pkg> -w <workspace-name>`?

Is the only option to edit the workspace `package.json` files manually?

What is the ETA for this functionality?

(Unrelated: I find the naming a bit strange, as a workspace usually refers to a set of projects, not to the projects themselves.)

ruyadorno:
> What is the status of `npm install <pkg> -w <workspace-name>`?

⏳ in progress: https://github.com/npm/arborist/pull/257

> Is the only option to edit the workspace package.json files manually?

It's not necessarily the _only_ choice but I'd be tempted to say it's the safest, recommended choice for now.

> What is the ETA for this functionality?

It might land in the next couple of weeks, stay tuned for the upcoming releases.

> (Unrelated: I find the naming a bit strange, as a workspace usually refers to a set of projects, not to the projects themselves.)

I see your point and it's def a valid concern that I don't think it was mentioned before and in the RFC itself "workspace" is a noun used interchangeably meaning both the feature/project as **npm workspaces** and referring to nested packages as **workspaces**. It made me think about a bunch more reasons for picking the name that I'd like to document in the RFC too:
- The `workspace` config is an option that enables both "running in workspace mode" + "filter by these names/dirs"
- `-w` is an available shorthand and a much simpler one to relate to "workspaces" in a general sense

ljharb:
@ruyadorno i've brought it up a number of times; to me a "workspace" is the repo, and the individual chunks in it are "packages" (ie, the default directory name in lerna for these).

ruyadorno:
Sorry about it @ljharb I know you brought up the terminology confusion multiple times 😬

zkochan:
I agree with @ljharb and @glen-84 on this. In pnpm we use workspace for the repository and packages inside the workspace are called projects or packages.

Some IDEs use the same terminology I believe. And Angular projects also use the workspace terminology for noting sets of packages. I don't know why Yarn decided to use workspace for calling packages in a monorepository (Yarn calls the monorepository a project and the packages inside the monorepository are called workspaces).

Maybe you should make a poll and see what your users would prefer. When I did a poll between pnpm users, for the majority workspaces meant a set of packages or projects, not vice versa.

ruyadorno:
Thanks for the feedback @zkochan @ljharb @glen-84, while we def appreciate the intention of your suggestion, the consensus within the @npm/cli team is to keep the current nomenclature for now - we can def revisit this decision in the future but at this point I'm going to close this PR as the RFC discussed in here has been ratified (and many of the commands have actually already landed in the cli).

I would encourage anyone interested in adjusting that terminology to follow up either in the form of a RRFC issue or as a GitHub Discussion in this repo.

ruyadorno:
another important update to folks still subscribed to this thread, most of the remaining planned workspace support landed in [**npm@7.14.0**](https://github.com/npm/cli/releases/tag/v7.14.0), such as support to add/remove deps to workspaces, e.g: `npm install <pkg> -w <workspace-name>` 🎉 

cc @glen-84

glen-84:
@ruyadorno https://github.com/npm/statusboard/issues/303#issuecomment-845884373.

PR #121, opened 2020-03-28
Added proposal for package version `link#[version](comment)` syntax

kael-shipman:
# What / Why

Submitting an RFC through the standard process outlined in the readme. All details explained in RFC document.

Probably obvious, but much easier to view in [markdown format](https://github.com/kael-shipman/rfcs/blob/latest/accepted/0026-package-version-link-keyword.md)

> ## Summary
> Allow package versions like `link#[semver](comment)` (comment is optional) to indicate that npm should `npm link` the given package AND that the version in `package.json` of the linked repo must fulfill the given semver. If the package is not found among existing npm links, or if the linked package doesn't match the given semver, npm should throw an exception and output the comment. > 
> For example:
> 
> ```json
> {
>   "dependencies": {
>     "my-dep": "link#^3.0.0",
>     "nother-dep": "link#^5.15.0(clone github.com:cool-producer/nother-dep.git and check out branch great-improvements)"
>   }
> }

## References

* n/a

### DIFF
--- a/accepted/0026-package-version-link-keyword.md
+++ b/accepted/0026-package-version-link-keyword.md
@@ -1,0 +1,159 @@
+# Implement Special `link#[semver](comment)` Keyword for Package Version
+
+## Summary
+
+Allow package versions like `link#[semver](comment)` (comment is optional) to indicate that npm should `npm link` the given package AND that the version in `package.json` of the linked repo must fulfill the given semver. If the package is not found among existing npm links, or if the linked package doesn't match the given semver, npm should throw an exception and output the comment. For example:
+
+```json
+{
+  "dependencies": {
+    "my-dep": "link#^3.0.0",
+    "nother-dep": "link#^5.15.0(clone github.com:cool-producer/nother-dep.git and check out branch great-improvements)"
+  }
+}
+```
+
+## Motivation
+
+It's quite frequent that we develop several individual code libraries in concert, often using libraries in development across several other independent projects. There are several ways to facilitate this, all with serious drawbacks. The ones we will NOT discuss here are using git refs, using lerna, and using local paths, none of which are adequate solutions for a variety of reasons (see subsequent sections).
+
+The best way to handle the problem currently is by using `npm link`, but this has the following serious drawbacks:
+
+* There's no way to communicate to other developers how to obtain the correct dependencies, since the target code has not yet been published. If a new developer were to clone your repo and run `npm install`, they would end up with broken code, since the system would obtain the dependency at an earlier, inadequate version (or not at all, if the dependency is new). They would furthermore have no way of knowing how to fix the problem.
+* There's no way to enforce version conformance of linked repositories to ensure they fulfill the requirements. You can `npm link` any version of a package into your codebase and npm does no further checking.
+
+## Detailed Explanation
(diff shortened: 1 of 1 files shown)


kael-shipman:
@ljharb did I miss something in the process? It looks like someone submitted a PR above mine that got immediate attention. Not sure how to kick start this conversation :).

ljharb:
@kael-shipman i'm not sure about that; i'm not a maintainer here, but in my experience in open source in general, PRs aren't addressed in the order they come in. There could be any number of reasons why npm folks haven't had the time to respond to this one.

kael-shipman:
@ljharb  Oh, my bad, I saw you active in the other PRs and assumed you were one of the people in charge of leading the conversations. Sorry bout that! I'll continue to wait around.

ruyadorno:
Putting this back into the agenda since now that **npm7** is out, powered by `@npmcli/arborist` and with support to workspaces and more, we might want to revisit the idea and maybe come out of the discussion with some sort of action items and/or decision?

ruyadorno:
[Feedback](https://github.com/npm/rfcs/blob/c8cb6fe56ca8c7fc3dad5dd26f057aee472dc837/meetings/2020-10-21.md#pr-121-rfc-added-proposal-for-package-version-linkversioncomment-syntax---kael-shipman) from the last [OpenRFC call](https://www.youtube.com/watch?v=ub9n6hjM_es):
- Adding new syntax to `package.json` is something we rather avoid since it creates all sorts of incompatibility problems among different package managers and also even among legacy versions of npm
- `workspaces` are now supported in **npm7**! They are a possible very good solution to the problems you described in the **Motivation** section of the RFC, since they will symlink things into place (just like `npm link`) but also enforce version conformance.
- In case `workspaces` aren't enough, a possible better solution to the set of problems that might better align with the goal of ecosystem compatibility is to implement new features in the form of a subcommand rather than new `package.json` syntax.
- Maybe we could add a subcommand/workflow in which users could `npm eject <pkg>` and have it be automatically copied from the `node_modules` into `./<pkg>` and automatically added to your `package.json` `workspaces` so that we could better cater to the hacking dependencies from your working tree scenario (which sounds more like a completely diff RFC)

ruyadorno:
Thanks @kael-shipman for bringing this one up 😊 we'll probably continue the conversation next week in the next OpenRFC call.

kael-shipman:
I so appreciate you all following up on this! It's been very hard for me to get out from under my work to join these meetings, especially since they coincide with a standing meeting that I already have in place every other Wednesday. I'm used to people taking an attitude of "if you can't make the meetings, you don't get a say," and it's really refreshing to see you all making such an effort to keep me in the loop, so thank you for that :).

As for your comments it does sound like there's quite a bit of momentum behind workspaces as a solution to many of the problems I've presented. I haven't felt very eager about the concept in the past, but I recognize that this may just be the way the community has gone, and as part of the community I think it makes sense to lean into that decision for now and try to improve what's there, rather than continue to advocate for potentially conflicting and certainly at least overlapping functionality.

Given that, I'm fine with yall closing this PR if you're satisfied that it has been sufficiently discussed.

Thanks again for the consideration!

PR #126, opened 2020-04-07
RFC: Adding types information to the Package JSON in the  registry

orta:
# What / Why

Add an unambiguous notation to the [packument](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#package-endpoints) package.json whether a dependency includes types inside the tarball. 

It would be great to know from the NPM API if a library included type definitions. This would allow the npm website, and other tooling to be able to help people make decisions about which dependencies they would like to use and support for types plays into that.

## References

Prior work on the search index used by npm's site: https://github.com/algolia/npm-search/pull/346

### DIFF
--- a/accepted/0026-add-types-metadata-to-package-json.md
+++ b/accepted/0026-add-types-metadata-to-package-json.md
@@ -1,0 +1,102 @@
+# Add Types Metadata to the Registry [Packument](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#package-endpoints)
+
+## Summary
+
+Add an unambiguous notation to the derived package.json whether a dependency includes types.
+
+## Motivation
+
+It would be great to know from the NPM API if a library included type definitions. This would allow the npm website, and other tooling to be able to help people make decisions about which dependencies they would like to use and support for types plays into that.
+
+## Detailed Explanation
+
+Working with one of my libraries which ships with both Flow and TypeScript types:
+
+[Package.json](https://github.com/danger/danger-js/blob/e7d9b6a515eabe0488a5f0ec196319c7fbe479b0/package.json)
+
+```json5
+{
+  "name": "danger",
+  "version": "10.1.0",
+  "description": "Unit tests for Team Culture",
+  "main": "distribution/danger.js",
+  "typings": "distribution/danger.d.ts",
+  "bin": {
+    // 
+  }
+  "prettier": {
+    // ...
+  },
+  "scripts": {
+    // ...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/danger/danger-js.git"
+  },
+  "keywords": [
+    "danger",
+    "ci"
+  ],
+  "author": "Orta Therox",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/danger/danger-js/issues"
+  },
+  "homepage": "https://github.com/danger/danger-js#readme",
+  "devDependencies": {
+    // ...
+  },
+  "dependencies": {
+    // ...
+  },
+}
+```
+
+This `package.json` unambiguously declares that it has TypeScript types. However, a flow user has no idea that I put the time in [to maintain](https://unpkg.com/browse/danger@10.1.0/distribution/danger.js.flow) support for them!
+
+I propose that [the packument version of this package](https://registry.npmjs.org/danger/10.1.0) have a new field `"_types"` which would look like:
+
+```json5
+{
+  // ...
+  "dependencies": {
+    // ...
+  },
+  "gitHead": "66e18daae5286ea64ed49484fd5835685684c156",
+  "_id": "danger@10.0.0",
+  "_types": {
+    "ts": "included",
+    "flow": "included"
+  }
+}
+
+```
+
+The shape of this copies the work we did in [Algolia's npm search index](https://github.com/algolia/npm-search/pull/346). 
+
+This RFC would explicitly _exclude_ information on 3rd party types hosted on [DefinitelyTyped](https://github.com/DefinitelyTyped/) and [FlowTyped](https://github.com/flow-typed/flow-typed) - the assumption is that anyone wanting to provide comprehensive type information for _any_ package would check their registries if no type information is present in the packument.
+
+### "Unambiguous"
+
+Both flow and TypeScript types work on projects which don't declare that they have types inside them. I could remove `"typings"` in the above `package.json` and TypeScript would exhibit the same behavior. That would mean that the current package.json would not have any indication that there is typed support for either tools.
+
+### Downside of Recommendation
+
+1. This would mean that only newly shipped packages with the latest CLI would have that information. In theory, you could backport the logic and re-run it across the whole registry... That's quite a bit of work though.
+1. The naive recommendation of "check for *.d.ts / *.js.flow" could mean that projects which have types setup incorrectly could be triggered as a false positive. Personally, I think this is a tiny edge case and one likely a human would intervene on with fixing their types. The project could "have types", just not in the right places.
+
+## Rationale and Alternatives
+
+1.  Don't have that metadata on the site. Both TS/Flow represent unofficial extensions of JavaScript and who knows what the future holds.
+1.   Ping the algolia npm index for that information. I would expect you to want to keep your server side dependencies low, and including this information in the registry would allow you to keep one source of information.
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


accepted/0026-add-types-metadata-to-package-json.md
```diff
@@ -0,0 +1,102 @@
...
+```json5
+{
+  // ...
+  "dependencies": {
+    // ...
+  },
+  "gitHead": "66e18daae5286ea64ed49484fd5835685684c156",
+  "_id": "danger@10.0.0",
+  "_types": {
+    "ts": "included",
```
ljharb, line 69: why just "included"? It might be nice for a package to be able to point to a DT (or the flow equivalent) package for its types, so users don't have to guess, and if so, that info would presumably be wanted in the packument also?

sandersn: Presumably this could point to the `typings` field if present?

ljharb: right, but what i mean is, it's highly useful to be able to extract "package dep" information from the packument without having to extract the entire tarball; if "typings" is only pointing to relative files, a boolean is sufficient, but if it's pointing to a package, the package name and version would be important too.

orta: I'd like it, but it seems unlikely for many to add the additional field to declare that they have a DT/FT representation. I wouldn't be surprised that a lot of projects don't know they have DT/FT files.

That was why I said that I think it probably should be left to another tool to denote whether that types package exists. 

In theory (and just for DefinitelyTyped), because the types are on npm - it might be reasonable for the CLI to ping whether it's equivalent `@types/module` exists in the registry and mark it automatically

orta:
I can come to today's (or any) upcoming RFC meeting if you'd like me to answer questions 👍

ruyadorno:
@orta you're def welcome to join us! 😊 meeting is in less than half an hour though, you can find here the info on how to join: #134

orta:
I've given this RFC an update based on our discussions yesterday (and from @Haroenv's feedback ) (and I've pinged the flow team for feedback )

The main focus of the update: 

 - Switching out the boolean `"included"` to the resolved filepath - depending on how https://github.com/microsoft/TypeScript/issues/38249 comes out, that could also include a references to "@types/myapp".

 - Discussing the idea of making the "types" field more explicit in the package.json and using `npm package` to encourage, then enforce the adoption of it. This would require the flow team to have a similar field to `"types"` and would eventually break some builds when people are uploading, but it's a strong 2nd for me.

orta:
Hi folks - I'll come along to the next Open RFC Meeting, I'm not sure what to make of the notes from the last one 👍

orta:
OK, this RFC is updated based on the meeting today in https://github.com/npm/rfcs/issues/153 

We're good to add this field to the packment during the package building phase. No-one had strong opinions around the name of the field, so I kept it as `_typeRoots`.  RFC should be good to merge 👍

ljharb:
As commented in the call: it'd be great to have the paths be *required* to start with `./` for relative paths, disallow paths starting with `/` to block absolute paths, and leave open the design space for bare identifiers to be "a different package".

orta:
Agreed! - updated that to be explicit in the RFC ( and linking it to the TS issue on the topic https://github.com/microsoft/TypeScript/issues/38249 )

orta:
We chatted about this in the July 15th meeting and came to the conclusion that we can simplify this RFC.

The work can happen client-side during the same time period as the git metadata is added via [read-package-json](https://github.com/npm/read-package-json). The resulting package.json can be modified to always include "types" or "flow" instead of relying on a unique `_typesRoot` field.

I'm gonna update the RFC text

orta:
This has a PR: https://github.com/npm/read-package-json/pull/92

ThisIsMissEm:
Looking at this, wouldn't it be better to have something like:

```
types: {
  typescript: "./distribution/danger.d.ts"
  flow: "distribution/danger.flow"
}
```

but also support specifying a package where types can be found:

```
types: {
  typescript: { package: "@types/danger" }
  flow: "distribution/danger.flow"
}
```

Which would allow for tooling to automatically install the right third-party managed types.

orta:
Yeah, this came up a lot during the discussions and somewhat covered in the RFC:

> This RFC would always add the relative path to an existing file if the fields are empty.
>
> This allows for downstream consumers to be able to differentiate between a module hosting their types and a file available inside the package.

However, given there's already a very large corpus of TypeScript code with `types` today which will definitely include "non-relative" relative paths because `"main"` also supports it. I'm open to discussions for a new standard like `{ package: "xxyy" }`, but that's kinda outside the scope for this RFC which aims to add fill the fields in if they are inferred.

I opened https://github.com/microsoft/TypeScript/issues/38249 to talk about ways in which that could be described in TypeScript

orta:
This is now a feature in the npm site 🎉 

Screen Shot 2020-12-16 at 5 59 31 PM

Which was my original goal, that said, I think there's still value in not having everyone have to replicate the TypeScript / Flow default file resolvers in their apps (with a downloaded copy of the tarball) in order know if a dep has types embedded

wraithgar:
This is going out today in `v7.22.0`!!

santialbo:
This is probably not the place to post this. Please let me know if there's a more appropiate place to put this in.
I've noticed that if a package.json doesn't explicitly specify the "types" property this makes the package to not show the DT badge on npm.com
From the [TS docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)
> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the types property, though it is advisable to do so.
Package: https://www.npmjs.com/package/libphonenumber-js
Screenshot 2022-02-17 at 09 48 52
index.d.ts is included
Screenshot 2022-02-17 at 09 48 24

orta:
I believe modern builds of npm will automatically add 'types' to your package.json (during deployment) via https://github.com/npm/read-package-json/pull/103 - so you can either update your npm client to have it automatically added, or add the 'types' value yourself

ljharb:
I shouldn’t have to edit my package.json for that tho; npm should infer it regardless.

SimenB:
Is that logic impacted by `"exports": {"types": "typesPath", "default": "jsPath"}`? I.e. "sibling to main" also respects "sibling to `resolveExports('.')`?

orta:
> I shouldn’t have to edit my package.json for that tho; npm should infer it regardless.

I am _assuming_ that the npm search index for the site isn't reaching into the tarball fs and is using the packument as the source of truth for the icon (which would be replicating the work in https://github.com/npm/read-package-json/pull/103 ). I could be wrong though. You wouldn't have to update the package json if the deploying npm version included https://github.com/npm/read-package-json/pull/103 because it gets baked into the richer package.json which is uploaded

> Is that logic impacted ... exports

No idea TBH, but probably not, I think the feature in the npm site was added before esm support was unprefixed in node.

ljharb:
@orta i have the TS logo displayed on a ton of my packages, but i guess that's because they have DT packages instead of shipping types?

orta:
Because I don't have access to the npm site source code all of this discussion is just my guesswork, but I'd imagine some of those packages have a `.d.ts` file in the right place and it's being picked up by the read-package-json change from v7.22.0+ - I could be wrong (checking the packument file would confirm)

For example, express which doesn't ship it's types in any way still shows up with a `[ DT]` logo instead:
Screen Shot 2022-02-17 at 9 03 25 PM

While I'm thinking about this, both on and off-topic to this current conversation, this RFC can now either be closed or merged because the changes have been fully integrated into npm too.

rauschma:
My impression: If the types sit next to `.js` files but aren’t mentioned in package.json then there is no badge. TypeScript is perfectly fine with that, though.

Two packages that IMO should have badges but don’t:

* https://www.npmjs.com/package/asserttt
* https://www.npmjs.com/package/@rauschma/helpers

ljharb:
https://www.npmjs.com/package/is-weakref doesn’t have the types in package.json but does have the badge.Are you running https://npmjs.com/@arethetypeswrong/cli on these packages?

rauschma:
Interesting! I checked:
* Neither has `"main"` (key difference with your package).
* arethetypeswrong is mostly OK with that. It only complains for node10 and the simpler one of the two packages:
  ```jsonc
  // OK for node10 and node16
  "exports": {
    "./*": "./dist/src/*"
  },
  // Only OK for node16
  "exports": {
    ".": "./dist/src/asserttt.js"
  },
  ```

ljharb:
fwiw I suspect you can avoid that last warning by removing type module, and making your ESM files be `.mjs`, but that doesn't seem like it should disrupt the TS badge.

PR #129, opened 2020-04-10
RFC: overrides

isaacs:
cc: @ljharb @wesleytodd

[formatted RFC](https://github.com/npm/rfcs/blob/isaacs/overrides/accepted/0000-overrides.md)


accepted/0000-overrides.md
```diff
@@ -0,0 +1,492 @@
...
+    +-- b
+    +-- c (1.2.3)
+```
+
+Only nodes `a`, `b`, and `c` under `x` are duplicated, because they are
+subject to a relevant `overrides` setting.
+
+## Questions and Bikeshedding
+
+- Should `bundleDependencies` and `shrinkwrap` dependencies be subject to
```
ljharb, line 488: I would not expect bundleDeps to be affected, since the whole point of them is "npm is not allowed to touch bundled deps". Not sure what you mean about "shrinkwrap deps"?

isaacs: Dependencies of a module that includes a `npm-shrinkwrap.json` file.  Again, the whole point of them is "npm is not allowed to use different deps than the tree specified in the shrinkwrap", so in most ways, they're treated just like bundled deps (albeit, bundled deps that we still have to fetch and unpack).

ljharb: It seems to me like that shouldn't be touched either.

ruyadorno:
I wanted to see an example for overriding a dep version and **also** overriding versions for some of its sub deps (I can see it being a requirement to many plugin-based ecosystems) e.g:

```json
{
  "overrides": {
    "y": "2.x",
    "y@2.x": {
      "x@1.x": "1.2.3"
    }
  }
}
```

isaacs:
> I wanted to see an example for overriding a dep version and **also** overriding versions for some of its sub deps (I can see it being a requirement to many plugin-based ecosystems) e.g:
> 
> ```json
> {
>   "overrides": {
>     "y": "2.x",
>     "y@2.x": {
>       "x@1.x": "1.2.3"
>     }
>   }
> }
> ```

Oh, that's interesting, and might run into issues with the "no re-evaluating" rule suggested in the RFC.

Let's say you have a `y@1` dep, which gets rewritten to `y@2.x`.  Then, because that was overridden already, it _won't_ get the overrides attached to it.  So, you could have `y@1` get overridden to `y@2`, but then get `x@1` getting version `1.9.9` instead of `1.2.3`.

Maybe it should get _object_ overrides attached to it, but _not_ string overrides, so we don't have infinite loops?

Eg, this would not work:

```js
{
  "overrides": {
    "y@1": "2",
    "y@2": "2.3.4",
    "y@2.3.4": "1.2.3" // uh oh
  }
}
```

A `y@1` dep gets rewritten to `y@2` which resolves to `y@2.9.9`, which matches `y@2` and gets rewritten to `y@2.3.4`, which matches `y@2.3.4` and gets rewritten to `y@1.2.3` which matches `y@1` and now we're in trouble.

But in this case:

```json
{
  "overrides": {
    "y@1": "2",
    "y@2": {
      "x@1": "1.2.3"
    }
}
```

The `y@1.x` dep gets rewritten to `y@2` which resolves to `y@2.3.4`.  This matches `y@2` which is an object value, so it gets attached to the node.  Then its dependents matching `x@1` get rewritten to `x@1.2.3`.

If we block string overriding to a single move, then this wouldn't work:

```json
{
  "overrides": {
    "y@1": "2",
    "y@2": "2.3.4",
    "y@2.3.4": {
      "x@1": "1.2.3"
    }
}
```

The `y@1` dep gets rewritten to `y@2`, which resolves to `y@2.9.9`.  This matches `y@2`, but the value there is a string, and it's already been overridden so we give up.  `y@2.9.9` then has no overrides attached, and gets `x@1.9.9`, which is not what the user expects.  They'd have to do this instead:

```json
{
  "overrides": {
    "y@1": "2.3.4",
    "y@2": "2.3.4",
    "y@2.3.4": {
      "x@1": "1.2.3"
    }
}
```

We _could_ perhaps handle the infinite loop string-value scenario by tracking which override rules were applied, and only applying each _rule_ a single time.  So in the loop case:

```js
{
  "overrides": {
    "y@1": "2",
    "y@2": "2.3.4",
    "y@2.3.4": "1.2.3" // uh oh
  }
}
```

The `y@1` dep gets rewritten to `2`, which resolves to `2.9.9`.  This matches the `y@2` rule, and gets rewritten to `2.3.4`.  This matches the `y@2.3.4` rule, and gets rewritten to `1.2.3`.  The `y@1` rule is not considered, because it was already applied to this dep, so it ends up as `y@1.2.3`.

If you started with `y@2.3.4`, it'd get rewritten to `1.2.3`, match `y@1` rewritten to `y@2` resolved to `2.9.9`, match `y@2` rewritten to `y@2.3.4`, and we're done.  (`y@2.3.4` -> `y@2.3.4`, not exactly "overridden".)

We might be able to do something clever to resolve these rules, detecting that the range `2` is a subset of `2`, then `2.3.4` is a subset of `2.3.4`, then `1.2.3` is a subset of `1`, so we have a loop.  But if the user is putting a tarball or git url on the right hand side, that doesn't work.

wesleytodd:
Overall I think this is looking good to me.  The only thing which I dont like much (although I don't consider this a blocker), is the case here: https://github.com/npm/rfcs/pull/129/files#diff-8a9b6b0e44ed001c48613a8c6d924cf9R309-R324

```json
{
  "y": "1.2.3",
  "y@1": {
    "x": "1.2.3"
  }
}
```

I am not sure I like the "just use any other specifier key to avoid the fact that you mean the same thing".  What if instead it was something like this?

```json
{
  "y": {
    ".": "1.2.3",
    "x": "1.2.3"
  }
}
```

Some signifier which references the 'parent specifier' would better express the intent of the user IMO.  Going to the edge cases it could even allow for something like this:

```json
{
  "y": {
    ".": "1.2.3",
    "x": "1.2.3"

    ".@2": {
      ".": "2.0.1",
      "x": "2.0.0"
    }
  }
}
```

I know this is pushing back to the complex area, and maybe the nesting objects is out of scope, but for the simple case I feel like saying "I want `1.2.3` of `y`, and also `1.2.3` of `x` under `y`" is more clear with this than by requiring a semver range just to ensure it is not a duplicate key.

isaacs:
Yeah, I kind of like `.` as an option to avoid the double-match.  (It also lets us go back to saying "first match wins, no combining", which simplifies the logic quite a bit.)

I very much do _not_ like `.@<specifier>` to further limit it.  I think if you want separate objects for `y@1` and `y@2`, you just create two objects.  So, that'd change your third example to:

```json
{
  "y@1": {
    ".": "1.2.3",
    "x": "1.2.3"
  },
  "y@2": {
    ".": "2.0.1",
    "x": "2.0.0"
  }
}
```

I'd go another step further on the `"."` point and say that `"."` may _only_ be a string value, not an object.

I think that's a little bit less confusing to read, and avoids the following nightmare:

```js
{
  "y": {
    ".@1": {
      ".@1.2": {
        ".@1.2.5": {
          ".": "1.2.4"
          "x": {
            ".@2": "2.0.1",
            ".": "2.0.2"
          }
        },
        ".@1.2.3": "1.2.4"
      },
      "thenight": {
        ".": {
          "is": "dark",
          ".": "1.2.3"
        },
        ".@and": {
           ".": "full",
           "of": {
             ".": "terrors"
           }
         }
      },
      ".": "1.2.7"
    }
  }
}
// quick, without cheating, what gets rewritten to 1.2.7?
```

isaacs:
Or, I guess the intent of that third example would be more like:

```json
{
  "y@2": {
    ".": "2.0.1",
    "x": "2.0.0"
  },
  "y": {
    ".": "1.2.3",
    "x": "1.2.3"
  }
}
```

So that all `y` other than `y@2` get the second block applied.

isaacs:
Then we could also say that `"<selector spec>": "<override spec>"` is actually a shorthand for `"<selector spec>": { ".": "<override spec>" }` which has some nice implications for the implementation.  We just always use the first object that matches (merged with all parents), and instead of string values having different treatment, the `"."` _key_ has different treatment.

wesleytodd:
> I very much do not like .@<specifier> to further limit it. I think if you want separate objects for y@1 and y@2, you just create two objects.

> I'd go another step further on the "." point and say that "." may only be a string value, not an object.

Sounds good  to me!

ruyadorno:
following up from the discussion at: https://github.com/npm/rfcs/discussions/78#discussioncomment-16512

@isaacs do you think in its current state this RFC enables that usecase?

zkochan:
The `"."` syntax seems very unusual to me. What about using an array instead?

```json
{
  "y@1": [
    "1.2.3",
    {
      "x": "1.2.3"
    }
  ],
  "y@2": [
    "2.0.1",
    {
      "x": "2.0.0"
    }
  ]
}
```

accepted/0000-overrides.md
```diff
@@ -0,0 +1,647 @@
...
+The resulting package tree on disk will have `foo` loading `bar@2.0.0` when
+it does `require('bar')`.  Despite this, it will _not_ be treated as an
+invalid dependency, because `bar@1.x` was valid prior to the override being
+applied.
+
+### String Overrides
+
+When the value of an override rule is a string, it is a replacement
+resolution target for resolutions matching the key.  String values _must_
+be a dependency specifier without a name.  (Aliases are supported using the
```
zkochan, line 104: It is not clear to me how are aliases treated.

For this case for instance:

```js
{
  "overrides": {
    "y@1": "1.2.3",
    "y@1.2.3": "2.3.4" // <-- this will never match anything
  }
}
```

What will this override?

Will it override this: `"y": "npm:x@1.0.0"`
or this `"x": "npm:y@1.0.0"`

accepted/0000-overrides.md
```diff
@@ -0,0 +1,647 @@
...
+}
+```
+
+### Object Overrides
+
+An object value in an overrides object defines a _child rule set_.
+
+If the first match for a given resolution is an object, then the object is
+a new rule set applied to all resolutions down that path in the dependency
+graph.
```
zkochan, line 194: > all resolutions down that path in the dependency graph

so in the example below, `bar` may be not a direct dependency of `foo`?

accepted/0000-overrides.md
```diff
@@ -0,0 +1,647 @@
...
+    ".": "1.0.0",
+    "bar": {
+      ".": "2.3.4",
+      "baz": "3.0.0"
+    },
+    "baz": "2.0.0",
+    "boo": "3.0.0"
+  },
+  "boo": "1.0.0"
+}
```
zkochan, line 290: Yarn also allows this complex overrides but I can't imagine a real world example when you'd want this.

Also, I don't think we could support something like the in pnpm.

We could support only overrides when the direct dependent is specified.

accepted/0000-overrides.md
isaacs: Neither of these would be overridden by this rule (at least, as the RFC is written here), because neither would match `y@1`.

It is interesting, though, and perhaps worth spelling out explicitly in the RFC, how aliases would work.

At least as I read it, if you had an override like this:

```
{
  "y@npm:x@1": "npm:x@1.2.3"
}
```

Then an installation of `y@npm:x@1.9.9` would be overridden to `y@npm:x@1.2.3`, because `y@npm:x@1.9.9` _is_ a match to `y@npm:x@1`.

accepted/0000-overrides.md
isaacs: Correct.  It would be any instance of `bar` at any point in the dependency graph traversal from any `foo` node.

isaacs:
@ruyadorno 

> following up from the discussion at: https://github.com/npm/rfcs/discussions/78#discussioncomment-16512
> 
> @isaacs do you think in its current state the proposal from this RFC enables that specific usecase?

No, it doesn't suggest any way to say "do not resolve this dependency".  I'm not opposed to that in principle, but it does complicate things somewhat.  (Eg, if the dependency is not optional, then it won't be installable, and it might not be obvious which version of a given thing _would_ be installable, if for example the blocked dep was added as a meta-dep in a minor update.)

@zkochan 

> The "." syntax seems very unusual to me. What about using an array instead?

It's a bikeshed, and I can see that it's functionally equivalent.  However, it is somewhat nice to be able to just check if it's an object or string, and not have to also check for it being an Array.  Since the Array could only reasonably contain 2 members, it feels like a violation of the zero-one-infinity rule.  (Eg, what happens if you do `{"foo": [{"bar":"1.2.3"},{"bar":"2.3.4"},"3.0.0","4.0.0"]}`?  How would we interpret that intent?)

Is there a key name that you can suggest to indicate "the current point in the traversal" that would be less unusual to you than `"."`, and also not be a valid package specifier?

I had thought maybe `"@"` would be somewhat intuitive, since it's kind of saying `<name>@<version>`.

```json
{
  "overrides": {
    "foo": {
      "@": "2.3.4",
      "bar": "1.2.3"
    }
  }
}
```

accepted/0000-overrides.md
isaacs: Ah, I see, because you might have two instances of the same version of a package subject to different overrides?  Eg:

```
root -> (a, b)
a -> (x@1.2.3)
b -> (x@1.2.3)
x -> (y)

overrides: {
  a: {
    x: {
      y: "1.2.3"
    }
  },
  b: {
    x: {
      y: "2.3.4"
    }
  }
}
```

So, the `x@1.2.3` depended upon by `a` would need a `y` of `1.2.3`, but the `x@1.2.3` depended upon by `b` would need `y@2.3.4`.

Meaning both `root -> a -> x` and `root -> b -> x` could not be symbolic links to the same physical `x@1.2.3` folder.  The realpath target would have to be unique to both the package resolution and its set of overrides.

Am I understanding that correctly?  If so, it seems like overrides is a challenge for pnpm even without applying to the full dependency subgraph.  If not (and making the real paths unique to both resolution and override set is doable) then I'm not sure how this override is any different:

```
{
  a: { y: "1.2.3" },
  b: { y: "2.3.4" }
}
```

zkochan: > If so, it seems like overrides is a challenge for pnpm even without applying to the full dependency subgraph. 

Overrides are doable with pnpm if a particular version of a package gets the same set of dependencies. I don't see issues with your last example. Pnpm would create a directory structure like this:

```
.pnpm
+ a@1.0.0
   + node_modules/y -> ../../y@1.2.3
+ b@1.0.0
   + node_modules/y -> ../../y@2.3.4
+ y@2.3.4
+ y@1.2.3
```

But in your first example, we would need to create separate directories for the same x. Something like

```
.pnpm
+ x@1.2.3_<hash1>
   + node_modules/y -> ../../y@1.2.3
+ x@1.2.3_<hash2>
   + node_modules/y -> ../../y@2.3.4
+ y@2.3.4
+ y@1.2.3
```

Note: we actually do this for peer dependencies... but I am not sure it worse adding this for overrides. At the moment, I don't believe overrides need to be so powerful as in this RFC (or as in Yarn).

zkochan:
Hm, what if we make it a bit more powerful. We can add an additional layer of nesting to tell the type of the overridden dependency. It would also solve the dot problem:

```json
{
  "overrides": {
    "foo": {
      "version": "2.3.4",
      "dependencies": {
        "bar": "1.2.3"
       },
       "optionalDependencies": {
         "qar@2": "3"
       }
    }
  }
}
```

isaacs:
> Hm, what if we make it a bit more powerful? We can add an additional layer of nesting to tell the type of the overridden dependency.

My objection to this approach is that it gets extremely verbose very quickly, and since we have a very subtly different "type" of thing in the key and value of the various dependency sets, it's a bit confusing.  That is, you'd have an object named `dependencies` which is not a typical "dependencies" object, because the key is a named specifier instead of a package name (where a package name is a valid named specifier), and a version string or object on the right hand side instead of a version range string (where a version string is a valid version range string).

Eg, instead of this:

```json
{
  "overrides": {
    "a": {
      ".": "1.0.0",
      "x": {
        "y": "1.2.3"
      }
    },
    "b": {
      ".": "2.0.0",
      "x": {
        "y": "2.3.4"
      }
    }
  }
}
```

You'd have something like this:

```json
{
  "overrides": {
    "dependencies": {
      "a": {
        "version": "1.0.0",
        "dependencies": {
          "x": {
            "dependencies": {
              "y": "1.2.3"
            }
          }
        }
      },
      "b": {
        "version": "2.0.0",
        "dependencies": {
          "x": {
            "dependencies": {
              "y": {
                "version": "2.3.4"
              }
            }
          }
        }
      }
    }
  }
}
```

It'd also seem to suggest that other fields can be overridden (bin, scripts, etc.) which bloats the scope of this feature considerably.

accepted/0000-overrides.md
isaacs: It seems like overrides would be doable in pnpm if a unique combination of (1) resolved version and (2) override set gets the same set of dependencies (which it would).

If you don't feel it's worthwhile to implement the feature in that way, that's really a trade-off between UX, power, and complexity which is really your choice to make.  However, the fact that it *can* be implemented by pnpm makes me feel like this should not be an obstacle to accepting this RFC and npm implementing it with these semantics.

isaacs:
Discussion from last OpenRFC call:

- Not going to do the `{"x":{"optionalDependencies":{"y":"1.2.3"}}` approach.  Having explored the edge cases, deemed wayy too many foot guns and implementation complexity involved.
- Sticking with `"."` as the "this package name's override" option.  Less cute than `"@"`, but symmetrical semantics with import maps and node's `"exports"` field, so makes sense to use the same key that those things do.

No implementation work has been done on this (unless you count scribbles in a notebook), and it's not going to be included in npm 7.0.0, but will be added in the 7.x line as a semver-minor feature update.

dominykas:
Just curious if anyone has seen a userland implementation of this (aside from Yarn's equivalent feature)? I figure it would be quite handy to try it out and that could maybe feed back into the spec?

gajus:
> Just curious if anyone has seen a userland implementation of this (aside from Yarn's equivalent feature)? I figure it would be quite handy to try it out and that could maybe feed back into the spec?

Not the spec implementation, but if anyone needs a workaround, I've documented a solution compatible with NPM v7.

https://stackoverflow.com/a/65014694/368691

manoj317:
Json To String
Json means- JSON - JavaScript Object Notation. JSON - for storing and exchanging data. 

Try This online [json to string](https://www.jsontostring.com/) online converter.
For Example:
{
  "browsers": {
    "firefox": {
      "name": "Firefox",
      "pref_url": "about:config",
      "releases": {
        "1": {
          "release_date": "2004-11-09",
          "status": "retired",
          "engine": "Gecko",
          "engine_version": "1.7"
        }
      }
    }
  }
}
To string:
"{\"browsers\":{\"firefox\":{\"name\":\"Firefox\",\"pref_url\":\"about:config\",\"releases\":{\"1\":{\"release_date\":\"2004-11-09\",\"status\":\"retired\",\"engine\":\"Gecko\",\"engine_version\":\"1.7\"}}}}}"

Artur-:
Is there any way to follow the progress of the implementation of this? https://github.com/npm/rfcs/pull/129#issuecomment-658906056 only mentions that it "will be added in the 7.x line"

bnb:
@Artur- there was discussion about this in today's RFC. There was an assertion that the CLI team's focus has been on workspaces, but there are plans to work on it soon (my assumption is that "soon" is "as workspaces work is finished up").

isaacs:
Moved to accepted.

amber-pli:
When will this be released? Which npm version will have this feature?

csvan:
Is there a way to track the status of this being implemented in NPM? It's currently a blocker for us migrating our codebase completely from Yarn and we would love to see it :)

PR #133, opened 2020-04-29
RFC: Remove --depth from npm outdated

claudiahdz:
cc. @isaacs

### DIFF
--- a/accepted/0027-remove-depth-outdated.md
+++ b/accepted/0027-remove-depth-outdated.md
@@ -1,0 +1,115 @@
+# Remove --depth from npm outdated
+
+## Summary
+
+Remove `--depth` in favor of a new option flag named `--all` that shows _all_ outdated packages in the tree.
+
+## Motivation ("The Why")
+Currently on npm v6 `npm outdated` has an optional depth flag to configure the maximum depth for checking outdated packages on the dependency tree. As it is right now, it takes an _integer_ value that represents the amount of nested levels deep it goes.
+
+In practice, what most users care about is `--depth 9999` which should display a list of _all_ outdated dependencies on the tree. Also, npm v7 has already recently removed the `--depth` flag from `npm update` in favor of updating all nodes in the tree at every depth, which defeats the purpose of having the functionality to inspect outdated packages at specific depths if they are all going to be updated either way.
+
+Removing `--depth` and adding a flag named `--all` aligns with the future plans for v7. Considering npm flags are global and apply to all commands, `--all` could be used not only on `npm outdated` but also in other commands such as `npm ls` in the context of "show all dependencies".
+
+## Detailed Explanation
+Going any level deep further than the top-level deps opens a complex scenario where a single dependency might need different versions to satisfy requirements from different packages depending on it. Basically, what this means is that the "wanted" version of a nested dependency could be either a  _single_ semver range or a _set_ of semver ranges. 
+
+When `--depth` is used on v6, the same dependency and its "wanted" value will be displayed and repeated on console with their location as the sole way of differentiating its level of nesting in the logical tree. This current output can tend to clog the console since the command doesn't tell appart deduped dependencies and we end up with a lot of noise that could be summarized differently.
+
+Ex.
+```
+➜  arborist git:(master) npm outdated --depth 9999 tar
+Package  Current  Wanted  Latest  Location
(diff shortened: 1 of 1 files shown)


accepted/0027-remove-depth-outdated.md
```diff
@@ -0,0 +1,89 @@
...
+│   └── tar@4.4.13
+└─┬ pacote@11.1.0
+  ├─┬ cacache@15.0.0
+  │ └── tar@6.0.1  deduped
+  └── tar@6.0.1
+```
+
+In this example, we only care about two versions of `tar`, `4.4.13` and `6.0.2`. Instead we are shown 4 lines of output.
+
+Moreover, npm v6 currently has inconsistent behavior. Doing `npm outdated --depth 9999` displays all outdated packages that are direct children of the root package, missing out dependencies that are nested at other levels. This gives a wrong impression since we are not really displaying _all_ outdated dependencies of the tree. However, doing `npm outdated foo --depth 9999`, will indeed display all appearances of `foo` on the tree no matter if they are direct children of the root package or not. This is precisely the functionality that displaying all deps should have too (`npm outdated` with no flags).
```
isaacs, line 41: > Doing `npm outdated --depth 9999` displays all outdated packages that are direct children of the root package, missing out dependencies that are nested at other levels. This gives a wrong impression since we are not really displaying _all_ outdated dependencies of the tree.

I don't think that's the case, unless I'm misreading this maybe?  `npm outdated --depth=9999` includes this output, for example:

```
log-update                  3.4.0    3.4.0    4.0.0  npm-audit-report > tap > ink
log-update                  3.4.0    3.4.0    4.0.0  npm-audit-report > tap > treport > ink
```

claudiahdz: I think in that case that particular dependency is being deduped and is _indeed_ a direct child of your root package, dependencies that are not deduped and are nested within their own dependant, do not seem to show up when doing only `npm oudated --depth 9999` but DO show up when you do `npm oudated --depth 9999 <packageName>`.

wesleytodd:
The addition of `--all`, as discussed on the RFC call, means a new config option `all`, which presents a hazard because these are global and would apply to more than just this command.  This also came up with the workspace filter flag.  We just need to be clear what the goals are for config/flags.

accepted/0027-remove-depth-outdated.md
isaacs: Hm, it's not though.  `log-update` is a dependency of ink, and my package depends on tap, which depends on ink.

```
$ find node_modules -name log-update
node_modules/tap/node_modules/log-update
```

It is a good example of the fact that `--depth` is pretty weird when talking about a dependency graph, though.  It looks like I have two outdated versions of `log-update`, when really they're the same dep node.

claudiahdz: Yeah, correct. The thing is that `log-update` is a direct child of `tap`.  So I misphrased that by saying the command shows only direct children of the _root_ package when I meant to say that the v6 algorithm is iterating recursively over the node's direct children (on the file system) instead of over `node.edgesOut`. 

Try doing `npm outdated --depth 9999 ansi-escapes` (basically any specific dep inside `node_modules/tap/log-update/node_modules`) and you'll see they are not displayed when calling outdated without a package-name.

Using `tree.inventory` and `node.edgesIn` on v7 will definitely solve this problem and properly give us a complete list of all outdated packages.

isaacs:
Update from RFC call: `--long` is not a suitable replacement for `--all`, because it's already used by `npm outdated` to provide additional information.

Worth considering if we could make `npm update` continue to only update root level deps, and then have the (current, as of this moment) `npm update` behavior be triggered by `npm update --all`.  Would definitely justify having the flag on outdated, and provide some nice symmetry, though I think you're more likely to want to _update_ everything that see _outdated versions_ of everything.  The fact that you don't have to care about nested deps means "just update it" is fine, and "show me what needs updating" is unnecessary noise most of the time.

ljharb:
Definitely the ideal default for `update` is "everything", and for `outdated` is "my direct deps only".

ruyadorno:
are there any breaking changes expected for the `--json` output?

claudiahdz:
@ruyadorno No, the only thing to note is that the location info has changed and is now `node.location` (its position on the physical tree) instead of its position on the logical dependency tree

wesleytodd:
Just to add here what I bought up on the call, could we display both the logical and physical path?

PR #135, opened 2020-05-01
Clarify/Outline the RFC Withdrawal Process & Amendment

darcyclarke:
We need some clarification around how we opt-out/rescind/withdraw previously *accepted* RFCs that may become outdated or are succeeded before ever being *implemented*.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
-# npm RFC Process
+# RFC Process
 
 This document describes the RFC process for the [npm
-CLI](https://github.com/npm/cli), and provides a way for the [Community & Open Source Team]() and the
+CLI](https://github.com/npm/cli), and provides a way for the [Community & Open Source Team](https://github.com/orgs/npm/teams/cli-team) and the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,11 +78,22 @@ integration into the CLI may also be deferred to a later date, or a later
...
-the PR. Feel free to do this even for work-in-progress code!
+the PR. Feel free to do this even for work-in-progress code.
+
+### Withdrawal
+
+From time to time npm collaborators will review RFCs awaiting implementation to
+ensure their accuracy and relevance. In cases where a previously ratified RFC is
+deemed to no longer be a viable candidate for implementation, an [**amendment
+section**](withdrawn/0000-template.md) will be added to the bottom of the
+document outlining the reason for repeal and subsequently moved to the
```
isaacs, line 95: My only nit is that the amendment section should be at the top, not the bottom, so that it's harder to miss.

darcyclarke: 👍🏻 *Updated*

PR #144, opened 2020-05-19
RFC: Add npm diff

ruyadorno:
### Quick overview:

Demo showing npm diff pkg in a repo with an actual tree installed

[See more in the rendered RFC](https://github.com/npm/rfcs/blob/c14a4cb8a44bd065fbfed924ce8948c17e8a88f8/accepted/0000-npm-diff.md)

### DIFF
--- a/accepted/0000-npm-diff.md
+++ b/accepted/0000-npm-diff.md
@@ -1,0 +1,109 @@
+# npm diff
+
+## Summary
+
+Add a new command that enables diff workflows (similar to `git diff`) for
+packages published in the registry.
+
+## Motivation
+
+- Package Consumers: Complements `npm audit` and `npm outdated` workflows by
+providing insight on what changed across different versions of a package.
+- Package Authors: Enables diff packlist-tracked-only file changes prior to
+publishing a new version of a package, or while debugging past changes.
+
+## Detailed Explanation
+
+Introduce a new `npm diff` command that accepts one or more package specs and
+uses these specs to compare files and print diff patches. Contents are fetched
+using `pacote`, accepting any type of specs that the npm installer understands.
+
+### npm diff --diff=`<spec>`
+
+Using a single `<spec>` allows users to retrieve diff output between an
+existing valid version of a package with the same **name** found in the actual
+dependency tree and the exact `<spec>` match from the registry.
+
+### npm diff --diff=`<spec-a>` --diff=`<spec-b>`
+
+Fetches/reads contents of two versions of a package, represented by
+`<spec-a>` and `<spec-b>` and prints diff output.
+
+### npm diff (no arguments)
+
+Meant as a helper tool to package authors, prints diff output between the
+current file system dirs/files (tracked by packlist) and the last published
+version of that package.
+
+## Rationale
+
+- A very common workflow is using npm outdated to figure out what dependencies
+need update and then manually reviewing what changed in between the current
+version you have in your project and whatever you can update to. That workflow
+involves multiple steps, some of them being extra mental hurdle (such as
+keeping track of the semver versions change and their meaning) some other very
+manual such as jumping around to repos, scanning through changelogs, veryfing
+semver contract is respected, etc.
+- It feels very natural to have a `npm diff` from the point of view of a
+user of `git diff`s - It does provide much more quick insight and transparency
+on code that makes its way into our projects in a
+([medium](https://en.wikipedia.org/wiki/Diff#Unified_format)) that most
+developers are already familiar with.
+
+## Alternatives
+
+- Not have a native `npm diff` command and deffer to userland tools.
+
+## Implementation
+
+Parses user args and reads package contents for two different specs and run a
+diff of each file, providing an usable patch output for users to work with.
+
+### Parsing diff arguments logic
+
+What should happen if using `npm diff` (no args) ?
+What should be the expected behavior when using `npm diff --diff=<pkg>` ?
+What if I use two `--diff=<pkg>@<version>` arguments?
+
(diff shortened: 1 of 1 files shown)


ljharb:
Interesting - this is diffing between the file contents, but i'd also want to know the theoretical lockfile diff - iow, what transitive dep name/version changes have happened across that span.

accepted/0000-npm-diff.md
```diff
@@ -0,0 +1,60 @@
...
+
+<img alt="Demo showing npm diff pkg in a repo with an actual tree installed" src="https://ruyadorno.github.io/svg-demos/npm-diff/npm-diff-yargs.svg" />
+
+### npm diff `<spec-a>` `<spec-b>`
+
+Fetches tarball contents of two versions of a package, represented by `<spec-a>` and `<spec-b>` and prints diff output.
+
+<img alt="Demo showing npm diff pkg in a repo with an actual tree installed" src="https://ruyadorno.github.io/svg-demos/npm-diff/npm-diff-simple-output.svg" />
+
+### npm diff --changelog
```
bnb, line 28: It would be wonderful to see the ability to pick a single file from the `npm diff` output. Perhaps this could just be a shortcut built on top of that?

This also leads me down the path of wondering if it would be possible to include CHANGELOG files in the npmjs.com UI.

ljharb:
I also asked during the RFC call about whether the diff should be from "current to wanted" by default rather than "current to latest", so as to avoid breaking changes (matching what `npm update` does by default)

accepted/0000-npm-diff.md
ruyadorno: oh yeah, a proper changelog section in the website UI would be neat, specially one that does fallback to looking up GitHub releases 🤩 

now back to your original point, I think it makes a lot of sense to have a file-filter argument, as I commented during the call it would enable us to do neat things in companion of the `--name-only` flag, workflows such as: `npm diff --name-only | grep index | xargs git diff` or have an interactive workflow that allows you to select from the list of files: `npm diff --name-only | npx ipt | xargs git diff`

ruyadorno:
btw we actually have a working prototype over at https://github.com/npm/cli/pull/1319

ruyadorno:
@coreyfarrell suggested in the tooling WG that it would be nice to have a `pager` config (similar to git) to go with that, to which I really like 😊

coreyfarrell:
@ruyadorno have you considered if any standard diff CLI flags will be supported?  Like `pager` support this is not critical but I very frequently use `git diff --ignore-all-space`.  This is great when blocks of unconditional code gets moved into a condition:
```diff
diff --git a/lib/internal/main/worker_thread.js b/lib/internal/main/worker_thread.js
index 6e3935d138..e6ae3952f3 100644
--- a/lib/internal/main/worker_thread.js
+++ b/lib/internal/main/worker_thread.js
@@ -113,8 +113,9 @@ port.on('message', (message) => {
     initializeDeprecations();
     initializeWASI();
     initializeCJSLoader();
-    initializeESMLoader();
+    initializeESMLoader(loaderPort);
 
+    if (!internal) {
       const CJSLoader = require('internal/modules/cjs/loader');
       assert(!CJSLoader.hasLoadedAnyUserCJSModule);
       loadPreloadModules();
@@ -122,6 +123,7 @@ port.on('message', (message) => {
       if (argv !== undefined) {
         process.argv = process.argv.concat(argv);
       }
+    }
     publicWorker.parentPort = publicPort;
     publicWorker.workerData = workerData;
 
```

With or without support for standard `diff` flags this feature will be great!

ljharb:
In general it’d be nice if it could defer to git diff - and all its options.

juliangruber:
Quickly sharing my learnings from working on https://github.com/juliangruber/npm-diff:

- `git diff` plus `colordiff` works great, however `colordiff` isn't available everywhere. A JS replacement might not be hard to write
- If `less` is used with `colordiff`, need `less -R`
- In my tests it was useful to ignore any changed files with `test/` and `Makefile` in their name, since as a consumer these don't change my situation. There could be false negatives ignoring `test`.
- In my tests it was also useful to ignore line changes containing `readme`, `_id`, `_from` and `_resolved`, as these were noise

accepted/0000-npm-diff.md
```diff
@@ -18,13 +18,13 @@ Introduce a new `npm diff` command that accepts one or more package specs and
 uses these specs to compare files and print diff patches. Contents are fetched
 using `pacote`, accepting any type of specs that the npm installer understands.
 
-### npm diff `<spec>`
+### npm diff --diff=`<spec>`
```
ljharb, line 21: the double "diff" reads pretty awkwardly, but i don't have a better bikeshed color to suggest.

styfle: How about arguments like this?

```
npm diff `<spec-a>` `<spec-b>`
```

ljharb: that’s just positional arguments (there’s no backticks convention in cli args), which I’d prefer, but I’m led to believe there’s a technical barrier to that.

ruyadorno: yes, we started there but we switched to named arguments in order to avoid the ambiguity between specs and file paths to filter

bnb:
Should this be merged?

ruyadorno:
yes! I'll just finish cleaning it up and merge it as implemented 😊

ruyadorno:
For future reference, this feature went out in `npm@7.5.0` and I wrote a little overview in a blog post format over here: https://dev.to/ruyadorno/npm-diff-23dh

PR #146, opened 2020-05-20
RFC: Notification system for cli updates

ruyadorno:
[See RFC](https://github.com/ruyadorno/rfcs/blob/1895425d6ed18765cd683db3aab5620bf149c3f4/accepted/0000-notification-system-for-cli-updates.md)

### DIFF
--- a/accepted/0000-notification-system-for-cli-updates.md
+++ b/accepted/0000-notification-system-for-cli-updates.md
@@ -1,0 +1,31 @@
+# Notification system for cli updates
+
+## Summary
+
+A new notification system to warn users about new versions of the **npm cli**.
+
+## Motivation
+
+- Remove an extra http request from the cli
+- Remove dependency on `update-notifier`
+- Original discussion: https://github.com/npm/rfcs/discussions/118
+
+## Rationale and Alternatives
+
+- Do nothing, keep the same system in place
+- More alternatives?
+
+## Implementation
+
+- Set a `user-agent` header in the requests from the **npm cli**
+- Add a check on the registry that notifies user if a greater version of the cli is available
+- Add a new header in the registry response that alerts for a new version of the cli
+- Also allow customization so that other registry clients can implement/benefit from the same system
+
+## Prior Art
+
+- https://www.npmjs.com/package/update-notifier
+
+## Unresolved Questions and Bikeshedding
+
+TBD


accepted/0000-notification-system-for-cli-updates.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Rationale and Alternatives
+
+- Do nothing, keep the same system in place
+- More alternatives?
+
+## Implementation
+
+- Set a `user-agent` header in the requests from the **npm cli**
+- Add a check on the registry that notifies user if a greater version of the cli is available
```
ljharb, line 21: this needs to take into account the node version; there's very complex logic (https://github.com/nvm-sh/nvm/blob/master/nvm.sh#L150-L269) for figuring out which npm version works on which node version, and it would be pretty bad if npm suggesting updating on a node version where the act of updating completely hosed the entire node install by breaking npm.

accepted/0000-notification-system-for-cli-updates.md
```diff
@@ -0,0 +1,31 @@
+# Notification system for cli updates
+
+## Summary
+
+A new notification system to warn users about new versions of the **npm cli**.
+
+## Motivation
+
+- Remove an extra http request from the cli
```
zkochan, line 9: I am not sure this makes sense. It is just one request per week. During an installation there are hundreds of requests.

ruyadorno: my guess is that when this was brought up it was with regards to commands that should not necessarily hit the registry, such as `npm ls` but i do agree with your point, it might be silly and not worth having this listed in the RFC after all

ruyadorno:
Feedback from [OpenRFC meeting](https://github.com/npm/rfcs/issues/147):
- state of the registry engineering resources makes it difficult to commit to adding new features to the registry right now

accepted/0000-notification-system-for-cli-updates.md
bnb: I think it's worth it to keep it if reducing HTTP calls where they may not be necessary is a goal of the project (I think that's a fair and valid goal).

ruyadorno:
Action Item:
- Update the RFC to reflect recent updates in npm7 beta so that the RFC should be mostly about registry-side changes and the requests the cli is going to perform

owlstronaut:
Implemented differently by https://github.com/npm/cli/commit/d062b2c02a4d6d5f1a274aa8eb9c5969ca6253db, closing this

PR #152, opened 2020-05-24
RFC: npm fund depth

fishcharlie:
[Formatted RFC](https://github.com/fishcharlie/rfcs/blob/fishcharlie/npm-fund-depth/accepted/0000-npm-fund-depth.md)

# What / Why

> This RFC proposes adding a `depth` flag to the `npm fund` command.

## References

* n/a

### DIFF
--- a/accepted/0000-npm-fund-depth.md
+++ b/accepted/0000-npm-fund-depth.md
@@ -1,0 +1,83 @@
+# npm fund depth
+
+## Summary
+
+This RFC introduces a `depth` flag to the `npm fund` command.
+
+## Motivation
+
+Adding this command solves the case of having funding information overload. Currently here is an example output of the `npm fund` command (only about 4 of the items listed are direct dependencies):
+
+```
+dynamoose % npm fund
+dynamoose@2.2.0
+├─┬ https://opencollective.com/typescript-eslint
+│ └── @typescript-eslint/eslint-plugin@2.30.0, @typescript-eslint/parser@2.30.0, @typescript-eslint/experimental-utils@2.30.0, @typescript-eslint/typescript-estree@2.30.0
+├─┬ https://opencollective.com/eslint
+│ └── eslint@6.8.0
+├─┬ https://opencollective.com/mochajs
+│ └── mocha@7.1.2
+├─┬ https://github.com/sponsors/mysticatea
+│ └── regexpp@3.1.0
+├─┬ https://github.com/sponsors/isaacs
+│ └── glob@7.1.6, rimraf@3.0.2
+├─┬ https://github.com/sponsors/sindresorhus
+│ └── globals@12.4.0, ansi-escapes@4.3.1, figures@3.2.0, type-fest@0.11.0, p-limit@2.2.2, make-dir@3.0.2
+├─┬ https://github.com/chalk/ansi-styles?sponsor=1
+│ └── ansi-styles@4.2.1
+├─┬ https://github.com/sponsors/jonschlinkert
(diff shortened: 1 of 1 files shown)


fishcharlie:
This is my first time submitting an RFC to npm 🎉. 🤞 that I did it correctly. However if I didn't, please let me know and I'll make any corrections needed.

I look forward to discussing this RFC and answering any questions that arise with the community!

accepted/0000-npm-fund-depth.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### No change
+
+Benefits:
+
+- More packages are represented in the `npm fund` output. However with this RFC the default is still to view all packages, therefore this isn't much of a drawback.
+
+Drawbacks:
+
+- Funding bloat happens quickly. It's impossible to fund all dependencies, and users are more likely to fund direct dependencies since they know how they interact with them and utilize them directly.
```
ljharb, line 58: what happens if one of your direct deps has no funding info, but it depends on dozens of packages that *do* declare funding info?

accepted/0000-npm-fund-depth.md
```diff
@@ -0,0 +1,83 @@
...
+- More difficult/impossible to get funding information for indirect dependencies.
+- Lack of consistency with other npm commands such as `npm ls`.
+
+## Implementation
+
+Unsure at this time. Likely will need feedback from npm team here.
+
+## Prior Art
+
+`npm fund` is a command unique to npm. There are no other known examples of a similar command in other package manger CLIs.
```
ljharb, line 79:
```suggestion
`npm fund` is a command unique to npm. There are no other known examples of a similar command in other package manager CLIs.
```

fishcharlie: Fixed in https://github.com/npm/rfcs/pull/152/commits/4bff9d4e7792aab0a65a3ebceda1e1a790384533

accepted/0000-npm-fund-depth.md
fishcharlie: @ljharb In this case, if you set `depth=0` I believe it should **not** output the sub-dependencies funding information. It should be strictly limited to printing funding information of your direct dependencies.

I wish there was an existing example of that behavior that we could emulate, however I can't think of a similar example for this case.

Do you think I should clarify that more in the document?

ruyadorno:
hi @fishcharlie, first of all thanks for the contribution 😊 I'm going to leave here the feedback we got from the last time we discussed it (as we went over it again during this week's [npm OpenRFC call](https://youtu.be/A1ApVcb_Sv4)).

In most of npm subcommands, we're shifting the focus away from the `--depth` flag in favor of using the `--all` flag. The `--all` flag is now supported in the `npm ls` as of [npm7 beta](https://blog.npmjs.org/post/627451891844055040/release-v700-beta7) as well as in `npm outdated` (and maybe some other cmds I don't recall at the moment).

There were a bunch of discussions when we first set to remove support to `--depth` in `npm outdated` in a previous RFC that you can check in case you want to dig more on the rationale there: https://github.com/npm/rfcs/pull/133

An important change @isaacs noticed is that we'd still want to preserve the default behavior of `npm fund` printing funding info of your entire install tree, so the RFC should also mention that the `all` config would now need a different default value (current is `false`). This new default value of `all` should now be `null` allowing the subcommands to override it with their expected default values (which will be `false` for **ls** and **outdated** and `true` for **fund**).

Next action item here now is to tweak the RFC to reflect these proposed changes 😄

PR #182, opened 2020-07-17
feat: add npm license RFC

bnb:
# What / Why
It would be ideal to provide license tooling within the CLI itself rather than leaving it to ecosystem OSS tooling or enterprise-grade registry tools/services to implement it.

### DIFF
--- a/accepted/0028-npm-license.md
+++ b/accepted/0028-npm-license.md
@@ -1,0 +1,93 @@
+# Add license reporting to npm
+
+## Summary
+
+It should be possible to get a full report of the licenses from all dependencies. This report should include the license as defined in `package.json` in addition to including a path to any relevant `LICENSE`-ish files within the project. Additionally, end-users should be able to have an allowlist and blocklist of "acceptable" licenses, a way to allow modules with licenses outside of that list that have been triaged, and have a way to install such a list.
+
+## Motivation
+
+{{Why are we doing this? What pain points does this resolve? What use cases does it support? What is the expected outcome? Use real, concrete examples to make your case!}}
+
+License reporting is something that there seems to be a non-trivial amount of ecosystem tooling around - largely built by folks at large corps or by companies that focus on dependencies as a core part of their business in some way - without having any kind of official recognition beyond the "license" property being displayed on the package view of npmjs.com.
+
+Given that this is something that a non-trivial number of users care about _and_ that the state of the ecosystem is relatively fragmented, there's an opportunity to include license tooling as a first-class tool within the ecosystem and both help incrementally improve the ecosystem's awareness/practices around this in addtion to providing blessed built-in tooling that can help resolve a problem the ecosystem is pretty consistently having to build their own tooling for.
+
+## Detailed Explanation
+
+- It should be possible to get a full report of the licenses from all dependencies.
+  - This should be runnable from a single command: `npm audit licenses`
+    - Offline (default)
+      - This report should collect all licenses via the `license` properties from `package.json` files in `node_modules`.
+    - Online (`--online`)
+      - This report should have an `--online` mode that will resolve the dependency tree and fetch the entire dependency tree's licenses without needing the modules on disk.
+    - Output
+      - This report should default to a pretty printed human-readable version.
+      - This report should be able to be output as JSON with a `--json` flag.
+- It should be possible to define a set of allowed, blocked, and ignored licenses
+  - Location
+    - This should be able to be defined both in `package.json` or something like `audit.json`.
+  - Contents
+    - These lists should an be in a `licenses` object with three properties - `allow` (array), `block` (array), and `ignore` (object).
+      - If in `package.json`, `licenses` should be a subproperty of `audit`.
+      - If in `audit.json`, `licenses` should be a top level property.
+      - In `licenses`, `allow` should be an array of values that would _ideally_ be SPDX License Identifiers or SPDX License Expressions but in reality will not end up always being that.
+      - In `licenses`, `block` should be an array of values that would _ideally_ be SPDX License Identifiers or SPDX License Expressions but in reality will not end up always being that.
+      - In `licenses`, `ignore` should be an object with two properties - `licenses` and `modules`.
+        - `licenses` is a list of license values (parsed from `license` in dependencies' `package.json`) to _ignore_ from any kind of checking.
+        - `modules` is a list of modules (parsed from `name` in dependencies' `package.json`) to _ignore_ from any kind of checking.
+  - Behavior on `npm install`
+    - If a `block` property exists, npm should not install any modules on-disk that have a value in `license` that matches any of the values in `block`.
+    - All blocked modules and the path they were introduced through should be reported as an `error` or `warn`.
+- It should be possible to triage licenses that are on disk (in `node_modules`) or would be resolved in the dependency tree
+  - This should be accomplished through `npm audit licenses triage`
+    - Offline (default)
+      - This report should collect all licenses via the `license` properties from `package.json` files in `node_modules`, filtering out any licenses that are in `allow` and  `block` in addition to any licenses or modules in `ignore`, and provide the user the option to `allow`, `block`, or `ignore` the license, one by one.
+    - Online (`--online`)
+      - This report should collect all licenses via the `license` properties from `package.json` files from a resolved dependency tree without needing the modules on disk, filtering out any licenses that are in `allow` and  `block` in addition to any licenses or modules in `ignore`, and provide the user the option to `allow`, `block`, or `ignore` the license, one by one.
+
+## Rationale and Alternatives
+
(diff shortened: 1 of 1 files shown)


bnb:
@ljharb I was intentionally vague in values. Once you allow `MIT` and a few others, you've allowed a majority of the ecosystem. For a few that have incorrect or missing licenses, though, you'd likely either want to add them to the ignore list (under `license` if they're acceptable but invalid SPDX IDs, or under `module` if they're entirely missing their license).

Does that answer your question, or are you asking about the potential of using installable license lists?

I would not be opposed and did consider this but didn't want to go overboard with the initial proposal. I would personally love to see installable license lists, since this would enable something like `npm i @mcdonalds/license-list` as a way to provide large teams a single source of truth to maintain and use. What this API would look like, I am unsure since it shifts the license list out from `package.json`/`audit.json`.

ljharb:
ah, thanks, i didn't read carefully enough. so "module" is for allowed modules - are there disallowed modules?

bnb:
@ljharb this is the structure:

```
"allow": ["MIT", "ISC"], // licenses that are allowed
"block": ["Unlicense", "WTFPL"], // these are not allowed
"ignore": { // these will all be ignored
  "license": ["BIGCOMPANYINTERNALMODULE"], // when npm encounters a module with this license it will be skipped
  "module": ["@bigco/bigmoneyfrontend"] // ignore this module whenever it's encountered no matter what its license is
}
```

ljharb:
I think that "module" needs to support an optional version specifier range - for example, maybe v1 doesn't have a specified license (but is MIT) but v2 is MIT and v3 is GPL, you wouldn't want *all* versions to be allowed, since then v3 would be allowed, but you also wouldn't want to error on v1?

bnb:
@ljharb would this work for you?

```
"allow": ["MIT", "ISC"], // licenses that are allowed
"block": ["Unlicense", "WTFPL"], // these are not allowed
"ignore": { // these will all be ignored
  "license": ["BIGCOMPANYINTERNALMODULE"], // when npm encounters a module with this license it will be skipped
  "module": {
    "@bigco/bigmoneyfrontend": "<semver-expression-like-you'd-expect-in-dependencies-or-devDependencies>"
  } // ignore this module whenever it's encountered no matter what its license is
}
```

ljharb:
To be clear, what I think is ideal is the config format that licensee already has - is there a reason we can't adopt that, since npm already uses licensee?

bnb:
If folks would like me to rewrite this with the licensee format, that's fine. I would prefer to get more input that it's the direction folks would want to go before I do that work 👍🏻

wesleytodd:
> "@bigco/bigmoneyfrontend": "<semver-expression>"

Would be anything parsable by `npm-package-arg`?  Same argument which @isaacs made for the expectation of supporting git repos and file system paths for the "parent" feature, right?

---

Is there anything we can take from the audit resolve spec which can be applied here?  As I mentioned on the call, it seems like a "snooze for 1 month" style is really what is effective to reduce noise.  If you add to "ignore" it will rely on you remembering to check again once the update from the transitive 3 levels deep gets propagated up to you, this will lead to stale info here I think.

bnb:
@isaacs would genuinely appreciate your feedback on this in text, if you have any, before I start editing so I can make sure I don't miss anything - I know you had some thoughts on the RFCs call but my brain did not adequately capture them in a way that I can turn them into a checklist.

And no rush, just wanted to make sure I get the request in and share the reason I've not continued to update.

ruyadorno:
Action item from the OpenRFC Call:
- Explore ways we could potentially use [licensee](https://github.com/jslicense/licensee.js)

bnb:
@ruyadorno it would be nice to hear expanded thoughts on that. Do the project's maintainers want to use the JSON format of licensee? Literally just pull in Licensee as a dep and provide a command as an alias for it? Something else?

ruyadorno:
I'm really sorry about the vague comment @bnb but it was very intentional 😬 I was very stressed out running the call and I honestly don't remember exactly what were the points, just commented here for the sake of not losing track of it (also remembered you mentioning you'd like textual comments for actionable things)

bnb:
Any ideas on how we could progress this? Is just vendoring licensee a thing we could do? Or read that format? 

Happy to work on this, I'm just not sure what next steps requests are from the npm team.

darcyclarke:
Adding `agenda` label just so we make sure to give feedback or next steps by next week

bnb:
next step discussed in today's meeting:

- figure out the specific DX / command strucutre.

It was suggested that an independent command or npm ls output would make sense.

Addressing those two, respectively:

- my guess would be that specifically for controlling incoming dependencies, this specific feature set under `npm audit` does make sense since it is technically management / control of your dependencies. It would also be something you'd want to have run automatically when you're installing _and_ periodically.
- while I'm not opposed to npm ls output of licenses, I think this proposal is distinct from npm ls since it provides actions / filtering that would apply to `npm install` and related commands.

bnb:
@darcyclarke @isaacs putting a relatively straightforward question out there: would we want to take a dependency on `licensee`? Happy to rewrite the proposal to that if it's the path forward.

ljharb:
I would very much prefer it *not* go under audit. CVEs are almost entirely false positives, and validating licenses is something that (based on one's legal department) is 100% certain. I would prefer to keep these tasks separate.

bnb:
>CVEs are almost entirely false positives,

IMO `npm audit` is different than `npm cve` 😅 

Some points on why I think audit is a good place:
- audit is a generic term that _historically_ has been focused on dependencies, but generically indicates... auditing your dependencies, one element of which is security.
- audit is already hooking into `npm install` in a way that users are used to. My expectation would be that we'd not want to continue modifying that experience, but rather tweak the modifications that have already been made.
- having a one stop shop of all kinds of meta management of dependencies feels like a better design decision than having to use multiple commands. Whether it's end-users or CI, having `npm audit` do security _and_ licenses (with the option of narrowing down the scope with something like `npm audit security` and `npm audit licenses`) rather than `npm audit` and `npm licenses` seems like the better option.
- this also sets up `audit` as a place for future expansion in this space. When, for example, there's some kind of RFC proposing enhanced integrity checks or something else that's "auditing" your dependencies, it could also go into audit. This would further enhance the toolset that is audit, which IMO - again - is good DX. Whether or not this is a design decision that the npm CLI wants to make is, of course, an important distinction to make.

ljharb:
You're right, but given that the current implementation of `npm audit` is basically solely CVEs, and thus the baggage of that ecosystem taints the term, I'm still not sure it's the best place.

My concern is that I very explicitly and intentionally set NPM_CONFIG_AUDIT to `false` due to this, which means i'd automatically disable license checking, and I'd have to figure out (assuming it exists) how to configure "run audit, but don't run the security checks".

bnb:
@ljharb would you be okay with a future where you could `NPM_CONFIG_AUDIT_SECURITY=false`  (`security` vs. `vulnerabilities` vs. `cve` vs. whatever bikeshedding aside) to just disable the security bits but still get the license bits/whatever future bits?

This would also assume that `NPM_CONFIG_AUDIT=false` disables everything and `NPM_CONFIG_AUDIT_LICENSES=false` only disables licenses.

bnb:
Additionally, I'd assume `--no-audit-security` and `--no-audit-licenses` flags would be included in this.

bnb:
I've updated the RFC to reflect using `licensee` while retaining the originally proposed DX.

bnb:
Updated further with the `npm audit --fix` and related changes that were discussed in today's RFC meeting.

bnb:
remaining unresolved question:

- package.json/audit.json/both

are we okay with introducing an `audit.json` configuration file? are we okay shoving this configuration into package.json? are we okay with doing both and having one take priority?

bnb:
worth noting, I've also opened an issue that looks like it'll result in a PR from me this week to licensee to replace `whitelist` with `allowlist`, with +1's to that path from the current maintainers: https://github.com/jslicense/licensee.js/issues/72

ljharb:
I'd prefer having "only not package.json" - it avoids inflating the size of the packument/published package, and it avoids adding more stuff into package.json.

However, some folks might prefer avoiding "another config file", so I'm not sure what would be best.

bnb:
depending on how https://github.com/npm/rfcs/pull/564 goes, it might be worth simply waiting for that to be implemented and the subsequent [command mapping](https://github.com/npm/rfcs/blob/e63094a66d3b690561c129599861f221ffb7305c/accepted/0000-dependency-selector-syntax.md#command-mapping-to-query-output) feature, so as not to waste time on implementing "normally" and then migrating.

ljharb:
Given that licensee already exists, I’m not sure what the value is of waiting?

bnb:
> Given that licensee already exists, I’m not sure what the value is of waiting?

I explored doing it with licensee in a pairing session with @izs and the conclusion we came to was that doing so was massively limiting, given that it would preclude a lot of the benefits you get from doing such a thing with Arborist, a number of which are baseline expectations folks would have of an npm CLI command.

ljharb:
Ah, that's new information, thanks.

PR #183, opened 2020-07-17
RFC21: bring back subset of npm_package_* environs

isaacs:
Discussion: https://twitter.com/JessTelford/status/1258549958728286208

CC: @JessTelford

### DIFF
--- a/accepted/0021-reduce-lifecycle-script-environment.md
+++ b/accepted/0021-reduce-lifecycle-script-environment.md
@@ -91,6 +91,18 @@ that config defaults are maintained, without blowing up the size of the
 environment for lifecycle processes, or requiring access to the npm config
 subsystem in every npm CLI dependency.
 
+In addition to these, the following environment variables will be
+preserved from npm v6's set, in order to provide an easier upgrade path:
+
+* `npm_package_name`
+* `npm_package_version`
+* `npm_package_config_*` Each key in the `config` object will be included,
+  but npm will _not_ override values with a `<pkgname>:<keyname>` config
+  value if one exists.
+* `npm_package_engines_*`
+* `npm_package_main`
+* `npm_package_bin_*`
+
 ## Rationale and Alternatives
 
 Possible alternatives:


isaacs:
Is this enough? Are there any on this list we shouldn't include?

accepted/0021-reduce-lifecycle-script-environment.md
```diff
@@ -91,6 +91,18 @@ that config defaults are maintained, without blowing up the size of the
...
+In addition to these, the following environment variables will be
+preserved from npm v6's set, in order to provide an easier upgrade path:
+
+* `npm_package_name`
+* `npm_package_version`
+* `npm_package_config_*` Each key in the `config` object will be included,
+  but npm will _not_ override values with a `<pkgname>:<keyname>` config
+  value if one exists.
+* `npm_package_engines_*`
+* `npm_package_main`
```
ljharb, line 103: what about `npm_package_exports`, `npm_package_private`, etc? or are these just the ones that have known uses?

isaacs: These are the ones that have known uses, have been around a long time, and aren't terribly complicated to put in the environment.

The `"exports"` field is an object, and if you have something like `"exports": {".": {"import": "./index.mjs"}}`, that'd env-ify into `npm_package_exports___import="./index.mjs"` (since `.` is stripped out) which is a bit weird.

In other words, there is a point where it's fair to say, look, if you wanna know what's in `package.json`, write a JavaScript program and parse it or do `require('./package.json')`.  The line drawn by RFC21 as it currently stands is "anything"; if you want package.json at all, you parse it, here's the filename, have fun.  The pushback (which I think is reasonable) is that there are a few things you might want to put in command lines in the package.json itself, without having to write a whole other script to do it.

So this is just the stuff that's likely to be used by a package lifecycle script, where you might want to have something like `"scripts": {"start":"node $npm_package_bin_start_server"}` or something.

isaacs:
Landed and published in 7.0.0-beta.4.

Issue #158, opened 2020-06-04
[RRFC] ability to run scripts standalone, without their pre- and post- associates


morgaan:
## Motivation ("The Why")

To provide a way to run npm script without the pre- and/or post- scripts

### When

When running `npm run <script>`

### Where

npm cli

### How

By passing flags for example:

- `npm run test --no-pre`, or
- `npm run test --no-post`, or
- `npm run test --no-pre-post`

#### Current Behaviour

pre- and post- scripts cannot be skipped, workaround exist but clumsy, for example:

```
"scripts": {
    "rawtest": "mocha -r ts-node/register src/**/*.spec.ts",
    "pretest": "<some flow that isn't always desired>"
    "test": "npm run rawtest",
  }
```

#### Desired Behaviour

- the pre- scripts should be skipped when run `npm run <script> --no-pre`
- the post- scripts should be skipped when run `npm run <script> --no-post`
- both pre- and post- scripts should be skipped when run `npm run <script> --no-pre-post`

### References

Was added by @beenotung there https://github.com/npm/cli/issues/484

gingur:
I think this could also solve https://npm.community/t/ignore-scripts-ignores-all-scripts/1403 , being able to run `npm install` without "post" scripts (but let dependency installs work as expected)

askirmas:
I found 2 better syntaxes until it is not implemented on core level
```json
{
  "test--dry": "mocha",
  "start --dry": "node server"
}
```
```bash
npm run test--dry && npm run start\ --dry
```
Besides, [git run without hooks](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--n)

ljharb:
@askirmas `npm run start -- --raw` and `npm test -- --raw` works without all those `--` hijinks in the scripts.

askirmas:
> @askirmas `npm run start -- --raw` and `npm test -- --raw` works without all those `--` hijinks in the scripts.

`start --raw` will run everything: `prestart`, `start` with arg `--raw` and `poststart`. That's not the point of issue

ljharb:
In that case, I use your technique often :-) about 240 of my projects all use `npm run tests-only` for just the tests, because `npm test` invokes `npm run pretest` and `npm run posttest`, and in CI, i don't want the linter running on the 100 node versions i want the tests running, for example.

askirmas:
But this hack makes *package.json* to look dirty.

ljharb:
[...]

PR #185, opened 2020-07-20
RFC: Add ability to skip script hooks

lumaxis:
See content.

### DIFF
--- a/accepted/0028-add-ability-to-skip-hooks.md
+++ b/accepted/0028-add-ability-to-skip-hooks.md
@@ -1,0 +1,25 @@
+# Add the ability to skip pre- and post hooks when running a script
+
+## Summary
+
+Today, when running any `npm` script, there is no way to skip a script's pre and/or post hook. I'd like this to be changed.
+
+## Motivation
+
+When running `npm install` today, it's very easy to ignore any `postinstall` scripts, for example, by using the `--ignore-scripts` option. When using `npm test` (which, on first glance, seems like a top-level command as well) however, the same does not work.
+Additionally, if for example you have a `pretest` script defined such as `"pretest": "npm run lint"` but you would want to run skip linting this time for whatever reason, that is not possible other than completely removing the `pretest` script from `package.json` temporarily.
+
+## Detailed Explanation
+
+I would suggest to change the behavior of the `--ignore-scripts` option to also skip the `pre` and `post` hooks of a given command.
+To me, this would feel consistent with the behavior of how the flag works with `npm install` already today. It would establish the semantics as "run only the command/script I specified, nothing else".
+
+## Rationale and Alternatives
+
+Instead, we could add a separate flags `--ignore-hooks` (or similar). This would have the benefit that it does not alter the existing behavior of `--ignore-scripts`. It would however be difficult to come up with a descriptive name for this new option and I'd imagine the differences and superficial overlap with `--ignore-scripts` be difficult to explain and not intuitive to understand.
+
+Alternatively, we can just not change anything and rely on users manually editing their `package.json`
+
+## Implementation
+
+_skipped for the moment_


isaacs:
It turns out we're already running the specified script for `npm run-script` (and thus for `test`, `start`, etc.) when `--ignore-scripts` is set.  (v6 does not do this, so this is an overlooked breaking change, now turned into a feature by this RFC ;)

The whole diff to do this in v7 is:

```diff
diff --git a/lib/run-script.js b/lib/run-script.js
index 270a91ab7..733534421 100644
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -80,11 +80,13 @@ const runScript = async (args) => {
 
   // positional args only added to the main event, not pre/post
   const events = [[event, args]]
-  if (scripts[`pre${event}`]) {
-    events.unshift([`pre${event}`, []])
-  }
-  if (scripts[`post${event}`]) {
-    events.push([`post${event}`, []])
+  if (!npm.flatOptions.ignoreScripts) {
+    if (scripts[`pre${event}`]) {
+      events.unshift([`pre${event}`, []])
+    }
+    if (scripts[`post${event}`]) {
+      events.push([`post${event}`, []])
+    }
   }
 
   const opts = {

```

ljharb:
ooh, so to confirm, `npm test --ignore-scripts` in npm 7 will no longer run pretest/posttest??

isaacs:
Well... in the current implementation on the beta branch, it'll run test _with_ pretest/posttest.  If we accept this RFC, it'll run without pretest/posttest.  If we do a different thing, it'll do a different thing :)

Whether the current implementation is a feature or a bug is something we'll discuss wednesday in the call.

isaacs:
Personally, I'm of the opinion that `npm test` should always run your test, even if you have `--ignore-scripts` set.  Because that's the command you're running, it's a bit silly to have a flag that says "don't do the thing I'm telling you specifically to do".  But I think it's reasonable to assume in that case that it should not run _other_ scripts that ride along with that command, sort of in the spirit of `--ignore-scripts`, which is what this RFC proposes.

ljharb:
I agree -  but it *shouldnt* run the pre and post scripts, ie, I’m strongly in favor of this rfc.

The behavior in npm <= 6 is bizarre in that it runs *nothing* :-)

ruyadorno:
Related to: https://github.com/npm/rfcs/issues/158

askirmas:
`npm run $script --ignore-scripts` looks too weird. It should be based on word `hooks` or `lifecycle` like `--without-hooks` or `--no-lifecycle`. Also can be leading `--omit-`, `--skip-`, `--dry-`.  

`--ignore-scripts` omits *dependency's*(1) *package*(2) lifecycle but discussed option skips lifecycle of *self*(1) *script*(2). It will be puzzling naming coincide and not reuse-option.

Besides, is it ok that `install` and `install {package}` looks like completely different scripts? Considering
```json
/* ./package.json */
{
  "scripts": {
    "preinstall": "echo 'main preinstall'",
    "postinstall": "echo 'main postinstall'"
  },
  "dependencies": {
    "sub": "file:sub"
  }
}

/* ./sub/package.json */
  "name": "sub",
  "version": "1.0.0",
  "scripts": {
    "preinstall": "echo 'sub preinstall'",
    "postinstall": "echo 'sub postinstall'"
  },
```
```bash
$ npm install
# main preinstall
# main postinstall
$ npm install ./sub
sub preinstall
sub postinstall
```

isaacs:
We discussed this on the call and decided that adding another term for this (even the term `lifecycle` which is used internally) is not ideal for UX.

Going to ratify this proposal as-is, and get the implementation in one of the next few v7 beta releases for play-testing.

lumaxis:
@isaacs Do we merge this or what's the process for the PR?

PR #197, opened 2020-08-08
RFC: Open Source Pay It Forward system (OSPIF)

bertyhell:
Create initial spec for Open Source Pay It Forward system 

created by  @bertyhell @gommorach


bertyhell:
After the call on [26 aug](https://github.com/npm/rfcs/issues/212) i think this issue can be closed. @ruyadorno @darcyclarke 

The way forward seems to be to create the OSPIF license file, so developers can choose to require payments in order to use their library. As mentioned https://licensezero.com/ might be a good place for me to inform about that.

The role of accountant already seems to exist: https://backyourstack.com/ although in a less automated form, where you have to upload your package.json file manually. An automated upload of the package.json file could easily be added to the build process using an npm run command and a separate package.

PR #201, opened 2020-08-12
npx: use registry from .npmrc

Christian24:
At the moment anything executed with `npx` is resolved against npm registry. However, for `npm` users can configure their own registry url. It would be good if `npx` respected this configuration option as well.

In corporate environments the official npm registry might not be accessible, because a corporate registry like `Nexus` or `verdaccio` is available. In these scenarios `npx` currently does not work, because it cannot reach npm registry. The corporate registry is configured in `.npmrc`, so the configuration could be honored by `npx` too to make it usable in these environments too. 

## References
 Fixes https://github.com/npm/npm/issues/19113

### DIFF
--- a/accepted/0000-npx-registry-config.md
+++ b/accepted/0000-npx-registry-config.md
@@ -1,0 +1,35 @@
+# npx should respect registry in .npmrc
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


Christian24:
This should no longer be needed with npm 7 it seems. Yay. :)

PR #210, opened 2020-08-22
RFC peer-specific overrides

KilianKilmister:
[formatted RFC](https://github.com/KilianKilmister/rfcs/blob/latest/accepted/0000-peer-overrides.md)


darcyclarke:
Hey @KilianKilmister! Appreciate you putting such a lengthy RFC together & coming out to the last couple of RFC meetings to explain this in more detail. As we spoke on the last call, at this time we don't intend to support package-specified `overrides`, only root-level. The discussion did bring up a good point that we will need to communicate this better once we get the implementation stage (potentially through a publish prompt but definitely in our docs).

With that in mind, I've left this current RFC off the agenda for today's meeting as I think it probably requires significant reworking & has gotten pretty verbose without much interest from others to collaborate.

Unfortunately, right now, the proposal would not be accepted as it stands based on both the implementation & how it's documented. I'll most likely close this PR if there isn't any traction/interest from others within the next week or so.

If you wanted to potentially break up some of the ideas you have into smaller [discussions](https://github.com/npm/rfcs/discussions) or an [RRFC](https://github.com/npm/rfcs/issues/new/choose), that may be a better way for us to move forward with these ideas.

KilianKilmister:
I really would have apreciated an earlier notice, as my prep-work for today has gone to waste now.

> Unfortunately, right now, the proposal would not be accepted as it stands based on both the implementation & how it's documented. I'll most likely close this PR if there isn't any traction/interest from others within the next week or so.

Of course not, as the scope had to drastically change based on the clarifications during the discussions. I planned on reformatting it and cleaning it up before today, but i couldn't really do that before some points were adressed.
Sadly i didn't recieve an answer for a major question of mine since i posted it a week ago. see: #218 

>  As we spoke on the last call, at this time we don't intend to support package-specified overrides, only root-level.

I mean that's the whole reason RFCs exist, to make a case about changing something (or the lack of it) and trying to change what's currently intended.

But beyond this current intention, this wasn't really talked about. At the meeting, the mechanism and functionality was clearly misunderstood. And basically all the time that was spent on this RFC had to be used to explain some of the mechanisms and plans for the changes from v6 to v7

> [...] without much interest from others to collaborate.

The question is wether this is because of the afformentioned miscommunication or because of an issue with the principle, because i have not recieved any feed-back on the latter, whic means:

- i don't know what some of the arguments against it are, as the ones i heard were/are already adressed in the RFC
- i don't know if these arguments have been adressed in any satisfying way
- what overall changes could be made to make the RFC more agreeable,
- wether the suggested implementation is regarded as feasable by people who know the code base much better than me
- wether there are existing alternatives i didn't consider,
- or wether there is a hypothetical alternative that would be better or more agreeable.

> If you wanted to potentially break up some of the ideas you have into smaller discussions or an RRFC, that may be a better way for us to move forward with these ideas.

It's really just one thing, but that's not important right now.
I have already invested a very large amount of time and effort into this with having to learn the architecture of arborist (a project i only knew the name of before i started), writing and rewriting and annotating, preparing for meerings, working out the concepts and trying to identify potential issues with all of them.

As i mentioned earlier, the RRFC/question i posted a week ago wasn't conclusively answered and i recieved zero feed-back on the RFC itself, so I have to ask myself wether filing other RRFCs or using [discussions](https://github.com/npm/rfcs/discussions) will have a different outcome, because if they don't I can't possibly justify investing more time and effort on this anywhere near the level of commitment i've held for the past weeks.

Right now i'm completly in the blind. All the things i know are:

- package specific overrides aren't planned for now, which is the whole reason i started this RFC.
- people aren't in favour of deep-overrides, something which is impossible with how I planned to do it anyways.

This just isn't enough for me to go off of.

### My only question right now

Is there something which can show/tell me, further commitment to this cause will have an effect? How can i know that my open questions will recieve an answer or wether i will get feedback on my general idea? And if that does happen, and i would competly reformat the RFC and adjust it accordingly, will i get feed-back for the specifics of my proposal?

KilianKilmister:
the closing of [this issue](https://github.com/npm/cli/issues/1681) together with the release of v7-beta.12 closes this topic for the time being.

PR #224, opened 2020-09-09
RFC: No auto-install for peerDependencies marked as optional

### LINKED ISSUE #221: [RRFC] No auto-install for peerDependencies marked as optional
## Motivation ("The Why")

As a package author, I want to specify supported version ranges of `peerDependencies`. I also want to set some peers as optional with `peerDependenciesMeta` and `{ "optional": true }` to avoid automatically installing them when an end-user installs a package I've authored.

### Example

As an example, let's take a database ORM package. The ORM package could support upwards of a dozen different databases and require an end-user to install the database adapter(s) needed for their project. These database adapters could be huge and could have `peerDependencies`, `postinstall` scripts, and other requirements.
It could significantly bloat a project's install time and dependency tree to pull in tons of packages that an end-developer would never even use.

#### Current Behaviour

npm v7 currently installs all `peerDependencies`, regardless if set as optional or not using `peerDependenciesMeta`.

#### Desired Behaviour

Personally, it seems ideal to me to not auto-install `peerDependencies` marked as optional. I believe it would be a massive benefit to the package ecosystem. 

`peerDependencies` not marked as optional should still install automatically.

### References

* n/a

jrylan:
## Summary

Avoid automatically installing `peerDependencies` marked as optional using `peerDependenciesMeta` and `{ "optional": true }`.

## Motivation

Package authors want to specify supported version ranges of `peerDependencies`. They also want to set some peers as optional with `peerDependenciesMeta` and `{ "optional": true }` to avoid automatically installing them when an end-user installs their package.

As an example, let's take a database ORM package. The ORM package could support upwards of a dozen different databases and require an end-user to install the database adapter(s) needed for their project. These database adapters could be huge and could have `peerDependencies`, `postinstall` scripts, and other requirements.

It could significantly bloat a project's install time and dependency tree to pull in tons of packages that an end-developer would never use.

## Detailed Explanation

- When a peer dependency is marked as `{ "optional": true }` using `peerDependenciesMeta`, it should not install automatically.

- `peerDependencies` without a `peerDependenciesMeta` value of `{ "optional": true }` should still install automatically.

### Example `package.json`

```json
{
  "peerDependencies": {
    "react": "^17.0.0"
  },
  "peerDependenciesMeta": {
    "react": {
      "optional": true
    }
  }
}
```

## References
  Related to #221
  Fixes #221
  Closes #221

### DIFF
--- a/accepted/0030-no-install-optional-peer-deps.md
+++ b/accepted/0030-no-install-optional-peer-deps.md
@@ -107,14 +107,4 @@ It's less than ideal and complicates publishing, to name a few issues:
   same effect. Instead, this proposal omits the need for a new field and
   leverages the already well-known and used `{ "optional": true }`.
 
-## Unresolved Questions and Bikeshedding
-
-- Is this the best option to move forward? Should `{ "autoinstall": false }` be
-  implemented instead?
-
-- Are pnpm and Yarn able to follow this implementation? cc/ @arcanis, @zkochan
-
-- Can a solution be implemented before npm v7 is final? It would be a breaking
-  change to alter the current behavior afterward.
-
 [0025-install-peer-deps]: https://github.com/npm/rfcs/blob/latest/accepted/0025-install-peer-deps.md
--- a/accepted/0030-no-install-optional-peer-deps.md
+++ b/accepted/0030-no-install-optional-peer-deps.md
@@ -1,0 +1,120 @@
+# No auto-install for peerDependencies marked as optional
+
+## Summary
+
+Avoid automatically installing `peerDependencies` marked as optional using
+`peerDependenciesMeta` and `{ "optional": true }`.
+
+## Motivation
+
+Package authors want to specify supported version ranges of
+`peerDependencies`. They also want to set some peers as optional with
+`peerDependenciesMeta` and `{ "optional": true }` to avoid automatically
+installing them when an end-user installs their package.
+
+As an example, let's take a database ORM package. The ORM package could support
+upwards of a dozen different databases and require an end-user to install the
+database adapter(s) needed for their project. These database adapters could be
+huge and could have `peerDependencies`, `postinstall` scripts, and other
+requirements.
+
+It could significantly bloat a project's install time and dependency tree to
+pull in tons of packages that an end-developer would never use.
+
+## Detailed Explanation
+
+- When a peer dependency is marked as `{ "optional": true }` using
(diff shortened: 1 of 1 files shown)


jrylan:
cc/ @arcanis, @zkochan - Are pnpm and Yarn able to follow this implementation? Any objections?

arcanis:
I've been clear that we won't change the peer dependencies default behavior (ie they won't be auto-installed), so in our case this probably won't affect us.

mkay581:
@arcanis even after [RFC 25](https://github.com/npm/rfcs/blob/latest/accepted/0025-install-peer-deps.md) is implemented? I realize this is a little off-topic but is there some other yarn thread where this comment is better suited?

arcanis:
As you can see in https://github.com/npm/rfcs/pull/43 and https://github.com/microsoft/types-publisher/pull/655#issuecomment-547375861, I've made it clear that I had significant concerns about RFC 25. I don't think the critical points raised were addressed, and as a result I don't think it would be reasonable to implement it.

zkochan:
I as well wasn't planning to implement #43 in pnpm. If users will demand it, I might consider some solutions. This RFC seems fine

jrylan:
@darcyclarke am I clear to remove the following from my PR?

```
## Unresolved Questions and Bikeshedding

  - Is this the best option to move forward? Should `{ "autoinstall": false }` be
   implemented instead?

  - Are pnpm and Yarn able to follow this implementation? cc/ @arcanis, @zkochan

  - Can a solution be implemented before npm v7 is final? It would be a breaking
   change to alter the current behavior afterward.
```

##  1. Is this the best option to move forward? Should `{ "autoinstall": false }` be implemented instead?

Seems like there is majority consensus that `{ "optional": true }` is the way to go?

##  2. Are pnpm and Yarn able to follow this implementation?

Yarn won't implement, pnpm might implement if user demand is there.

## 3. Can a solution be implemented before npm v7 is final? It would be a breaking change to alter the current behavior afterward.

I think everyone is on the same page that this needs to be resolved before the final v7 release.

ruyadorno:
ooops, we should merge/accept here too, reopening

jrylan:
@ruyadorno I'll edit to remove the `## Unresolved Questions and Bikeshedding` section.

ruyadorno:
Thanks for the contribution @jrylan 😄

PR #232, opened 2020-09-18
npm audit for a not yet installed package

### LINKED ISSUE #223: [RRFC] npm audit <package> for a not yet installed package
## Motivation ("The Why")
In some environments (aka corporate) the NPM Registry might be behind a firewall like [Nexus Firewall ](https://www.sonatype.com/nexus/firewall?smtNoRedir=1)(which automatically blocks vulnerable packages and responds to npm with status code 403. It would be nice to have the ability to query for vulnerabilities before installing a package. 

### Example
`npm audit <package name> --remote`
This would probably not have to report back a detailed report, a summary of vulnerabilities (low, moderate or high) would probably be enough. 
### How
Query npm audit for a not yet installed, single package. 
#### Current Behaviour
There is nothing I am aware of. 

#### Desired Behaviour
Report a summary of vulnerabilities for a not yet installed package back.

````
foo package security summary:

0 high vulnerabilities 
0 moderate vulnerabilities 
1 low vulnerability 
````
### References

* n/a

Christian24:
Describes functionality for the CLI that allows users to request an advisory (aka npm audit report) on a package before installing it. 

## References
 Closes #223

### DIFF
--- a/accepted/0000-audit-package.md
+++ b/accepted/0000-audit-package.md
@@ -1,0 +1,36 @@
+# Audit for a not yet installed package
+
+## Summary
+
+Allows the user to request and receive an audit result for a package that has not yet been installed. 
+
+## Motivation
+
+* In some environments (aka corporate) the npm registry might be behind a firewall like [Nexus Firewall](https://www.sonatype.com/nexus/firewall?smtNoRedir=1) (which automatically blocks vulnerable packages and responds to npm with status code 403. In those scenarios it is hard to say which package is causing trouble. 
+* packages might have vulnerabilities, so security implications are useful to know before installation.
+
+
+
+## Detailed Explanation
+
+To query audit information on a specific package the `view` command is extended with an `--audit` parameter. The CLI would then query the `advisory endpoint` and get a full advisory for the specified package and version as well as its dependencies. The advisory is similar to `npm audit`, but references to `npm audit fix` can be omitted. 
+
+## Rationale and Alternatives
+
+* An alternative would be to provide this information on the package detail page in on npmjs.com, but this would probably be harder to implement (since data would need to be cached) and it might be taken as "blame" by package authors. 
+
+## Implementation
+
+{{Give a high-level overview of implementation requirements and concerns. Be specific about areas of code that need to change, and what their potential effects are. Discuss which repositories and sub-components will be affected, and what its overall code effect might be.}}
+
+{{THIS SECTION IS REQUIRED FOR RATIFICATION -- you can skip it if you don't know the technical details when first submitting the proposal, but it must be there before it's accepted}}
+
+## Prior Art
+
+* This basically reuses almost the entire `npm audit` output. 
+
+## Unresolved Questions and Bikeshedding
+
+{{Write about any arbitrary decisions that need to be made (syntax, colors, formatting, minor UX decisions), and any questions for the proposal that have not been answered.}}
+
+{{THIS SECTION SHOULD BE REMOVED BEFORE RATIFICATION}}


ruyadorno:
I believe it should be an opt-out, so everytime an user runs `npm view <pkg>` it should notify you about any vulns by default, so in order to opt-out you'd need to `npm view <pkg> --no-audit` and any users that already chose opt-out from audits in their `.npmrc` files/configs (using `audit=false`) will already be opting-out here.

ljharb:
@ruyadorno how would that affect `npm view --json`? Would it risk slowing down this often-used command?

Christian24:
So, even though there have been no new comments here, I was pretty convinced last week that putting this information into the `npm audit` command makes the most sense (as was also the plan of the #223), since that is the context most users are familar with audit results.

However, @ruyadorno I agree with you that this is package data and it should possibly be available as output by npm view by default. So, I added the output to npm view too. 

@ljharb I added a `--no-audit` parameter for you, which makes this more in line with `install`.

I have updated the RFC accordingly.

darcyclarke:
@Christian24 I think we [only discussed making this optional information for `npm view`](https://github.com/npm/rfcs/blob/latest/meetings/2020-09-23.md) (potentially behind a flag like `--with-audit`); I recall us not wanting to bloat/slow down `npm view` since all of the information shown in that command can be fetched from the registry Packument (ie. package + document) today. 

I'd probably revert & keep this RFC scoped to `npm audit view` for now & update it to include references into how we would implement this. 

As I think I also noted in the last call, we're going to probably need to create a temp directory to install into & resolve the package's dependencies. How the information will look when it's logged from `npm audit view x` is probably subtly different from how `npm audit` looks today as well (ie. we'd probably want to remove references to addressing/resolving issues w/ `"npm audit fix --force"` or `"Will install <x>, which is a breaking change"` etc.). Good starting points to look for where we'd update/add support are in - [`lib/utils/audit.js`](https://github.com/npm/cli/blob/release/v7.0.0/lib/audit.js), [`lib/utils/reify-output.js`](https://github.com/npm/cli/blob/release/v7.0.0/lib/utils/reify-output.js) & [`npm-audit-reporter`](https://github.com/npm/npm-audit-report)). 

I think we can circle back on this once you've updated/added detail on how the implementation/output would work/look (as noted above).

P.S. For fun, I wrote a quick bash script to do this/walk through the steps in the interim https://gist.github.com/darcyclarke/6d9e9de555997e9aa9fe828fe1fdef7d

ljharb:
Do note, because of `file` deps, creating a temp directory for this will be a bit trickier than it appears: https://github.com/ljharb/aud/issues/2

Christian24:
@darcyclarke thanks for the feedback.

> @Christian24 I think we [only discussed making this optional information for `npm view`](https://github.com/npm/rfcs/blob/latest/meetings/2020-09-23.md) (potentially behind a flag like `--with-audit`); I recall us not wanting to bloat/slow down `npm view` since all of the information shown in that command can be fetched from the registry Packument (ie. package + document) today.

Not that I am against this, but I remember @ruyadorno pointing out it should be opt-out and @ljharb has a use case for `view`, where he uses parameters anyway and so another parameter seemed fine to me, especially since this makes the parameter in line with `install`.

> I'd probably revert & keep this RFC scoped to `npm audit x` for now & update it to include references into how we would implement this.
> 

Okay, fine by me. I think adding it to `audit` is definitely enough for now. It just seemed to me some people wanted to add this to `view` too. 

> As I think I also noted in the last call, we're going to probably need to create a temp directory to install into & resolve the package's dependencies. 

Would we have to do this? I think this would not work in instances where packages are restricted by a 403 (https://github.com/npm/rfcs/blob/latest/implemented/0015-add-restricted-versions-handling.md). This is exactly what happens with Nexus Firewall for us and is the reason why I introduced the RFC. In our case any not yet used dependency with a security vulnerability is forbidden, so we couldn't even install it. So, we wouldn't get audit data. The way I understood @isaacs is that he wanted to use the advisory endpoint. 

How the information will look when it's logged from `npm audit x` is probably subtly different from how `npm audit` looks today as well (ie. we'd probably want to remove references to addressing/resolving issues w/ `"npm audit fix --force"` or `"Will install <x>, which is a breaking change"` etc.). Good starting points to look for where we'd update/add support are in - [`lib/utils/audit.js`](https://github.com/npm/cli/blob/release/v7.0.0/lib/audit.js), [`lib/utils/reify-output.js`](https://github.com/npm/cli/blob/release/v7.0.0/lib/utils/reify-output.js) & [`npm-audit-reporter`](https://github.com/npm/npm-audit-report)).

There is some info in the RFC already, but I will check it against v7 and will try to update with some implementation details. 

> I think we can circle back on this once you've updated/added detail on how the implementation/output would work/look (as noted above).
> 
> P.S. For fun, I wrote a quick bash script to do this/walk through the steps in the interim https://gist.github.com/darcyclarke/6d9e9de555997e9aa9fe828fe1fdef7d

ruyadorno:
> Okay, fine by me. I think adding it to `audit` is definitely enough for now. It just seemed to me some people wanted to add this to `view` too.

I was on the camp of moving it to `npm view` when first proposed but the more I think about it the more I think it should really live in `npm audit` instead as initially proposed (really sorry about the back and forth @Christian24 😬).

Making the link with a previous discussion we had about validating license files, it was proposed that the license validation should also live along with **audit** and the main reason being that auditing was from its inception always intended to be more than just vulnerabilities check and it should encompass multiple kinds of validation for your project. 

With that in mind it seems to me that having individual auditing of packages live in `npm audit` is a better long-term solution and it could also potentially provide these different validations (like license scan) if they ever get implemented.

Christian24:
@darcyclarke I tried your script. With npm v6 at least I get a 403 here. Couldn't try npm v7, because corporate restrictions. `install --package-lock-only` seems to get the 403. 

@ruyadorno Understood. I will remove all references to `view` from the RFC.

Christian24:
Sorry folks, this took quite a while. Since @darcyclarke's script does not work for me, I tried to go the route of adding a new endpoint for the CLI to query. I added some ideas of how this could be implemented. This might potentially be wrong, so it would be cool if @ruyadorno, @isaacs or someone from the CLI team could take a look. Additionally, I added some information on the `fix` part of the `npm audit` report. If possible it would be cool to keep this output, but only change it slightly to give users information on whetever issues can be fixed.

PR #235, opened 2020-09-21
Allow server generated header values

doddi:
Allow a server to generate values for header configuration as detailed in https://github.com/npm/rfcs/pull/138

This rfc is in addition to the aforementioned rfc

### DIFF
--- a/accepted/0000-server-generated-http-header-configuration.md
+++ b/accepted/0000-server-generated-http-header-configuration.md
@@ -1,0 +1,52 @@
+# Server generate HTTP header configuration
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


ljharb:
Does this have privacy concerns? Just like cookies, it seems like it would let a registry track unauthenticated users (and auth’d users’ machines) without the user’s knowledge or permission.

darcyclarke:
Removing from agenda for now as this needs to come after https://github.com/npm/rfcs/pull/138 &, as @ljharb noted & we spoke briefly on the call, this sounds a lot like cookies & we may not want to reinvent the wheel here.

PR #272, opened 2020-11-03
RFC: add preferDev field to package.json

Ginden:
Certain packages are used almost exclusively for development. Eg. typescript, tslint etc. It's usually a mistake to add them as normal dependencies.

I suggest to add field "preferDev": boolean (with default false) to package.json. User can't install preferDev package as production package without specifying `--save-prod` flag.

## References

npm/cli#1337

### DIFF
--- a/accepted/0000-add-prefer-dev-to-package-json.md
+++ b/accepted/0000-add-prefer-dev-to-package-json.md
@@ -1,0 +1,28 @@
+# Add field "preferDev" to package.json
+
+## Summary
+
+Add `preferDev` field to `package.json` to indicate that libraries setting this field to `true` are intended as development dependencies.
+
+## Motivation
+
+It's usually mistake to include packages like `typescript`, `eslint`, `webpack` as normal dependencies. It unnecessary increases size of `node_modules` and potentially bundle size.
+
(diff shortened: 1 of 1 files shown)


ljharb:
eslint presumably would preferDev, but it’s a runtime dep in standard, for example. I’m not sure package authors have sufficient context to know whether their package needs to be dev or prod.

wesleytodd:
If this goes the direction of a warning, I would think you only show it on explicit package installs:

```
npm i eslint --save-prod
```

and not:

```
npm i
```

isaacs:
My comments on this are essentially the same as [my comments on `preferPeer`](https://github.com/npm/feedback/discussions/100#discussioncomment-136744), for all the same reasons.

RilDev:
Also making sure that is appears as the default install choice on NPM would be convenient!

npm-change-default-install-recomendation

Using the format `npm i [package-name] -D` for example.

Reddit: [How To Change The Default Install Suggestion On NPM](https://www.reddit.com/r/npm/comments/k3sy0m/how_to_change_the_default_install_suggestion_on/)

totongcute (approved):

PR #273, opened 2020-11-04
RFC: npm workspaces - Config management

ruyadorno:
Adds a **workspaces** config management workflow.

[See RFC]()

### DIFF
--- a/accepted/0000-workspaces-config-management.md
+++ b/accepted/0000-workspaces-config-management.md
@@ -1,0 +1,98 @@
+# npm workspaces: Config management
+
+## Summary
+
+Series of subcommands to allow managing what paths/folders should be handled as **workspaces**.
+
+## Motivation
+
+Minimize manual editing of the `package.json` file while providing more resilient and automated workflows for defining **workspaces** within your project.
+
+## Detailed Explanation
+
+- Add positional `workspaces` argument e.g: `lib/workspaces.js` that provides subcommands to enable workflows for **workspaces*** config management.
+
+## Rationale and Alternatives
+
+- The obvious alternative is to choose to not provide subcommands to handle **workspaces** config management.
+- Others? TBD
+
+## Implementation
+
+```
+# Given a structure:
+.
+├── package.json { "name": "foo" }
+└── packages
+    ├── dep-a
+    │   └── package.json { "name": "dep-a", "version": "1.0.0" }
+    └── dep-b
+        └── package.json { "name": "dep-b", "version": "1.3.1" }
+
+$ npm workspaces ls
+
+$ npm workspaces add ./packages/dep-a
+
+$ cat package.json
+{
+  "name": "foo",
+  "workspaces": [
+    "./packages/dep-a"
+  ]
+}
+
+$ npm workspaces ls
+dep-a@1.0.0 -> ./packages/dep-a
+
+$ npm workspaces add ./dep-b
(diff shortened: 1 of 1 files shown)


ljharb:
Per discussion on the call today, i have questions about how “add” works with a glob, or how “rm” works when the package to be removed is already included in a glob.

ruyadorno:
To elaborate a bit more the confusion comes from the fact that supporting globs can get confusing really fast, one of the examples mentioned during the weekly meeting was trying to add negative globs, e.g: `npm ws add !test` and how that can lead to hazardous results given different expectations from users.

So the tricky question is what is a behavior that would be reasonable and actually helpful for the majority of users with regards to `add/remove`.

`npm workspaces ls` on the other hand seems to have no contention at all and might be implemented regardless of the decision on `add` and/or `remove` in this RFC.

Christian24:
I am generally in favor of adding some commands to make working with workspaces easy. I generally like how `lerna` makes using workspaces easy, but we shouldn't attempt to rebuild `lerna`. 

* Maybe something like `workspace init` to start a guide which helps with initial setup would be cool?
* Git submodules might also be an interesting as inspiration: https://git-scm.com/book/en/v2/Git-Tools-Submodules
* Especially something like `foreach` might be nice, for things like `audit` etc. However, since this isn't config management feel free to tell me to shut up @ruyadorno  ;)

Christian24:
@ruyadorno you had mentioned on the call there have been talks with the team behind lerna, so it might be cool to add that as a kind of context. Additionally. a link to I think workspaces v2 is what it is called would be nice here too, so people can follow up with the previous discussion.

ruyadorno:
> @ruyadorno you had mentioned on the call there have been talks with the team behind lerna, so it might be cool to add that as a kind of context. Additionally. a link to I think workspaces v2 is what it is called would be nice here too, so people can follow up with the previous discussion.

@Christian24 I mentioned that in the context of the RFC for running commands: #117 we do have a bunch of the conversations documented there 😊

accepted/0000-workspaces-config-management.md
```diff
@@ -0,0 +1,98 @@
...
+├── package.json { "name": "foo" }
+└── packages
+    ├── dep-a
+    │   └── package.json { "name": "dep-a", "version": "1.0.0" }
+    └── dep-b
+        └── package.json { "name": "dep-b", "version": "1.3.1" }
+
+$ npm workspaces ls
+
+$ npm workspaces add ./packages/dep-a
```
jamiebuilds, line 34: Reading `npm workspaces add <package>` I would not expect it to mutate the root config. I would expect it to mirror `npm workspaces install tap` in #117 since `npm add/rm` are already aliases to install/remove dependencies.

I think `npm workspaces <subcommand>` can either be a command that operates on `package.json#workspaces` or every `<workspaces>/package.json` mixing and matching is confusing.

One thing we did in [Bolt](https://github.com/boltpkg/bolt) was have a `bolt project/p` command that allowed you to run commands on the root package containing all the workspaces.

Alternatives:

- `npm workspaces-config add/rm`
- `npm project workspaces add/rm`
- `npm workspaces add-workspace/rm-workspace`

HTML-CODDER (approved):
ok

HTML-CODDER (approved):
RFC: npm workspaces - Config management #273

PR #279, opened 2020-11-09
RFC for --default-command

isaacs:
[rendered](https://github.com/npm/rfcs/blob/isaacs/default-command/accepted/0000-default-command.md)

### DIFF
--- a/accepted/0000-default-command.md
+++ b/accepted/0000-default-command.md
@@ -1,0 +1,173 @@
+# Add `default-command` Config
+
+## Summary
+
+Create a config value `default-command`, which is the default command npm
+runs when the first positional argument does not match any known command.
+
+Default to `help`, to preserve the current behavior of npm.
+
+## Motivation
+
+If the root project package.json file contains a `"scripts"` section, then
+yarn will treat any defined script names as top-level commands, if they do
+not conflict with any known npm command.
+
+We frequently get requests from users to add this behavior, but since it
+would be a sigificant departure from previous behavior of the CLI, it
+usually gets shut down.
+
+However, we _do_ have some default behavior in npm.  When the first
+positional argument doesn't match a known command (or an alias or unique
+abbrev expansion), then the CLI calls the `help` command.  If the term does
+not match a known help topic, then `npm help` in turn calls `npm
+help-search`.  If that then finds a single matching help topic, it calls
+`npm help` with that help topic.
+
+It's very helpful!
+
+What if, instead of having the default command be hard-coded to `npm help`,
+we made it configurable, so that users could decide that they'd rather have
+something else (usually `run-script`) be invoked instead?
+
+## Detailed Explanation
+
+- Add a new config key, `default-command`.  This value must be a known npm
+  command (no aliases or abbreviations), and will have the default value of
+  `'help'`.
+- When the first positional argument does not match a known npm subcommand,
+  it will invoke the `default-command` option with the remainder of the
+  arguments.
+- When the first argument _does_ match a known npm subcommand, then the
+  `default-command` config value is not invoked.  (For example, `npm
+  --default-command=run-script install foo` will still install `foo` if
+  there is an `install` script in `package.json`.)
+- If a new command is added in the future, and this means that current
+  users of `npm foo` being an alias for `npm run-script foo`, then that's
+  an acceptable disruption.  New commands are added very rarely, and only
+  after a careful evaluation of existing userland packages.  But, this is a
+  power-user convenience feature, and we do not expect _many_ of our users
+  will take advantage of it (though we expect those that use it will be
+  happy to have it, and will understand the risks).
+- The `default-command` must only be a single command name.  It does not
+  support setting configurations or specifying additional arguments.  For
+  example, `npm config set default-command 'install --global'` would not be
+  allowed.
+
+## Rationale and Alternatives
+
+`npm help` is a good default for most users, especially new users.
+However, power users and those coming from Yarn may prefer to run their
+build scripts with `npm build` rather than `npm run build`.
+
+The rationale for being willing to clobber convenience use cases (eg,
+scripts named "install", etc.) when a new command is added is twofold.
+First, we are somewhat following in Yarn's footsteps, and Yarn only
+auto-runs scripts when the command is not known.  Second, we have to
+reserve the right to continue to expand the functionality of the CLI, or
+else development of npm is unreasonably limited.
+
+That said, it _does_ impose a non-zero cost in adding new commands to the
+CLI, since users _may_ be relying on a `--default-command` config to handle
+that term in some other way.
+
+### Alternative 1 - run-script _and then_ help
+
+One alternative would be to search for a match among known npm subcommands,
+then `run-script` options, and only _then_ fall back to searching help.
+
+This would be least surprising to current Yarn users, still pretty helpful,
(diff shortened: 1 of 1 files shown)


accepted/0000-default-command.md
```diff
@@ -0,0 +1,173 @@
...
+This would be least surprising to current Yarn users, still pretty helpful,
+and not require any opt-in.  However, it is not extensible to other command
+use cases, and still imposes the cost of adding new commands.
+
+### Alternative 2 - just make `help` know about scripts
+
+Another alternative would be to make the current default `help-search`
+behavior look at the `"scripts"` field in the current project's
+`package.json` file.  If the term matches there, it could suggest that the
+user might be thinking of the `run-script` command.
```
ljharb, line 88: this seems like a good improvement regardless, and would help the default/common case.

accepted/0000-default-command.md
```diff
@@ -0,0 +1,173 @@
...
+interesting use cases:
+
+- `npm config set default-command exec` This turns npm into something like
+  a more powerful `npx`.  You can do `npm install` to install your
+  dependencies, `npm create-react-app` to init a new react app, or `npm
+  rimraf folder` to remove a folder recursively.
+- `npm config set default-command install` Since `npm install` is the most
+  common use case for npm, why not just drop the `install`?  Run `npm` with
+  no args to install all your dependencies, or `npm react --save-peer` to
+  install react and save as a peer dependency.
```
ljharb, line 127: this seems like it would be really confusing for a non-advanced user who uses a project with default-command set in `.npmrc`.

Maybe it could provide some kind of output like:
```sh
> npm foo
Running `npm run-script foo`...
"Output of npm foo here"

> npm foo --silent
"Output of npm foo here"
```

isaacs: That's a good idea.  Could also probably want to have a `npm.log.notice('default command notice blah')` so that it automatically gets placed in the debug log when stuff goes sideways.

accepted/0000-default-command.md
isaacs: Yeah, it's somewhat orthogonal to this RFC, and something we could just do regardless.

isaacs:
Current status:

- "Just improve `help-search`" seems like a pretty strong contender which removes a lot of the hazards.
- Would really like to hear from some of the users who were asking for the default to be `run-script`.  Would they use this?
- Does anyone want the default to be something _other_ than `help` or `run-script`?
- Increasing difficulty/hazard of adding new commands is a pretty big obstacle.  If we move forward to this, we basically need to decide that we are ok with whatever breakage that may cause in the future.  Making any new command addition a semver-major version bump is not acceptable.

ljharb:
@isaacs i could see also wanting the default to be `exec`, but i wouldn't expect literally anything else besides help, run-script, or exec.

legodude17:
I could definitely see myself setting the default command to `install`, because that is the main thing I do with npm

styfle:
> Would really like to hear from some of the users who were asking for the default to be run-script. Would they use this?

As a yarn user and npm user, I would definitely use this.

However, I would take it a step further and change the default from `help` to `run-script` in the next major of npm.

I didn't even realize that `help` was the default because I usually append `--help` if I need to see more info.

dominguezcelada:
> > Would really like to hear from some of the users who were asking for the default to be run-script. Would they use this?
> 
> As a yarn user and npm user, I would definitely use this.
> 
> However, I would take it a step further and change the default from `help` to `run-script` in the next major of npm.
> 
> I didn't even realize that `help` was the default because I usually append `--help` if I need to see more info.

I agree on this!

dominguezcelada:
I think is not explicitly specified but I think it will be covered out of the box:
When enabling this feature, `npm run $command` will still be working right?

I think it would be a nice to have when if we consider a `.npmrc` shared inside the repo to all its users.

dominguezcelada:
> * Would really like to hear from some of the users who were asking for the default to be `run-script`.  Would they use this?
@wesbos

Jolg42:
Running scripts without specifying `run` would be amazing 💚

I know people including myself using `yarn run` just for that. Even if my project is managed by npm I still use yarn run to run my scripts locally 😅

wesbos:
So if I'm reading this correctly, if we set the `default-command` in `package.json` to `run-script`, it will make: `npm dev` and `npm styles` run `npm run dev` and `npm run styles` assuming that npm doesn't have a command with that name?

the opt-in is interesting. I like this proposal, but I'm a bit worried that people will wonder why `npm dev` works sometimes and other times it's `npm run dev`. Further adding to the frustration that it doesn't _just work_.   Not sure if that is warranted though..

isaacs:
> So if I'm reading this correctly, if we set the default-command in package.json to run-script

`s/package.json/.npmrc/`, but otherwise yeah.

The much bigger obstacle, honestly, is that it means `npm styles` or `npm dev` can never be a command that we do add, without it potentially being a breaking change for anyone relying on `npm styles` being a shorthand for `npm run styles`.  So, instead of adding a new command being a minor version update, it's SemVer major.

There's some support for adding a third bin called `npr` that'd be an alias for `npm run`.  Or, another safe opt-in way is that you put `alias npr=npm run` in your `~/.bashrc` file.

wesbos:
Makes sense, thanks! Most projects don't have an npmrc file and relying on a global config isnt ideal. I can't see people wanting to add another config file to their project to get a shortcut. As much as I want it. 

It's unpopular, but I think you should just let us use anything, and if you choose to release 'npm styles' for whatever reason, just warn that there is overlap and npm takes precedence

orta:
I think yarn has this right out of the door and npm should just copy its behavior, the npm CLI should look up:

 - npm command names
 - scripts in package.json
 - ./node_modules/.bin/

When looking up a command like `npm build` or `npm tsc`.

eps1lon:
What I always find confusing with npm: Given 3 `scripts` called `start`, `test`, `build`. `npm start` works the same as `npm run start`, `npm test` works the same as `npm run test`, `npm build` throws, `npm run build` works.

joshmanders:
I agree with the comments here of making it default. We're all pretty used to it already with `yarn` anyway. If I create a npm script called `bananas` and npm introduces a `bananas` command, that's on me, same as it is with yarn. 

Only thing I'd say is if released in v7, make it default to `help` but allow overriding with `npmrc` but then in v8 make it default to `run-script`.

ljharb:
imo it would be far too disruptive to make it a default; there are far more users (and docs/tutorials) used to npm's behavior than are used to yarn's.

@orta i believe in npm 7, `npx` won't install things without a prompt, so that concern should go away once it's marked "latest".

orta:
I don't think that should be a problem, all of the current npm behavior is accounted for in the yarn behavior described above to my knowledge. E.g. it won't change any current instructions.

The biggest issue is around the idea of a package script being broken by having an npm command taking priority, but that's a reasonable trade-off (and one discussed above a lot)

orta:
The pnpm folks just said they'd be open to the yarn pattern of command look-ups in their CLI.

I'm open to going through the npm RFC process for adding this behavior to npm as an alternative to this RFC, then we can get consistent behavior everywhere.

isaacs:
This is not a direction we're going to go.

- Making the addition of top-level commands a breaking change is a downside we are not willing to accept.
- Having `npm foo` behave differently on a per-project basis is also undesirable.

Closing this now, as the initial issue is better addressed by #339.

PR #281, opened 2020-11-10
NPM package release hashes stored on the blockchain

sean-lawless:
Initial RFC proposal for integrating NPM with the Immutable Ecosystem for additional blockchain security.

### DIFF
--- a/accepted/0000-blockchain-hash.md
+++ b/accepted/0000-blockchain-hash.md
@@ -1,0 +1,39 @@
+# Package release hashes stored on the blockchain
+
+## Summary
+
+A new technology now exists to store software release hashes on the immutable blockchain. Created by a nonprofit, the Immutable Ecosystem allows software creators to use their Ethereum wallet as their release signing key, managing their software releases through a free Dapp (https://ecosystem.immutablesoft.org). Supporting release hash lookup as well as reverse lookup, integration into NPM could take many forms and has the potential to simplify processes.
+
+## Motivation
+
+We believe this could provide additional security to the NPM ecosystem for discerning creators and end users.
+
+## Detailed Explanation
+
+Instead of using a centralized NPM database of release URI's and hashes that requires security monitoring and is subject to attacks, let us use the immutable blockchain. Specifically this RFC is to allow the OPTION to use the blockchain, throught the Immutable Ecosystem, as the authoritative source of package releases. Maybe this takes the form of an internal linkage between NPM and the Immutable Ecosystem (IE) for creators who indicate as such. A release step by package creators to push an IE release to NPM, or vice versa.
+
+## Rationale and Alternatives
+
+The alternative is the current centralized NPM database. Many other alternatives exists but to our knowledge these are all centralized systems prone to attack and abuse/neglect by their maintainers and/or corporate owners.
+
+As Software Creators it is time to take back our destiny. For too long have we been classified as vendors (Independent Software Vendors). We are NOT vendors, we are Creators who deserve and desire the right of self determination.
+
+## Implementation
+
+Through the Immutable Ecosystem (IM) a release hash and URI can be queried from the blockchain from the product, version and [optional] architecture. Or the URI, product, version, etc. can be queried from the blockchain from the release hash itself, providing a reverse lookup so to speak.
+
+There are various angles of implementation and I hope to complete this section after receiving feedback on which avenue is the best to pursue.
+
+{{THIS SECTION IS REQUIRED FOR RATIFICATION -- you can skip it if you don't know the technical details when first submitting the proposal, but it must be there before it's accepted}}
+
+## Prior Art
+
+{{This section is optional if there are no actual prior examples in other tools}}
+
+{{Discuss existing examples of this change in other tools, and how they've addressed various concerns discussed above, and what the effect of those decisions has been}}
+
+## Unresolved Questions and Bikeshedding
+
+{{Write about any arbitrary decisions that need to be made (syntax, colors, formatting, minor UX decisions), and any questions for the proposal that have not been answered.}}
+
+{{THIS SECTION SHOULD BE REMOVED BEFORE RATIFICATION}}
--- a/accepted/0000-blockchain-hash.md
+++ b/accepted/0000-blockchain-hash.md
@@ -1,8 +1,8 @@
-# Package release hashes stored on the blockchain
+# NPM package release hashes stored on the blockchain
 
 ## Summary
 
-A new technology now exists to store software release hashes on the immutable blockchain. Created by a nonprofit, the Immutable Ecosystem allows software creators to use their Ethereum wallet as their release signing key, managing their software releases through a free Dapp (https://ecosystem.immutablesoft.org). Supporting release hash lookup as well as reverse lookup, integration into NPM could take many forms and has the potential to simplify processes.
+A new technology now exists to store software release hashes on the immutable blockchain. Created by a nonprofit, the Immutable Ecosystem allows software creators to use their Ethereum wallet as their release signing key, managing their software releases through a free Dapp (https://ecosystem.immutablesoft.org). Supporting release hash lookup as well as reverse lookup, integration into NPM could take many forms and has the potential to secure and simplify these processes.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


ljharb:
Would the local npm client write to this blockchain? How would a user verify things? Why is a blockchain better than a checksum, like the one already stored in the packument, and also in local lockfiles?

sean-lawless:
It is the same checksum (sha256), the improvement is that the checksum is stored in the immutable blockchain and is immune to tampering. Forever. Period.

A creator, when they release a package, will write the new release URI, checksum, etc. to the blockchain, even using a hardware wallet as an HSM for release signing. The npm client would never write to the blockchain, only read; a Json request to an Infura node to query the Ethereum blockchain (for example). The NPM client install/update, etc. could verify that the hash stored on the blockchain for that package MATCHES the one stored in the local lockfile and/or NPM database (packument). The integration could even go completely inside NPM (to routinely verify the packument checksums themselves), without changing the end user interface or experience (speed).

If everyone is 100% certain the packument can never be hacked, or that man in the middle attacks upon npm clients are not possible, then this is likely a non starter.

ljharb:
Everything on npmjs.com is immune to tampering forever, period, also. If you don't trust npmjs.com, then a blockchain isn't going to help you; if you do, then it seems like it's not adding anything.

sean-lawless:
How is npmjs.com immune to tampering forever, period. Backups? A cloud provider agreement? Git Merkle roots stored on a hard drive/cloud?

I think there is still some confusion. This is not about trust in npmjs.com. I use and trust the NPM maintainers but am still deeply concerned about supply side attacks because they happen. Frequently enough to be a concern. As a blockchain developer it has happened to me during deployment of my smart contracts... a malicious package dependency lifted my wallet address during a live deployment (truffle still does not work well with hardware wallets). As an Etheruem developer, NPM is the weak link in our Dapp's security. With its web of dependencies it is impossible to do an accurate supply side audit. It is still not clear to me if this compromise came from an approved NPM package, or if my development computer was compromised and a malicious package installed in node_module from outside the NPM ecosystem. Both are possible and difficult to detect/audit. With a web search it is clear I am not the only one affected/concerned.

Only registered and approved contributors/maintainers will have their wallet address registered with ImmutableSoft and be able to release package updates. Package updates can ONLY be released if they are signed with the registered [hardware] wallet of the contributor/maintainer. This is one of the core security improvements of this proposal, securing supply side contributions, and leveraging that security at the client edge. The npm client using blockchain is only helpful if individual contributors/maintainers do so. Apologies if this was not clearly described in the RFC proposal.

Sure NPM will be providing info required to check the blockchain (package and version) but in this case the blockchain DOES help. With this change, a supply side attack would require that an attacker compromise a maintainers/contributors Ethereum hardware wallet (HSM) used to write/sign the package release URI/hash to the blockchain. Any attack on npmjs.com, or exposed login credentials of a maintainer, would not allow an attacker to introduce a vulnerable package.

ljharb:
Could you store checksums in a place you trust (in a blockchain, perhaps) on your own, and compare those to the checksums that are in your lockfile?

I think that making authors be required to "sign" their published packages would be a major change, not every author would adopt it (i doubt a majority of authors have even enabled 2FA), and a great many packages might never start signing releases, especially the ones that aren't updated anymore.

sean-lawless:
We could mirror and store the package checksums on the blockchain, accessible to the public. We have tried hard to solve this problem from that angle but it always resulted in a loss of security/trust. We did not create the package and are not in a position to verify it has not been tampered with. If we mirror NPM we inherit its insecurities, if any. It might help local clients to ensure their lock files have not been tampered with, but again, no better than checking the NPM server(s). A hybrid decentralized/centralized approach like this always inherits the worst of both worlds; additional complexity without reducing vulnerabilities.

Let's be clear, this proposal is for the OPTION for publishers to sign their packages. It might even be more valuable for this RFC to take a generic approach to this and allow ANY type of additional signing and not be specific to our solution (blockchain). Say with a X509 cert and CA chain like a PC application/TLS connection. I suspect some of your authors are familiar with and would be supportive of this.

NPM clients in this scenario (again optionally with a flag, etc.) would include an additional step to perform this secondary authentication check (and report packages without these signatures, fail and report to NPM/user if NPM authentication fails but NPM checksum does not). Even if only 2% of NPM clients and 20% of publishers enabled the feature it would help secure the entire NPM ecosystem (more than a hybrid approach) as these clients/packages will inform NPM, in real time, of any irregularities/attacks. If this secondary authentication of packages fails to discourage or identify attacks over time, we can dustbin it. If it does help identify active threats, as I think it will, we can encourage its use further.

I'd be happy to re-draft this in a more general form that is not exclusive to our blockchain solution. Is there any prior art or history on this?

iarna:
It would actually be very difficult to tamper with package contents of already published packages even with unrestricted access to npm services:

Because npm stores hashes in its lock files and refuses to install packages whose hash does not match the lockfile, any attempt to change packages that are in use would result in almost immediate build failures from CI systems. This kind of immediate detection, massively decreases the value of this sort of attack and renders it less likely.

That said, very difficult is not impossible.

I wonder if you would get the same benefit without the adoption hassle if the npm registry were the one to publish the hash to the blockchain with the client validated that post-publish? This would give you the same security guarantee: you'd know if npm posted a fake hash to start with, because you'd verify that immediately after publishing, and you'd know it was immune to future tampering because of the blockchain.

sean-lawless:
@iarna Thank you for your response. "I wonder if you would get the same benefit without the adoption hassle if the npm registry were the one to publish the hash to the blockchain with the client validating that post-publish?"

Yes, I agree this might be the ideal integration point. I am not familiar enough with NPM internals to know where to look to scope out the difficulty of this. It might be even a better solution to secure the NPM registry itself from tampering on the blockchain. Is the NPM registry a single file? How often is the registry updated/modified? If all package hashes are stored in the registry then all packages could be secured at once by securing the NPM registry. Client/NPM can then choose to ensure the validity of the registry file by checking on-chain before use. If the registry is validated on chain than all hashes stored in the registry can be assumed valid. It does not stop the problem of a creator poisoning am official package but it should mitigate the risk of compromised build systems and local lockfile attacks.

sean-lawless:
"Because npm stores hashes in its lock files and refuses to install packages whose hash does not match the lockfile, any attempt to change packages that are in use would result in almost immediate build failures from CI systems. This kind of immediate detection, massively decreases the value of this sort of attack and renders it less likely."

I am starting to realize the problem more. In an automated install/cloud deployment the lockfile prevents a malicious/unofficial package from being installed by a 3rd party URI during the installation process. This lockfile security is designed for the installation process.

But using NPM for Ethereum truffle development (or any other long term deployment), the node_modules/lockfile can become months/years old. If the install was 3rd party auditable against the blockchain, it would still not solve the problem of validating an existing installation 3 months later. Does 'npm-check' verify the installed packages (node_modules) against the lockfile or just check for updates? (ie. would it detect if node_modules was tampered with?). Does npm-check verify the lockfile against NPMs registry? How would it, the packages are since decompressed into subdirectories and files?

Securing NPM registry (or individual packages) on the blockchain would minimized the security dependency upon npm.com and is the focus of this thread. But I now understand that a post install verification might be even more helpful to myself and those working with a full node. Would a compress, hash and compare to lockfile be a feasible implementation of this and has anyone already considered this?

ljharb:
Isn't that what `npm ls` right after an `npm ci` already does?

sean-lawless:
@ljharb A clean install would be longer (downloads) but agree that it can be used to achieve the same goal.

Since the lockfile is a portion of the NPM registry written to flat file (correct me if I am wrong here), any NPM project could secure their application further by stamping their lockfile on-chain. A check before/after a clean install could ensure the lockfile remains identical to the original stamped on-chain, detecting scenarios of an unwanted update to the lockfile. Git might also be an effective way to detect this.

ljharb:
@sean-lawless if `~/.npm` were preserved, nothing would theoretically need to be downloaded more than once.

If you have a private npm registry (as you should regardless), and if you checksum the contents of the lockfile and package.json, i'm not sure what further guarantees would provide beyond that.

sean-lawless:
@ljharb Thanks for the help. I think storing application specific package-lock/package.json checksums on-chain (ie. stamping) is the best path uncovered so far and a good starting point. Optional to use and no direct integration required as it could be done pre or post install. A disadvantage is that any minor change to the package files (updates, etc.) would require stamping new file(s).

ljharb:
@sean-lawless you could always reduce the package.json down to the various dependencies fields only, and only include that subset in the checksum.

Any dependency updates *should* require stamping new files, since that's the whole point.

sean-lawless:
@ljharb I would still prefer to find a way to do this that everyone can leverage (not package file specific). I am not sure of the format but if a checksum of a private NPM registry was possible, and easily usable by anyone else with a private registry, it would be a much better solution (scales without additional blockchain writes ie. $).

iarna:
> Is the NPM registry a single file? 

The registry is a set of databases (specifics likely have changed since I knew them, to one degree or another) that communicates to the outside world via some preset APIs and the couchdb replication feed.

The database entries are individual JSON documents. (Example: [npm's](http://registry.npmjs.com/npm)) These JSON documents contain every version ever published and include a slightly modified copy of their `package.json` files. Couch is not a relational database, so there is no in-database linking between modules and their dependencies. All that's stored there is what's in the package.json: a name and a version specifier.

If you're thinking that the registry stores resolution trees like the lockfile: It does not. Nor is it a single file somewhere that has all of it the way there is in the older language ecosystems, ie, pypi, or cpan.

> How often is the registry updated/modified? 

Well, looking at the updates RSS feed, it's bursty. In the last ten minutes 9 modules updated, then 30 minutes before that 15 modules updated over 3 minutes. So looks like one every minute or two, but there can be brief periods of very high activity. It's also gonna correlate to hours that people are working around the world as general activity does.

> Does 'npm-check' verify the installed packages (node_modules) against the lockfile or just check for updates? (ie. would it detect if node_modules was tampered with?). 

npm install verifies hashes on every module before it installs them. If they don't match the lockfile it refuses to complete. This is also true for the other node package managers (ie the yarns and pnpm). This is why the lockfile contains the hash.

There is no facility for validating the modules installed on disk in npm. (Others may support this, for example, I suspect that yarn2's pnp mode may provide this guarantee to some degree?) The reason npm doesn't support this is because it's normal for for modules to do postinstall build steps, which results in files on disk that don't match the artifact. This is most obvious in the use of C dependencies, but is used for a wide variety of things. For instance, some modules (electron) download non-registry resources during their install.

If you're concerned that your node_modules on disk has been changed by an attacker, the best practice is: Delete the vm/container, make a new one, reinstall your project from git.

If you'd like to be alerted if they've changed there's already a bunch of more generic security tools that let you alert on changes to files in a directory.

brodybits:
> The database entries are individual JSON documents. (Example: [npm's](http://registry.npmjs.com/npm))

I would highly recommend that all users and tools use a secure link like this one: https://registry.npmjs.com/npm

iarna:
> I would highly recommend that all users and tools use a secure link like this one: https://registry.npmjs.com/npm

I mean yeah, that was just me being lazy and typing urls in straight into my browser. But I don't think that's otherwise relevant.

Issue #291, opened 2020-11-21
[RRFC] More permissive manifest format


TheJaredWilcurt:
## Motivation

JSON was designed to be a strict format for machines to send data to other machines. It's great for that, but has some issues when humans are expected to interact with it. Mainly it does not permit comments. As projects have grown larger and more complex, our `package.json` files have also grown. It is not uncommon for a Vue, Angular, or React project to have dozens of dependencies and npm scripts, or embedded configurations for tools like ESLint, Jest, etc. stored in the manifest.

This can make the maintenance of a project more difficult when it comes to how and were to document things related to the `package.json` file. Most commonly things around why dependencies were added, how dependencies are related, why a dependency is pinned to a specific version, and reasons for specific npm scripts. Currently `package.json` auto-alphabetizes dependencies, removing any intentional grouping of like-items (`babel-jest`, `jest`, `vue-jest` would logically be grouped and related to unit testing, but with a single `npm install` they are mixed in with the rest).

### Example

It would be helpful if there was a more permissive format allowed for writing a manifest file. I don't have any strong opinions on format or implementation. The main two I know of is JSON5 and HJSON. Of the two I would go with HJSON, just because it is more permissive, allowing users more freedom to write the files as they like. However JSON5 is much more popular. HJSON seems to allow anything YAML does, JSON does, anything in between, and more.

Title | Weekly npm installs | Github Stars | Age | Last release
:-- | --: | --: | :-- | :--
HJSON | 227,911 | 332 | 7 years | 3 days ago
JSON5 | 35,144,422 | 4,103 | 8 years | 8 months ago

Other systems like `node-config` handle [multiple formats](https://github.com/lorenwest/node-config/wiki/Configuration-Files#file-formats) (JSON, JSON5, HJSON, YML, CJS, Coffee, CSON, .properties, TOML, TS, XML). Though this is likely overkill for this.

#### Current Behaviour

Currently, only the very restrictive `.json` format is allowed for manifests. It does not allow manually organizing or documenting via comments. This results in each team having to invent their own solutions. Perhaps a github wiki page that documents all the dependencies, of course this page was created by someone that quit the company 2 years ago (me) and no one ever maintained. Or perhaps creating a "ManifestComments" field such as:

```json
{
  "name": "example",
  "version": "0.0.1",
  "scripts": {
    "start": "node index.js",
    "start-local": "node index.js --port=10000"
  },
  "ManifestComments": [
    "npm run start-local is used when you are running a local database from our backend repo",
[...]

PR #295, opened 2020-12-04
RFC: JSON5 and JSONL for manifest

Yash-Singh1:
# What / Why

RFC for `JSON5` and `JSONL` support. Rendered on: https://github.com/Yash-Singh1/rfcs/blob/patch-1/accepted/0029-read-json5-and-jsonl.md

## References

Relates to npm/cli#793

### DIFF
--- a/accepted/0029-read-json5-and-jsonl.md
+++ b/accepted/0029-read-json5-and-jsonl.md
@@ -1,0 +1,30 @@
+# Consider reading JSON5 and JSONL for package.json
+
+## Summary
+
+If a `package.json5` or a `package.jsonl` is present, then read that if `package.json` isn't there.
+
+## Motivation
+
+`JSON5` allows comments and multiple more features for `JSON`. `JSONL` isn't really required since the `manifest` isn't a list, but could be a little enhancement.
+
+## Detailed Explanation
+
+Read the `manifest` in the following order for files (important to least important):
+- `package.json`
+- `package.json5`
+- `package.jsonl`
+
+## Rationale and Alternatives
+
+- I could just use `package.json`
+- I could generate a `JSON` file from my `JSON5` file
+
+## Implementation
+
+1. Create a helper function for looking for the manifest in the order given above
+2. Implement this helper function in the commands and code
+
+## Prior Art
+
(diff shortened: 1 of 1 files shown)


ljharb:
See https://github.com/npm/rfcs/discussions/230, #291, for reasons why this is not a good idea.

ljharb:
also, what is “jsonL”? I’ve never heard of it.

Yash-Singh1:
Sorry for not elaborating on that. Look at JSONL's site: https://jsonlines.org/
It allows you to make an array easily. Since `package.json` is not an array, it can be parsed as JSON

Yash-Singh1:
Also, here are some programs that used JSONL: https://jsonlines.org/on_the_web/

ljharb:
That is a very short list when compared to the millions and millions of tools on npm today that parse, validate, *and write to* package.json.

For a format to be even remotely viable, i believe that the JavaScript language itself (or at least, node) would need the native ability to serialize and unserialize it. JSON, i believe, is the only reasonable candidate at the moment.

Yash-Singh1:
> That is a very short list when compared to the millions and millions of tools on npm today that parse, validate, _and write to_ package.json.
> 
> For a format to be even remotely viable, i believe that the JavaScript language itself (or at least, node) would need the native ability to serialize and unserialize it. JSON, i believe, is the only reasonable candidate at the moment.

I agree on your point

I stated later above:
> It allows you to make an array easily. Since `package.json` is not an array, it can be parsed as `JSON`
That means that the `package.jsonl` can be read as if it were `JSON`

This means that `npm` will just accept the `.jsonl` file extension as if it were a `.json` because the `manifest` is not an array.

ljharb:
So too would every ecosystem tool. When `package.json` and `package.jsonl` both exist, which one wins?

Yash-Singh1:
I think that question has an answer, JSON. I was just asking to add in JSONL, not to make JSONL win over JSON.

Yash-Singh1:
> So too would every ecosystem tool. When `package.json` and `package.jsonl` both exist, which one wins?

👍🏿 Agreed, it probably won't be used way too much.

PR #298, opened 2021-01-04
Add script-file support in `npm run`

Schniz:
Hi! that's my first RFC. Please let me know if I'm missing something, or if I should close it 😄

I believe that adding this ability would make it easy to manage big Node.js projects.

Thanks for reading!

### DIFF
--- a/accepted/0000-support-script-files-in-npm-run.md
+++ b/accepted/0000-support-script-files-in-npm-run.md
@@ -1,0 +1,53 @@
+# Script file fallback in `npm run`
+
+## Summary
+
+`package.json` scripts are very handy. We can rely them to create industry standards like `npm run build` and `npm run test` to build and test files respectively.
+This RFC suggests a fallback for script files, instead of explicitly listing them in `scripts`.
+
+## Motivation
+
+In big Node.js repos, the `package.json#scripts` JSON key might get very crowded. It's widely common to see `test:e2e`, `test:unit`, `test:integration` and many more things.
+The problem is that declaring scripts in JSON is hard to maintain and to reason about:
+
+- Any logic you put into the scripts is hard to manage, because they're practically one-liners
+- You can't use prettier/shellcheck/TypeScript or any other tool to check your syntax and enforce standards
+- If you do want to extract your script into a standalone script file, you will still need to have an explicit reference to the file from `package.json#scripts`
+
+npm could suggest a standard location to write script files so they will be picked up automatically instead of explicitly referencing them in `package.json#scripts`.
+
+## Detailed Explanation
+
+Practically, if `package.json#scripts#SCRIPT_NAME` is missing, and `scripts/SCRIPT_NAME` exists and executable, we should run this script.
+
+This will make it easier to use tools like [`argv`](https://www.npmjs.com/package/yargs), [`commander`](https://www.npmjs.com/package/commander), or [`cmd-ts`](https://www.npmjs.com/package/cmd-ts) to create scripts that receive arguments with proper validation instead of overloading scripts.
+So, instead of `npm run test:e2e` you could run `npm run test e2e` — and declare it in your favorite language that supports reusing code, instead of a Bash one-liner.
+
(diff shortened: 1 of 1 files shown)


accepted/0000-support-script-files-in-npm-run.md
```diff
@@ -0,0 +1,53 @@
...
+## Motivation
+
+In big Node.js repos, the `package.json#scripts` JSON key might get very crowded. It's widely common to see `test:e2e`, `test:unit`, `test:integration` and many more things.
+The problem is that declaring scripts in JSON is hard to maintain and to reason about:
+
+- Any logic you put into the scripts is hard to manage, because they're practically one-liners
+- You can't use prettier/shellcheck/TypeScript or any other tool to check your syntax and enforce standards
+- If you do want to extract your script into a standalone script file, you will still need to have an explicit reference to the file from `package.json#scripts`
+
+npm could suggest a standard location to write script files so they will be picked up automatically instead of explicitly referencing them in `package.json#scripts`.
```
ljharb, line 17: The problems you list seem to be solved by an explicit call out to a separate script file - which means this RFC is only solving the problem of "explicitly adding a `scripts` entry"?

accepted/0000-support-script-files-in-npm-run.md
```diff
@@ -0,0 +1,53 @@
...
+Practically, if `package.json#scripts#SCRIPT_NAME` is missing, and `scripts/SCRIPT_NAME` exists and executable, we should run this script.
+
+This will make it easier to use tools like [`argv`](https://www.npmjs.com/package/yargs), [`commander`](https://www.npmjs.com/package/commander), or [`cmd-ts`](https://www.npmjs.com/package/cmd-ts) to create scripts that receive arguments with proper validation instead of overloading scripts.
+So, instead of `npm run test:e2e` you could run `npm run test e2e` — and declare it in your favorite language that supports reusing code, instead of a Bash one-liner.
+
+We can also support `namespaced:script`, which uses a colon to separate the calls — a standard we probably got from Rake, and look for a file in `scripts/namespaced/script`.
+
+## Rationale and Alternatives
+
+- [scripty](https://www.npmjs.com/package/scripty) is a package that does exactly that. But you still need to explictly declare your scripts in `package.json#scripts` to all call `scripty`, which be better avoided
```
ljharb, line 30: why is this better to avoid? It seems to me that it would be *clearer*, because there's no magic implicitness that someone would need to somehow know.

Schniz:
You make a very good case @ljharb. This is true and maybe this RFC doesn’t really solve a real issue. 

I regularly see big projects with 50+ scripts. They’re getting out of hand and lack standards, call each other, have inline regular expressions or do crazy things.

I guess what I’m looking for is for a standard way to declare tasks, instead of one liners, while avoiding having magical tasks that appear out of nowhere.

 I know that this RFC only suggests a really small change but it’s a convention that can potentially make a big impact on maintainability :)

ljharb:
Can you help me understand how a folder full of 50 files, that appear out of nowhere in `npm run` (because they're not explicitly added anywhere), that could have inline regular expressions or do crazy things, is more maintainable than the same thing in an explicit `scripts` block?

Schniz:
> could have inline regular expressions or do crazy things, is more maintainable than the same thing in an explicit `scripts` block?

They _can_ have inline regular expressions. In npm scripts all you can do is inline — unless you call an external script.

* A script can be written in any language, so I guess most people won't turn to Bash to do that
* So they can use JS, TS, or even Bash with syntax highlighting
* They can use a linter/formatter

But that's fine we disagree, closing the RFC.

ljharb:
To be clear, RFC decisions aren't up to me, and closing it might be premature.

I'm more trying to understand, given that users already can put things in external scripts for the benefits you describe, what this RFC offers. It seems like the only difference is that a user would be able to skip writing `"foo": "./scripts/foo"` in the scripts block, with the tradeoff that another user wouldn't necessarily realize that `npm run foo` magically ran `./scripts/foo` after setting up some environment variables. What am I missing?

PR #317, opened 2021-02-09
RFC: Publish set the tag accordingly to the semver version number

Divlo:
Currently : Versions numbers have no direct connection with dist-tags.
It should be possible to run `npm publish` without specifying the `--tag` argument, instead it will pick the tag based on the semver version number (inside `package.json`).

Examples of various versions values inside `package.json`

`16.0.3` release the package attached to the `latest` tag
`16.0.3-tagname.1` release the package attached to the `tagname` tag
`16.0.3-beta.1` release the package attached to the `beta` tag

## References

- Initial discussion in `npm/feedback` : <https://github.com/npm/feedback/discussions/194>
- Issue on `yarnpkg` about the same feature : <https://github.com/yarnpkg/berry/issues/2444>
- Mutability of the `latest` tag discussion : <https://github.com/npm/feedback/discussions/109>

### DIFF
--- a/accepted/0000-publish-set-tag-accordingly-to-semver-version.md
+++ b/accepted/0000-publish-set-tag-accordingly-to-semver-version.md
@@ -1,0 +1,45 @@
+# Publish set the tag accordingly to the semver version number
+
+## Summary
+
+Currently : Versions numbers have no direct connection with dist-tags.
+It should be possible to run `npm publish` without specifying the `--tag` argument, instead it will pick the tag based on the semver version number (inside `package.json`).
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


ljharb:
An alternative might be a new config value that opts in to this behavior, that then the project could put in its `.npmrc`  - then there's no surprises for existing authors, but there's an ergonomic path to this behavior for those that know about it.

wesleytodd:
Just to document another thing I hope we could change related to dist-tags.  If `--save-exact` was used when installing a `dist-tag` specifier, it would be great if that saved the exact value.  So `npm i --save-exact foo@next` would result in `"foo": "next"` in the `package.json`.

PR #319, opened 2021-02-10
feat(multiple-dist-tags): add proposal for multiple dist-tags

wraithgar:
This is a suggestion that grew out of a discussion surrounding https://github.com/npm/rfcs/pull/317.

It will afford a way for end-users to set all the dist-tags they want at once, at the time they intend to.

Examples:

```sh
$ npm publish --tag latest,next,latest-7
```

```
$ npm dist-tag add npm@6.14.11 latest-6,lts
```

```sh
$ npm dist-tag rm npm next-2,next-3,next-4
```

### DIFF
--- a/accepted/0000-multiple-dist-tags.md
+++ b/accepted/0000-multiple-dist-tags.md
@@ -26,7 +26,7 @@ comma-separated list of tags.
 ## Rationale and Alternatives
 
 One approach is to have postinstall scripts that handle this. This is
-what some maintaners are doing now and it is not very practical because
+what some maintainers are doing now and it is not very practical because
 it is simply working around the limitations of the existing system only
 allowing one tag at a time.  Adding lifecycle methods is less preferable
 to being able to specify the tags all at once when you would be changing
--- a/accepted/0000-multiple-dist-tags.md
+++ b/accepted/0000-multiple-dist-tags.md
@@ -1,0 +1,94 @@
+# Allow for multiple dist-tags in a single command
+
+## Summary
+
+When publishing, and when adding and removing dist-tags, the user should
+be able to specify multiple dist-tags
+
+## Motivation
+
+Allowing for multple tags to be set on a given version at once will
+streamline many users' workflows.  Currently they must add tags
+one-at-a-time to any new versions they published, and can only set one
+of those tags during the publish action itself.
+
+## Detailed Explanation
+
+Both `npm publish` and `npm dist-tag` will be affected.
+
+The `--tag` parameter for `npm-publish` will parse as an optionally
+comma-separated list of tags.  Since a comma is not a valid character
+currently in tags, this does not present a breaking change.
+
+The `add` and `rm` commands would parse their `<tag>` parameter as a
+comma-separated list of tags.
+
+## Rationale and Alternatives
+
+One approach is to have postinstall scripts that handle this. This is
+what some maintaners are doing now and it is not very practical because
+it is simply working around the limitations of the existing system only
+allowing one tag at a time.  Adding lifecycle methods is less preferable
+to being able to specify the tags all at once when you would be changing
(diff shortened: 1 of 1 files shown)


accepted/0000-multiple-dist-tags.md
```diff
@@ -0,0 +1,94 @@
...
+comma-separated list of tags.  Since a comma is not a valid character
+currently in tags, this does not present a breaking change.
+
+The `add` and `rm` commands would parse their `<tag>` parameter as a
+comma-separated list of tags.
+
+## Rationale and Alternatives
+
+One approach is to have postinstall scripts that handle this. This is
+what some maintaners are doing now and it is not very practical because
```
ljharb, line 29:
```suggestion
what some maintainers are doing now and it is not very practical because
```

ljharb (approved):
Will versions of npm without this feature fail loudly on `,` in the argument? (i believe so)

wraithgar:
> Will versions of npm without this feature fail loudly on `,` in the argument? (i believe so)

Good question, no it won't.  I was under the impression that dist-tags could not contain values that would url encode but apparently I am wrong.   This means we have to either choose the new cli flag approach, or find a syntax that *will* error out with the current cli, and hopefully current registries.

```sh
~/D/n/gar-create-test (main↑4|✔) $ npm dist-tag add @gar/create-test@1.0.0 foo,bar
This operation requires a one-time password.
Enter OTP: ******
+foo,bar: @gar/create-test@1.0.0
~/D/n/gar-create-test (main↑4|✔) $ npm dist-tag ls @gar/create-test
false: 1.0.11
foo,bar: 1.0.0
latest: 1.0.13
```

wraithgar:
Multiple values for the same parameter might be better implemented as multiple declarations on the cli i.e.

```sh
npm publish --tag latest --tag lts --tag next
```

```sh
npm tag rm npm next-2 next-3 next-4
```

ljharb:
What happens in npm already with multiple tags in those fashions?

wraithgar:
> What happens in npm already with multiple tags in those fashions?

It takes the last one you gave

```sh
~/D/n/gar-create-test (main↑5|✔) $ npm publish --tag latest --tag next --tag false
...
~/D/n/gar-create-test (main↑5|✔) $ npm dist-tag ls @gar/create-test
false: 1.0.15
latest: 1.0.13
```

wesleytodd:
I have packages with all sorts of weird tags because of accidental typos or bugs in the generated tags.  For example `"latest"` where an extra set of quotes was thrown in on accident.  I wonder if some normalization could be done which would prevent further changes from being this difficult?  If we did a slugification on tag strings I think it would help prevent bugs/errors and result in a better overall UX despite the fact that it might break some use cases in the wild.

Divlo:
In my opinion, it is weird to have the same version with mutiple tags attached to it, version is either stable (`latest`) or in `beta`, or in `alpha`, etc.
I would rather have a better deal with version number and tag, as it is described in #317.
I think that tags and versions number should work together, better relationship between them.

Most of npm packages are using tags (others that the default `latest`) to get feedback from a small part of users, before releasing for the majority of users, so they can improve the release before publishing it as ready for production in `latest`.
Is there is any use case for tags, that I'm missing ?

As far as I understand to the issue, we can already publish a version to mutiple tags by doing multiple time the command ?
For example, currently, I want to release my new version to `latest`,  ` next` and `canary`, I need to do it like so : 
```sh
npm publish --tag latest
npm publish --tag next
npm publish --tag canary
```
Is it right ?

As I said, I don't think it's a good idea to publish a version to mutiple tags, but I agree that it's better to do it like that : `npm publish --tag latest --tag next --tag canary` than the current way to do it.

bnb (approved):

PR #321, opened 2021-02-11
feat(no-tag-publish): add proposal for a no-tag publish

wraithgar:
This is a suggestion that grew out of a discussion surrounding https://github.com/npm/rfcs/pull/317.

It will afford a way for end-users to publish new versions of packages without having to set a `dist-tag` for the version being published.

The name of the parameter is still tbd in the `Bikeshedding` section.

```sh
npm publish --skip-tag
```

## References
Related to https://github.com/npm/rfcs/pull/319
Related to https://github.com/npm/rfcs/pull/317

### DIFF
--- a/accepted/0000-no-tag-publish.md
+++ b/accepted/0000-no-tag-publish.md
@@ -1,0 +1,60 @@
+Publishing a package without a dist-tag
+
+## Summary
+
+When publishing, the user should be able to specify that no new
+dist-tags should be included with the release.
+
+## Motivation
+
+Allowing for the behavior of various registries to deal with defaul
+tags, and allowing users to publish legacy versions of their packages
+without having to either create new tags for legacy lanes, or juggle the
+`latest` tag after the fact.
+
+## Detailed Explanation
+
+This will only affect `npm publish`
+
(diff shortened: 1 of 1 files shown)


ljharb:
`--no-tag` still seems like the best option; i doubt anyone's relying on that producing a `false` dist-tag (are there any packages with one in the registry?)

wraithgar:
> `--no-tag` still seems like the best option; i doubt anyone's relying on that producing a `false` dist-tag (are there any packages with one in the registry?)

We can't audit private packages for this anyways.

ljharb:
Surely private packages on the public registry can be audited for this?

wesleytodd:
Even if they are, I would consider that behavior a bug not a feature.  If they want to have a string value of `false` they should be using `--tag="false"` IMO.

ljharb:
@wraithgar this is still a very valuable feature, can this be reopened?

wraithgar:
The npm registry itself doesn't support it.

wraithgar:
Also this is in line with our new rfc approach that it's something someone owns and is actively going to implement.

ljharb:
and there's no opportunity to get it onto the registry's roadmap?

PR #332, opened 2021-02-26
RFC:  Support Yarn style command/script/bin lookups from the CLI with npm prefix

orta:
## References

Alternative to #279

### DIFF
--- a/accepted/0000-missing-command-lookups.md
+++ b/accepted/0000-missing-command-lookups.md
@@ -1,0 +1,70 @@
+# Support Yarn style command/script/bin lookups from the CLI
+
+## Summary
+
+Allow for a CLI experience like:
+
+```sh
+npm install typescript
+npm tsc
+```
+
+## Motivation
+
+I have two main motivations:
+
+- Personal, looking at the list of yarn commands I [use in an average month](https://gist.github.com/orta/489dbfa1b7467804955ea7800adf1494#file-last_month_yarn-txt) (1200) roughly 30% (400) of those commands are using the shorthand syntax to run a command installed into `node_modules/.bin` in a project.
+
+- Professional, it'd be great for the TypeScript docs to be able to simply say: `npm install typescript --save-dev' then `npm tsc index.ts` and there to be no ambiguity like our current recommendation for `npx tsc index.ts`.
+
+## Detailed Explanation
+
+When npm receives a command it does not recognise, npm first resolves the next string to see if it matches:
+
+ - Available "scripts" in `package.json`
+ - Commands matching in `./node_modules/.bin/`
+
+Failing that, then show help.
+
+## Rationale and Alternatives
+
+1. Having a `script`s section like this:
+
+    ```json
+    {
+      "scripts": {
+        "tsc": "tsc",
+        "eslint": "eslint",
+        "prettier": "prettier"
+      }
+    }
+    ```
+
+    Is a waste of space in the JSON, the `node_modules/.bin` folder is typically full of commands and after a while you either start prefixing commands with `node_modules/.bin` instead or you fill up the scripts section.
+
+2. This change removes the concept of "blessed" scripts. `npm test` works but `npm test:lint` does not. Eventually folks memorize the ones they can use shorthand for, but there's no reason for the distinction other than at some point it was decided to be given a short-cut. These special cases can be removed with this change.
+
+3. It does not break any existing setups. 
+ 
+ - `npm tsc` does not work today, anyone who is doing that is seeing a help message.
+ - `npm test:lint` also does not work today.
+
+
+Alternative to:
+
+- [RFC 279:for --default-command](https://github.com/npm/rfcs/pull/279) - tries to hit the same issue, but because it'd a user option it wouldn't be useful at the eco-system level and only people who know it exists would use it.
+
+## Implementation
+
+I'm not sure on why so many npm commands on npm require the `--`.
+It feels like an internal abstraction (given that it had to be added to `npm exec` in npm 7) but I'd still advocate for it's removal here. `npm -- tsc` is a strange command to write. 
+
+## Prior Art
+
+Yarn does this, Berry does this and PNPM has indicated they're open to supporting it too.
+
+## Unresolved Questions and Bikeshedding
+
+I assume npm uses `--` to match the [unix format](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean) but I'd still advocate for it's removal here. `npm -- tsc` is a strange command to write, and the intention is quite obvious (given that npx also went with the same decision)
+
+I assume there are flags to npm which you may want to be able to include e.g. `npm --registry-something=mything.com deploy`. Which would call `npm deploy` after those options are set. I  assume that the npm client knows all of these commands ahead of time, and can either remove those before looking up the script/bin, or we can assume that any string beginning with `-` won't be a script/binary.
--- a/accepted/0000-missing-command-lookups.md
+++ b/accepted/0000-missing-command-lookups.md
@@ -13,7 +13,7 @@ npm tsc
 
 I have two main motivations:
 
-- Personal, looking at the list of yarn commands I [use in an average month](https://gist.github.com/orta/489dbfa1b7467804955ea7800adf1494#file-last_month_yarn-txt) (1200) roughly 30% (400) of those commands are using the shorthand syntax to run a command installed into `node_modules/.bin` in a project.
+- Personal, looking at the list of yarn commands I [use in an average month](https://gist.github.com/orta/489dbfa1b7467804955ea7800adf1494#file-last_month_yarn-txt) (1200) roughly 30% (400 (^1)) of those commands are using the shorthand syntax to run a command installed into `node_modules/.bin` in a project.
 
 - Professional, it'd be great for the TypeScript docs to be able to simply say: `npm install typescript --save-dev' then `npm tsc index.ts` and there to be no ambiguity like our current recommendation for `npx tsc index.ts`.
 
@@ -56,8 +56,7 @@ Alternative to:
 
 ## Implementation
 
-I'm not sure on why so many npm commands on npm require the `--`.
-It feels like an internal abstraction (given that it had to be added to `npm exec` in npm 7) but I'd still advocate for it's removal here. `npm -- tsc` is a strange command to write. 
+I think the biggest question is the `--` below.
 
 ## Prior Art
 
@@ -68,3 +67,7 @@ Yarn does this, Berry does this and PNPM has indicated they're open to supportin
 I assume npm uses `--` to match the [unix format](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean) but I'd still advocate for it's removal here. `npm -- tsc` is a strange command to write, and the intention is quite obvious (given that npx also went with the same decision)
 
 I assume there are flags to npm which you may want to be able to include e.g. `npm --registry-something=mything.com deploy`. Which would call `npm deploy` after those options are set. I  assume that the npm client knows all of these commands ahead of time, and can either remove those before looking up the script/bin, or we can assume that any string beginning with `-` won't be a script/binary.
+
+---
+
+^1 - I basically just removed any traditional command, then slowly dropped anything which idn't look like it used a script or binary. Then rounded down to =assume there would be a bunch of exceptions.


accepted/0000-missing-command-lookups.md
```diff
@@ -0,0 +1,70 @@
...
+npm tsc
+```
+
+## Motivation
+
+I have two main motivations:
+
+- Personal, looking at the list of yarn commands I [use in an average month](https://gist.github.com/orta/489dbfa1b7467804955ea7800adf1494#file-last_month_yarn-txt) (1200) roughly 30% (400) of those commands are using the shorthand syntax to run a command installed into `node_modules/.bin` in a project.
+
+- Professional, it'd be great for the TypeScript docs to be able to simply say: `npm install typescript --save-dev' then `npm tsc index.ts` and there to be no ambiguity like our current recommendation for `npx tsc index.ts`.
```
ljharb, line 18: if you suggest they install it first, what ambiguity is there in `npx tsc index.js`?

accepted/0000-missing-command-lookups.md
```diff
@@ -0,0 +1,70 @@
...
+        "prettier": "prettier"
+      }
+    }
+    ```
+
+    Is a waste of space in the JSON, the `node_modules/.bin` folder is typically full of commands and after a while you either start prefixing commands with `node_modules/.bin` instead or you fill up the scripts section.
+
+2. This change removes the concept of "blessed" scripts. `npm test` works but `npm test:lint` does not. Eventually folks memorize the ones they can use shorthand for, but there's no reason for the distinction other than at some point it was decided to be given a short-cut. These special cases can be removed with this change.
+
+3. It does not break any existing setups.
```
ljharb, line 47: if npm added an "npm build" command the release after shipping this RFC, though, it would break a great many existing setups.

accepted/0000-missing-command-lookups.md
```diff
@@ -0,0 +1,70 @@
...
+ - `npm test:lint` also does not work today.
+
+
+Alternative to:
+
+- [RFC 279:for --default-command](https://github.com/npm/rfcs/pull/279) - tries to hit the same issue, but because it'd a user option it wouldn't be useful at the eco-system level and only people who know it exists would use it.
+
+## Implementation
+
+I'm not sure on why so many npm commands on npm require the `--`.
```
ljharb, line 59: this is a normal unix convention to tell one command to stop processing `--` commands and pass the rest on untouched; imo people are unfamiliar with it largely because a few popular tools have opted to try to hide that convention from users.

accepted/0000-missing-command-lookups.md
orta: I keep getting feedback because people assume "npx = remote", and not "local then remote", I get that they are not right, but I also thought that until I researched what we should recommend.

ljharb: in npm 7, npx will prompt before auto-installing, so as npm 7 propagates, more people hopefully will have their expectations corrected.

accepted/0000-missing-command-lookups.md
orta: Heh, yeah, this bit was my notes before I wrote below - dropped it in the current version

accepted/0000-missing-command-lookups.md
orta: This is a possibility yes, at that point there would probably be a major npm semver and breaking changes like that can be handled.

accepted/0000-missing-command-lookups.md
```diff
@@ -0,0 +1,73 @@
...
+
+3. It does not break any existing setups. 
+ 
+ - `npm tsc` does not work today, anyone who is doing that is seeing a help message.
+ - `npm test:lint` also does not work today.
+
+
+Alternative to:
+
+- [RFC 279:for --default-command](https://github.com/npm/rfcs/pull/279) - tries to hit the same issue, but because it'd a user option it wouldn't be useful at the eco-system level and only people who know it exists would use it.
```
ruyadorno, line 55: The way I see it we can still have the option proposed in the `default-command` RFC. It would allow us to ship the ability to swap your `default-command` in npm7 and when npm8 comes we can simply turn the default value to `default-command=run-script`

orta: Yeah, that could definitely work if `run-script` was expanded to also include binaries 👍🏻

accepted/0000-missing-command-lookups.md
ljharb: Right - but the consequence is that npm could no longer add top-level commands without it being a semver-major change; and regardless of semver-major _allowing_ them to do it, that doesn't change the disruption it might cause for users.

orta: IMO that's worth it. This behavior accounts for roughly 40% of my command line usage in yarn. I don't maintain npm, so that's just a anecdata but this feature has been a constant win for yarn from day 1. Even post npm 7 and in yarn deprecated land, I regularly recommend people to not use npm just for this alone.

There's possible mitigation strategies like warnings when they match a potential upcoming command, or just outright allowing the script to take precedence over _new_ commands. npm could emit that this is blocking a known CLI command and recommend a rename while still running the same script. I am assuming under the hood a new `npm publish2` isn't directly calling out to the shell with `npm pack2` for example, and so adding a script `pack2` wouldn't take precedence in the internals and people's projects wouldn't break.

accepted/0000-missing-command-lookups.md
eyelidlessness: I also only learned recently that `npx` prefers local resolution. I’d avoided using `npx` from the start because it makes me super nervous to run remote code (I’m also “that guy” who warns people not to pipe to `curl` even if you read the source because it can be detected server side).

Now knowing about local resolution, I’ll use `npx` in projects where I don’t choose the package manager, but only in the `scripts` section where I’m prompted to also verify there’s a local dependency. And even then it makes me super nervous that the dependency will be removed and introduce a silent vulnerability. So nervous that anytime I go to use an `npx` command and I know my git `HEAD` has changed I **try to remember** to check the `package.json`, and I just know someday I’m gonna forget. Not a great UX! (I’m sure some combination of `npm` flags and a `grep` and an `alias` could make this safe for my preferences, but then I’m moving that risk to anyone I forget to warn!)

IMO requiring a local install, whatever the ultimate API, is an important feature for security and repeatability. And those are worth the distinction of a new API, even if TIMTOWTDI.

accepted/0000-missing-command-lookups.md
ruyadorno: I see, that would be a diff command then!

ljharb: It seems like pursuing a new command that was "`exec` without any option to install" and "fall back to `run-script`", and then seeing how useful that command is when set to the `default-command`, would be a viable and appropriately cautious path forward here?

isaacs:
I agree that this behavior is convenient, and I like convenience.  That said, there are a few footguns and drawbacks that we identified in the `--default-command` RFC which would have to be addressed to move forward on this:

First, it makes the addition of any new command semver-major.  We could either decide that this is fine, effectively locking the set of top-level commands for v7, or we could keep track of which commands were introduced since this change, and make _those_ commands prefer the script/bin with a warning, and switch them over in the next major.  That's a complexity cost, but not a huge one.  The bigger impact of this is that we will have to be careful to introduce new commands even in a semver-major release, because breaking changes tend to be obstacles to users upgrading, and we work very hard to make upgrades as easy as possible.

Second, running scripts is one thing, but running any binary in `node_modules/.bin` is a bit more risky.  For example, let's say that we add a new `npm foo` command, but mark it as "still preferring the script" for now.  Then, in a semver-minor release, one of your dependencies introduces a `foo` binary, breaking your build by taking over the top-level npm command.  In this case, a user has done nothing but reasonable semver-minor upgrades, and find themselves in a semver-major situation unexpectedly.  So that's not good.

I think that would mean that not _only_ do we have to search for existing scripts on the registry, but _also_ for existing bins on the registry, to ensure that we're not breaking anyone by adding a new command, and even there, we can't be sure who's using the shorthand and who isn't.  So any command addition is effectively semver-major no matter what, and potentially a build-breaking addition that we will have a hard time predicting.  (We'd probably want to back-port a list of new commands, so that npm v7 could warn about those if it sees them in the package's `bin` or `scripts` blocks.)

I would be totally fine with adding a `npm exec-local` command which does not install anything and only looks in the local `node_modules/.bin` contexts.  That would be very easy.  Fun shorthand, too `npm xl` ;)

The resolution of the `--default-command` RFC discussion was to upgrade `help-search` to _also_ search bins and scripts, and suggest the appropriate command to run.

eyelidlessness:
With so much focus on the impact of new commands in semver-minor versions, it seems like a corresponding RFC to semver addressing the fact that new commands _are_ breaking changes may be warranted. This isn’t a theoretical problem only NPM has to consider, the behavior referenced has existed in Yarn for a long time. Updating semver to recognize the issue wouldn’t necessarily prevent it, but it would certainly raise awareness.

ljharb:
New commands are absolutely not breaking changes in the common case, so that would be an incorrect RFC. It is a theoretical problem for npm with this rfc, and is a long-standing real problem for yarn, but most projects haven’t made the choice yarn has, and as such, for most projects new commands are semver-minor.

eyelidlessness:
> New commands are absolutely not breaking changes in the common case

I used to feel the same way, but I was already turning a corner, and this discussion has slammed the needle to 11 for me. In the common case, iterating over all possible command/outcome cases should be considered unknown. If additive changes are considered non-breaking, then a list of known commands is considered non-exhaustive. Anything built around the tooling or documentation could be considered unpredictable for unknown commands at best.

This is all probably a stronger argument against Yarn-style behavior, but it’s so common that people already safeguard against it.

In a semver scenario the equivalent is breaking any `Object.keys`/comparable expectations. This is why TypeScript can’t type that method. Adding new functionality seems safe until you consider how people depend on the scope of existing functionality.

That objection either means redirecting this RFC to different semantics or accepting that new commands are, in fact, breaking changes.

ljharb:
It is indeed a strong argument against yarn-style behavior; and i see no evidence anyone is protecting against it.

Additive changes are not breaking, full stop. If you’re doing reflection and relying on the absence of a property, you’re voiding the warranty.

eyelidlessness:
That’s a somewhat surprising response. So far the strongest resistance to the behavior has been the risk of additive behavior breaking existing workflows, a tacit acknowledgment that additive changes are in fact breaking changes.

The claim no one is protecting against it seems to suggest that either Yarn, having long had the behavior, has either experienced such breakage, or has just had dumb luck not experiencing it. For my usage I’ve avoided script names that would be ambiguous. I’d even do so without Yarn’s behavior because ambiguity is a mental tax. I think this is common not just in Yarn usage but in NPM/PNPM usage as well.

ljharb:
@eyelidlessness adding new top-level commands to a CLI is safe in general; it isn’t safe for yarn, and wouldn’t be safe for npm if this RFC ships as-is. I assume yarn has either chosen to not add new top-level commands, or, has chosen to endure the breakage.

Any word that is a likely new top-level npm or yarn command is one I’d wager is a commonly used script name - “build”, for example. This RFC changes the decision to add a new command from “what’s a good name?” to “let’s exhaust our resources to figure out if the best names are already used, and even then it’ll probably break an enterprise codebase”.

accepted/0000-missing-command-lookups.md
```diff
@@ -0,0 +1,73 @@
...
+npm tsc
+```
+
+## Motivation
+
+I have two main motivations:
+
+- Personal, looking at the list of yarn commands I [use in an average month](https://gist.github.com/orta/489dbfa1b7467804955ea7800adf1494#file-last_month_yarn-txt) (1200) roughly 30% (400 (^1)) of those commands are using the shorthand syntax to run a command installed into `node_modules/.bin` in a project.
+
+- Professional, it'd be great for the TypeScript docs to be able to simply say: `npm install typescript --save-dev' then `npm tsc index.ts` and there to be no ambiguity like our current recommendation for `npx tsc index.ts`.
```
Yash-Singh1, line 18:
```suggestion
- Professional, it'd be great for the TypeScript docs to be able to simply say: `npm install typescript --save-dev` then `npm tsc index.ts` and there to be no ambiguity like our current recommendation for `npx tsc index.ts`.
```

orta:
The results of this RFC a few meetings along. I'll be closing up this RFC as we've kinda hit the end-point of this one in that there's not really new discussion to be had.

### Running scripts/bins from a top level global e.g. `npm tsc` 

This is looking to be a no-go. Too many future constraints on npm, and it will cause any npm CLI change to _effectively_ be a server major.

There are two proposals worth keeping an eye on:
- https://github.com/npm/rfcs/pull/339 is likely to offer advice and tell you what commands to run.  Helping humans fix their mistakes.
- https://github.com/npm/rfcs/pull/279 when mixed with the below `npm exec-local` would get _an individual's_ setup to effective have the yarn-like top-level commands

### Abstracting local bins and scripts

No confirmations on what this could look like due to the tension of having to know that you use:

- `npm run` to do a script (`npm run compile`)
- `npm exec -- ` or `npx` to launch a local dep (`npm exec -- tsc`)

@isaacs had a good idea above with `npm exec-local` which could cover those two as a single command (and could maybe not have the `--`) - I don't know if there's an RFC for that but I can look into that if desired.

PR #339, opened 2021-03-10
rfc for improving command suggestions

nlf:
this is intended to improve discoverability of a working command when a user runs something that does not work, notably this is _an_ approach to addressing the needs of users who expect `npm foo` to work. instead of making an assumption that they want a script or binary, we will suggest that they try an explicit command that does one of those actions.

## References

Related to #336
Related to #332
Related to #279

### DIFF
--- a/accepted/0000-improve-command-suggestions.md
+++ b/accepted/0000-improve-command-suggestions.md
@@ -1,0 +1,104 @@
+# {{TITLE: Improve command suggestions}}
+
+## Summary
+
+Currently, we have a few places where a slightly incorrect command will display an error but print a "Did you mean?" type of suggestion.
+
+For example, `npm run tes` when you have a script named `test` will suggest running `npm run test`.
+
+These suggestions, however, are very limited and have minimal (if any) awareness of other commands.
+
+I propose that we allow an error from a failed command to suggest entirely different commands, this would allow `npm run tap` when no script named `tap` exists to suggest `npm exec -- tap`, and `npm exec test` to suggest `npm run test`.
+
+In addition, I propose that we remove typo-based "affordances" from commands in favor of this new suggestions engine. The breaking change there would be these affordances would stop working, and would instead inform the user of what command we believe we intended them to run.
+
+## Motivation
+
+Discoverability of the correct command to run when a user makes a common mistake is something we can improve. It's easy to confuse when to use `npm exec` vs `npm run`, and even easier to assume that `npm` works like `yarn` and allows you to `npm foo` for a script named `"foo"`. Instead of giving an unknown command error, we could instead present the user with a suggestion based on what we think they're trying to do. This is not quite as useful as unknown top level commands running scripts or bins, but it is far more explicit and safe.
+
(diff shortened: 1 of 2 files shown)


accepted/0000-improve-command-suggestions.md
```diff
@@ -0,0 +1,104 @@
...
+(run with -l or --long to see more context)t
+```
+
+Proposed behavior of `npm foo` when both a script named `"foo"` and a package named `foo` with a discoverable bin exist:
+```
+npm ERR! unknown command: foo
+npm ERR!
+npm ERR! Did you mean one of these?
+npm ERR!     npm run foo
+npm ERR!     npm exec -- foo
```
ljharb, line 70:
```suggestion
npm ERR! Did you mean one of these?
npm ERR!     npm run foo        # (runs the "foo" script)
npm ERR!     npm exec -- foo    # (runs the "foo" binary in node_modules)
```
maybe something like this?

nlf: great idea, each suggested command can come with a description

accepted/0000-improve-command-suggestions.md
```diff
@@ -0,0 +1,104 @@
...
+
+We do, however, maintain the benefits of helping the user determine what command they actually intended to run.
+
+Improving discoverability of the correct commands is a safer approach than creating ambiguous behavior.
+
+## Implementation
+
+I would suggest that we add a blessed property to errors thrown by commands `suggestions`. In the CLI framework, if a command rejects with an error that has this property, we display the suggestions.
+
+As for discovering the suggestions, we already have logic to find similarly named scripts based on levenshtein distance. This would need to be abstracted to a utility that can be consumed in `run-script` as well as `exec` and `help-search`.
```
isaacs, line 87: So if we do this, we could have the commands do something like `throw Object.assign(new Error('unknown command'), { suggestions: listOfEveryAvailableCommand })` and let the Levenshtein distance pick the best ones to prompt?

isaacs:
I think (ironically) this RFC is blending together a few things into one, and in the spirit of this RFC, it should be made explicit. They are:

1. Unambiguous partial abbrev spellings for known things (like `npm ins` for `npm install`)
2. Misspelling affordances (like `npm isntall`)
3. Unknown top level commands suggesting `run`/`exec`
4. "Blessed" affordances/shorthands like the `npm i` alias for `npm install`

For (3), I think that this is exactly the right approach, and seems to be where we keep landing every time we discuss the tradeoffs involved in making `npm tsc` work like yarn.

For (2), this seems reasonable enough, though it is a breaking change, and definitely not uncontroversial.  We could do a lot worse than copying git, and we have enough commands that some of the affordances have started to get a bit cluttered.

For (1), it's worth noting that git recently _added_ support for abbreviations on command-line switches, but not on top level commands.  If a command is an unambiguous abbreviation for a known top-level command, it seems like we have enough to know what the user intended.

For (4), I think the list could be examined and we could make some changes, but I am going to die on the hill of `npm i` and `npm t` continuing to work as they currently do.  Making the user type out the full word for those common commands just seems rude.

Ideas for a follow-up RFC:

- Config option to tell npm to just take the first suggestion.
- Config options to define shorthands and aliases.
- Choice in the prompt to save the suggestion as an alias when we do have an interactive mode, so users who truly cannot stop typing `hlep` can opt into the affordance.
- Similar rejection/suggestion for CLI configs, in some future time when we can drop support for parsing unknown config options.

PR #356, opened 2021-03-30
rfc: npm audit fix provides overrides

iarna:
Allow npm audit fix to update overrides for non-semver compatible changes.

* Edit: Not ALL semver incompatible changes. But as an optional kind of advice that says "this is semver incompatible but we assure you it will work anyway".

See [RFC](https://github.com/npm/rfcs/blob/41daf15ed78503a39b26e76771c4bda779e61fa9/accepted/0000-audit-overrides.md) for detail.

## References
Depends on: #129

### DIFF
--- a/accepted/0000-audit-overrides.md
+++ b/accepted/0000-audit-overrides.md
@@ -35,7 +35,7 @@ expectations in ways that are potentially fragile.)
 ### "Hot fixes"
 
 A critical security flaw is found in a dependency of one of your
-depenedencies.  The author of the transitive dependency is unavailable to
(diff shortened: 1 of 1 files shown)


isaacs:
Decided in 2021-04-14 meeting: ratify once overrides RFC is ratified (which is only pending final review)

isaacs:
I think that the endpoint return data suggested here is entirely appropriate.  If we see an `overrides` in the batch advisory result, we can just apply it to the root package.json.  (Or not and just suggest it, or apply if a flag is set, etc.  UX tbd.)

Resolving questions and moving to accepted.

PR #357, opened 2021-03-30
Fallback to kebab-case command

Yash-Singh1:
See [preview](https://github.com/Yash-Singh1/rfcs/blob/patch-1/accepted/0000-fallback-kebab-case-command.md#readme)

### DIFF
--- a/accepted/0000-fallback-kebab-case-command.md
+++ b/accepted/0000-fallback-kebab-case-command.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-If the command is unknown, then fallback to the kebab-case version of it if it is in the `PATH`.
+If the command is unknown, then fallback to it prefixed with `npm-` if that is in the `PATH`.
 
 ## Motivation
 
--- a/accepted/0000-fallback-kebab-case-command.md
+++ b/accepted/0000-fallback-kebab-case-command.md
@@ -1,0 +1,37 @@
+# Fallback to kebab-case command
+
+## Summary
+
+If the command is unknown, then fallback to the kebab-case version of it if it is in the `PATH`.
+
+## Motivation
+
+It allows users to create custom commands for `npm` such as:
+
+```sh
+npm my-command
+```
+
+## Detailed Explanation
+
+Let's say that `npm my-command` is run. Then `npm` should first see if their is a builtin command called `my-command`. If not then it will look through the path for `npm-my-command`. If that fails, then the current fallback will be run:
+
+1. Command exists?
+2. Custom command on `PATH`?
+3. Fallback
+
+## Rationale and Alternatives
+
+It is currently not possible to do this.
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


accepted/0000-fallback-kebab-case-command.md
```diff
@@ -0,0 +1,37 @@
...
+
+It allows users to create custom commands for `npm` such as:
+
+```sh
+npm my-command
+```
+
+## Detailed Explanation
+
+Let's say that `npm my-command` is run. Then `npm` should first see if their is a builtin command called `my-command`. If not then it will look through the path for `npm-my-command`. If that fails, then the current fallback will be run:
```
ljharb, line 17: what's the advantage of using `npm-my-command`? is this a convention that's used (besides perhaps git)?

accepted/0000-fallback-kebab-case-command.md
```diff
@@ -0,0 +1,37 @@
+# Fallback to kebab-case command
+
+## Summary
+
+If the command is unknown, then fallback to the kebab-case version of it if it is in the `PATH`.
```
ljharb, line 5: so if i do `npm fooBar` you'd expect it to be the same as `npm foo-bar`?

Yash-Singh1: No, `npm fooBar` would result in searching for `npm-fooBar` and `npm foo-bar` would result in searching for `npm-foo-bar`.

accepted/0000-fallback-kebab-case-command.md
Yash-Singh1: I didn't completely understand your question. If you were asking why I chose `my-command`, that was just an example. If you meant why custom commands should be a thing: Creating custom `npm` command will allow developers to "extend" `npm` with more CLI features for users.

ljharb: Allowing that would make "npm adds a new top-level command" a breaking change, and would also make `npm`'s interface different on a per-package basis.

accepted/0000-fallback-kebab-case-command.md
ljharb: in that case, i wouldn't call it "the kebab-case version of it", i'd call it "the command prefixed with `npm-`".

isaacs:
Doesn't this have all the same problems that killed the [`--default-command`](https://github.com/npm/rfcs/pull/279#issuecomment-823638195) proposal?

darcyclarke:
@isaacs yes; It's the same non-deterministic approach we've discussed at length previously & I believe the general consensus, & our team's own feelings, were that we don't want to move forward with any kind of fallback support for top-level commands. This type of implementation can cause confusion between developers (ex. docs/how-tos get written with customized/nuanced commands that aren't supported out-of-the-box) & clobbers the namespace, hampering our future development

PR #386, opened 2021-05-27
Add check-installed-peers rfc

Roaders:
Discusses the ability to be able to mark a peer dependency as optional so that it does not have to be installed by a consumer but if it is installed by a consumer then the version should be checked against the version specified by the installed dependency.

### DIFF
--- a/accepted/0000-check-installed-peers.md
+++ b/accepted/0000-check-installed-peers.md
@@ -1,0 +1,49 @@
+# Add `check-installed` option to `peerDependenciesMeta`
+
+## Summary
+
+Currently `peerDependenciesMeta` (as far as I know) only adds the ability to basically turn off peer dependency warnings for your peer. I would like to add the ability to mark a dependency that may or not be installed (`optional` but that terminology is already used) but who's version should be checked if it is installed.
+
+## Motivation
+
+I have encountered several situations where a package can have optional peers depending on the use case of the package. If the peer is installed though it will be important to have the correct version. Examples include:
+
+ * A config package that contains configs for various libraries. If the `eslint` config is used by the consuming package then the `eslint` peer needs to be installed but must be a minimum version for the config.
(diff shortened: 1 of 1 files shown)


accepted/0000-check-installed-peers.md
```diff
@@ -0,0 +1,49 @@
+# Add `check-installed` option to `peerDependenciesMeta`
+
+## Summary
+
+Currently `peerDependenciesMeta` (as far as I know) only adds the ability to basically turn off peer dependency warnings for your peer. I would like to add the ability to mark a dependency that may or not be installed (`optional` but that terminology is already used) but who's version should be checked if it is installed.
```
ljharb, line 5: This seems to me like it should be the default behavior. An optional peer dep will likely be try/catch required. If the wrong version exists, the program will happily use it, and perhaps break in terrifyingly silent ways.

I can't think of any use case where the package exists, the version is incompatible, and that's desirable.

Roaders: > This seems to me like it should be the default behavior. An optional peer dep will likely be try/catch required. If the wrong version exists, the program will happily use it, and perhaps break in terrifyingly silent ways.
> 
> I can't think of any use case where the package exists, the version is incompatible, and that's desirable.

Yes I completely agree. It would be much easier and better to modify existing behaviour but I was trying to avoid breaking what has already been implemented.

arcanis: Fwiw what this RFC describes is how Yarn implements it. That npm has a slightly different behavior wasn't intended and mostly just an oversight when I ported it. Seems a good idea to align.

Roaders:
On the open RFC Meeting last night we discussed this and it was confirmed that this is how it should be implemented in npm 7. I have tested this though and it does not seem to work as expected.

Roaders:
Ok, after saying it didn't work as expected. It does IRL (in the real world). It seems where I work we have some specific issues to us.
I will close this PR.
Thanks for the help.

PR #402, opened 2021-06-23
RFC: npm pkg

ruyadorno:
[See rendered RFC](https://github.com/npm/rfcs/blob/dd01b3d3aa54682c32ef1a0d329afa1ef528eaf7/accepted/0000-npm-pkg.md)

### DIFF
--- a/implemented/0039-npm-pkg.md
+++ b/implemented/0039-npm-pkg.md
@@ -1,0 +1,97 @@
+# npm pkg
+
+## Summary
+
+A new top-level command that manages any property in your `package.json`.
+
+## Motivation
+
+Some users prefer managing their `package.json` using **npm cli** comands rather than manually editing a JSON file.
+
+As of today, some of the **npm cli** commands already handle the automated management of the `package.json` file, such as `npm install` adding / updating dependency declarations, `npm version` bumping the version value, or in the case of `npm set-script` a top-level command that serves specifically the purpose of managing `"scripts"` data in the `package.json` file.
+
+## Detailed Explanation
+
+- Add a `npm pkg` command that has the following subcommands:
+  - `npm pkg get <value>`
+  - `npm pkg set <key>=<value>`
+  - `npm pkg delete <key>`
+
+## Rationale
+
+Making manage your `package.json` file something less manually curated and more of an automated experience.
+
+## Implementation
+
+- Add a new `lib/pkg.js` top-level command that will handle adding / removing / modifying properties in your `package.json` file.
+- Reuses the same logic existing to query for object properties from `lib/view.js`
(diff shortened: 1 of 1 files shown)


accepted/0000-npm-pkg.md
```diff
@@ -0,0 +1,93 @@
...
+$ npm pkg get scripts
+{
+  "lint": "eslint"
+}
+$ npm pkg get scripts.lint
+"eslint"
+$ npm pkg set scripts.lint.eslint.foo=bar
+error: "scripts.lint" exists and is not an object
+$ npm pkg delete scripts
+$ npm pkg set scripts.lint.eslint.foo=bar
```
ljharb, line 44: any reason this couldn't be combined into, like, `npm pkg set -f scripts.lint.eslint.foo=bar`?

accepted/0000-npm-pkg.md
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+}
+$ npm pkg set tap.timeout=60 --pkg-type=number
+$ npm pkg get tap
+{
+  "timeout": 60
+}
+$ npm pkg set contributors[0].name=Ruy contributors[0].email=foo@bar.ca
+$ npm pkg set contributors[1].name=Gar contributors[1].email=foo@bar.ca
```
ljharb, line 59: an example of how to add multiple items to the array in one command would be nice :-) (like, an "array push" example)

ruyadorno: I really dig your `[]` idea but that was not part of the `npm view` syntax, so I haven't brought it into it yet 😁 but I'm still considering adding it to the parser!

now adding multiple items at once might be tricky... I really think it's a good idea to steer away from handling objects (like we discussed in the last meeting) so at first the idea is to only handle one literal value at a time

accepted/0000-npm-pkg.md
ruyadorno: I like that idea! 😄

accepted/0000-npm-pkg.md
ljharb: `npm pkg set contributors[].name=Bob contributors[].name=Jan` to add two new objects with the respective names?

ruyadorno: as long as that new `[]` syntax lands, yeah it should!

ruyadorno: @ljharb just to make sure you're aware, this syntax has not yet landed in this initial implementation but I'm looking forward to seeing it land in the future 😄

ruyadorno: actually added it for `npm pkg set` here: https://github.com/npm/cli/pull/3536

PR #434, opened 2021-08-13
Support package-lock.json v3 in npm 7

remcohaszing:
npm 7 introduces new features which aren’t compatible with older npm versions, such as workspaces. This makes npm 7 a minimum requirement for a project.

Both `package-lock.json` version 1 and 3 files can be huge. `package-lock.json` 2 is an intermediate format whose goal is to add backwards compatibility with `package-lock.json` version 1 and npm 6. It does so by writing both the dependency trees in the version 1 and 3 formats. This means the file is even twice as big as it needs to be. This also means the same change appears twice in a diff, but in a different format, meaning it needs to be reviewed twice. This is unnecessary if the minimal required npm version is 7.

## References

Related to https://github.com/npm/feedback/discussions/517


accepted/0000-npm7-support-package-lock-6.md
```diff
@@ -0,0 +1,59 @@
+# Support `package-lock.json` v3 in npm 7
+
+## Summary
+
+Currently npm 7 can read `package-lock.json` versions 1, 2, and 3, but it always writes version 3. It would be nice if npm 7 could write `package-lock.json` version 3 instead.
```
ljharb, line 5:
```suggestion
Currently npm 7 can read `package-lock.json` versions 1, 2, and 3, but it always writes version 2. It would be nice if npm 7 could write `package-lock.json` version 3 instead.
```
?

accepted/0000-npm7-support-package-lock-6.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Motivation
+
+npm 7 introduces new features which aren’t compatible with older npm versions, such as workspaces. This makes npm 7 a minimum requirement for a project.
+
+Both `package-lock.json` version 1 and 3 files can be huge. `package-lock.json` 2 is an intermediate format whose goal is to add backwards compatibility with `package-lock.json` version 1 and npm 6. It does so by writing both the dependency trees in the version 1 and 3 formats. This means the file is even twice as big as it needs to be. This also means the same change appears twice in a diff, but in a different format, meaning it needs to be reviewed twice. This is unnecessary if the minimal required npm version is 7.
```
ljharb, line 11:
```suggestion
Both `package-lock.json` version 1 and 2 files can be huge. `package-lock.json` 2 is an intermediate format whose goal is to add backwards compatibility with `package-lock.json` version 1 and npm 6. It does so by writing both the dependency trees in the version 1 and 3 formats. This means the file is even twice as big as it needs to be. This also means the same change appears twice in a diff, but in a different format, meaning it needs to be reviewed twice. This is unnecessary if the minimal required npm version is 7.
```
?

ljharb:
Instead of letting the user control this, could it pivot based on the `engines.npm` value? In other words, if npm 7 is already declared as a requirement, then it knows it doesn't need to write npm 6 back compat stuff.

accepted/0000-npm7-support-package-lock-6.md
remcohaszing: Nope, I meant versions 1 and 3. Version 2 includes both, which makes it even bigger.

remcohaszing:
> Instead of letting the user control this, could it pivot based on the `engines.npm` value? In other words, if npm 7 is already declared as a requirement, then it knows it doesn't need to write npm 6 back compat stuff.

I didn’t think of that, but this is an interesting idea.

I think the original proposal is better though, because:

1. Since the `engines` field in `package.json` is published with the package, I think this implies it’s a runtime requirement, whereas the lockfile format is a development requirement for the package itself.
2. The original proposal to store it in `.npmrc` is consistent with the usage of `package-lock=false`

Another idea might be to reuse the `package-lock` field in `.npmrc` that already exists.

```ini
# Don’t write package-lock.json at all
package-lock=false
# Write package-lock.json version 1
package-lock=v1
# Write package-lock.json version 2
package-lock=v2
# Write package-lock.json version 3
package-lock=v3
# Write the package-lock.json version npm thinks is best (default behaviour, equal to v2 for npm 7)
package-lock=true
```

accepted/0000-npm7-support-package-lock-6.md
```diff
@@ -0,0 +1,59 @@
...
+
+```ini
+package-lock-version=3
+```
+
+Major npm versions may choose to drop support for old formats.
+
+## Rationale and Alternatives
+
+1. A command line flag could be supported instead of an option in `.npmrc`. However, typically one would want to support this on a project level.
```
bnb, line 39: My understanding is that typically command line flags are _also_ usable in `.npmrc` - so supporting `npm install --package-lock-version=<number>` would also allow you to use `package-lock-version=<number>` in `.npmrc`.

IMO the path of both is ideal.

bnb:
I particularly like the recommendation to extend the `package-lock` property. One thing to note is that you'd probably want to start shipping npm with a range of acceptable numbers, a way to find those numbers (perhaps `npm  info package-lock` or something), and to error on a number that isn't within the range of acceptable numbers.

isaacs:
We should definitely not make `--package-lock` more than a simple boolean.  It's much better to have _one_ config that's a boolean saying "should I write a package-lock.json _at all_" and another config that's a number saying "what kind of package lock should be written, when/if we write one".

Mixed-type configs are an antipattern.

isaacs:
Suggestions:

- The config key should be `lockfile-version`, not `package-lock-version`, since it applies to both `npm-shrinkwrap.json` and `package-lock.json`, and corresponds to the `lockfileVersion` field in the lock file itself.
- It should be just a plain old config value, which can be set any of the places that config values can be set.  So, `--lockfile-version` cli option, `npm_config_lockfile_version` env, project .npmrc, user .npmrc, or global npmrc files.
- It should be a numeric config value, with a default value of `2`.  Non-numbers (eg, `v2`, `version 2`) etc. will be ignored.
- Config value should be flattened to `lockfileVersion` option passed to `@npmcli/arborist`.  (See npm/arborist#329 for proposed implementation.)

isaacs:
> I believe Arborist should always default to lockfile v2 for now for backwards compatibility. This should then be reconsidered for npm 8.

Right, but the use case I'm imagining is:

```
npm i --lockfile-version=3 # create a v3 lockfile
npm i foo # now it just reset the lockfile version back to 2
```

If the lockfile is _already_ v3, maybe we should default to that, rather than downgrading it?

isaacs:
> It would seem like a bug to me if it downgrades it or upgrades it implicitly.

"Upgrade it implicitly" is what we already do.  The v1 lockfile doesn't have all the information we need to fully construct the tree, meaning that installs are slower and less deterministic, don't track changes in link targets, etc.

But _downgrading_ implicitly definitely seems like the wrong thing to do, since you must have opted into the new version explicitly to be in that state.

xiaoxiyao:
Which version of npm 7 started to support lockfileVersion v3?

remcohaszing:
The initial support for lockfile version 3 had some bugs. It’s been usable in practice since npm version 8.1.2.

PR #436, opened 2021-08-23
new installation mode: isolated-mode

VincentBailly:
This RFC is a proposal to add a new opt-in installation mode.

This mode would be the missing ingredient to make workspaces truly scalable.


p01:
Thank you!

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+Several build tools took advantage of these declared dependencies to split the builds into multiple steps (one per workspace) which could be cached or run in parallel. This caching and parallelism usually improves significantly build performance.
+
+The current default installation strategy does not properly communicates to Node.js the information about the dependency graph. Node.js having an inaccurate view of the dependency graph leads the build systems to wrongly skip build operations, leading to risk of bugs slipping through.
+
+
+The installation mode proposed by this RFC would communicate accurately to Node.js the dependency graph, allowing build tools to correctly reason about it.
+
+## Naming
+
+The proposed name for this mode is `pure-mode`, meaning "pure" as in "no side effects". This name comes from the parallel with functional programming; in functional programming, functions don't have side effects, in `pure-mode` workspaces have no side effects.
```
p01, line 22:
```suggestion
The proposed name for this mode is `pure-mode`, meaning "pure" as in "no side effects". This name comes from the parallel with functional programming; in functional programming, functions are called `pure` because they don't have side effects, in `pure-mode` workspaces have no side effects.
```

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Configuration
+
+The `pure-mode` is a project level setting, it does not make much sense to install some dependencies in `pure-mode` and others in `hoisted-mode`. For this reason, to configure a project in `pure-mode`, a setting has to be added to the `.npmrc` of the project and the lockfile has to be re-generated. A tool may be made available to convert `hoisted-mode` lockfiles to `pure-mode` lockfiles.
+
+## Compatibility
+
+### Different versions of npm
+
+As workspaces are supported only from npm version 7, there is no concern about making `pure-mode` compatible with versions olders than 7.0.0.
```
bnb, line 299: perhaps worth only supporting `package-lock` v3 or even a new v4 and above, then?

I honestly would not mind a feature this impactful being available starting in a new semver major to both make it clear which versions are supported by the feature and provide a good incentive to update.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+
+The tree structure will be serialized to the lockfile in a way that any version of npm 7.0.0 can understand. This means that any version of npm 7 will be able to install a project configured in `pure-mode`.
+
+Modifying a dependency in a `pure-mode` project with a version of npm not supporting `pure-mode` is likely going to leave the project in a broken state and should be avoided.
+
+When using `pure-mode` it will be recommended to properly configure the `engines` field of the `package.json` to make sure developers who want to modify the dependencies don't shoot themselves in the foot and put the project in an instable state.
+
+### Projects not using workspaces
+
+The `pure-mode` will also be available for projects not using the workspace feature. This is useful as a library author, using `pure-mode` gives me the guarantee that my library will work as intended when consumed by projects which use `pure-mode`.
```
bnb, line 309: Disappointed to see that non-workspaces projects are excluded from `pure-mode`. I don't fully understand what the assertion about being useful as a library author is... could you expand on that and why it's a good justification for preventing `pure-mode` from being used by non-workspaces projects?

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+### [Import maps](https://github.com/WICG/import-maps)
+
+Standard supported by [a few browsers](https://caniuse.com/import-maps) and [deno](https://deno.land/) which makes it possible to implement `pure-mode`.
+
+## Unresolved Questions and Bikeshedding
+
+- Should we use symlinks or junctions on Windows? Both of them have drawbacks:
+
+  - Junctions have to be representated by an absolute path, this means that junctions cannot be committed to git or packed into a package.
+  - Symlinks can only be created in elevated shell [or when Windows is in "developer mode"](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/#LCiVBWTgQF5s7fmL.97).
```
bnb, line 340: I don't think it's unreasonable to ask people to turn on `developer mode` or run in an elevated shell, but I could be wrong.

bnb (commented):
Largely some small fixes, but there are a few questions. I would _really_ love to see `pure-mode` available for non-workspaces both for consistency (creating two classes of DX isn't a good DX imo) and for potential perf gains (if this were implemented as a user-level package store, workspaces could *also* opt-in to that and see even better perf improvements in some situations), but if overall the npm CLI team and other RFC participants disagree I'm willing to back off of that.

I also see this as a path towards flexibility: if we have a `pure-mode` and a `hoisted-mode`, what's the reason we wouldn't add another mode when someone who is excessively smart finds yet another more optimized installation approach with good DX?

Overall, I'm very happy with this. Thank you for the hard work @VincentBailly <3

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
+# Pure mode
```
ljharb, line 1: i don't think "pure" is the right term to use here

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+Several build tools took advantage of these declared dependencies to split the builds into multiple steps (one per workspace) which could be cached or run in parallel. This caching and parallelism usually improves significantly build performance.
+
+The current default installation strategy does not properly communicates to Node.js the information about the dependency graph. Node.js having an inaccurate view of the dependency graph leads the build systems to wrongly skip build operations, leading to risk of bugs slipping through.
+
+
+The installation mode proposed by this RFC would communicate accurately to Node.js the dependency graph, allowing build tools to correctly reason about it.
+
+## Naming
+
+The proposed name for this mode is `pure-mode`, meaning "pure" as in "no side effects". This name comes from the parallel with functional programming; in functional programming, functions are called `pure` because they don't have side effects, in `pure-mode` workspaces have no side effects.
```
ljharb, line 22: node module resolution does and *should* mean that a parent package.json's node_modules should be requireable in a child - so i don't think "no side effects" is actually what this would be.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Forgetting to declare dependencies
+
+When workspaces can successfully use code of a package without having a dependency on it, people forget to declare their dependencies. This lead to situations where updating the dependencies of one workspace breaks a seemingly unrelated workspace.
+
+It is worth noting that static code analysis tools can help significantly reduce the frequency of these mistakes.
+
+### Duplication
+
+The `hoisted-mode` is not always successful at sharing common dependencies. Conflicts in versions lead to packages being installed multiple times on disk. These conflicts become more frequent when the project scales. The frequency of these conflicts can be reduced by an effort to align dependencies' versions across the project.
```
ljharb, line 40: "sharing" is i think the key word here. I've discussed this concept on the RFC call multiple times; basically, every package.json should have a node_modules next to it, full stop.

However, other sibling packages, and *also* any common peer deps between sibling packages, should be *shared* - without hoisting.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### Installation on disk
+
+```
+  root /
+   ┬
+   │
+   ├─> node_modules / .npm / ─┬─> A@1.0.0-21f95f7 / node_modules / A / ─┬─> [content of package A]
+   │                          │                                         │
+   │                          │                                         └─> node_modules / B ( symlink to ../../../../B@1.0.0-0d98566/node_modules/B )
```
zkochan, line 206: This is not a good idea. This will result in circular symlinks. The symlink to be should be at 

```
├─> node_modules / .npm / ─┬─> A@1.0.0-21f95f7 / node_modules / B
```

From the docs of pnpm

> avoid circular symlinks. Dependencies of packages are placed in the same folder in which the dependent packages are. For Node.js it doesn't make a difference whether dependencies are inside the package's node_modules or in any other node_modules in the parent directories.

https://pnpm.io/symlinked-node-modules-structure

accepted/0000-pure-mode.md
p01: The new mode does bring very significant performance improvements,  however as an opt-in `pure-mode` can be introduced as a semver minor.

accepted/0000-pure-mode.md
VincentBailly: @bnb you must have read too fast. The proposal is to make `pure-mode` available for non-workspaces projects.

accepted/0000-pure-mode.md
VincentBailly: Thanks a lot @zkochan, it makes total sense, I updated the RFC

accepted/0000-pure-mode.md
VincentBailly: The suggested `pure-mode` requires a package to declare a `peer-dependency` in order to be able to impact the dependencies of its parent. Importing code from one of your siblings without having a peer-dependency declared contributes to the problems that this RFC is trying to solve.

accepted/0000-pure-mode.md
VincentBailly: Are you suggesting that this new mode should be called `shared-mode`?

accepted/0000-pure-mode.md
VincentBailly: I updated the reasoning for choosing the term `pure-mode`. Please have a look, and if you don't agree please explain why.

accepted/0000-pure-mode.md
VincentBailly: I will leave that decision to the npm CLI team. I am comfortable with either.

accepted/0000-pure-mode.md
VincentBailly: The package-lock does not appear to need an upgrade to support `pure-mode` because it already support describing symlinks.

Shipping this mode in a major release may simplify the compatibility story, I will let the npm CLI team judge what is best here.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Each workspace has its own `node_modules` folder.
+- Each `node_modules` folder contain symlinks.
+- The symlinks in a `node_modules` folder correspond to the matching workspace's dependencies.
+- Each symlink is given the same name as the dependency it represents.
+- Each symlink points to a location in the package store where the corresponding package is installed.
+
+In the package store:
+
+- The package store is a folder stored in the root level `node_modules` folder.
```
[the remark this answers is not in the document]
VincentBailly: Is it clearer now?

accepted/0000-pure-mode.md
ljharb: I don’t think that’s sufficient - users will always want every child package to be shared also, whether it’s a peer or not.

accepted/0000-pure-mode.md
ljharb: Yes, and that it should automatically share both duplicate peers as well as child packages.

accepted/0000-pure-mode.md
VincentBailly: Can you clarify the scenario that this mode needs to support?

Is there any data (article, discussion, blog post...) about users expecting to have access to parent's dependencies?

accepted/0000-pure-mode.md
VincentBailly: hoisting is also a sharing mechanism so `shared-mode` does not really contrast with hoisting.

accepted/0000-pure-mode.md
ljharb: That’s how node has always worked - every use of `require` expects that. That’s not what we’re talking about tho.

I’m claiming that what users want is to have both peers and siblings automatically shared, when matching. Currently, lerna/yarn/npm does this via hoisting, which has tons of drawbacks. I think what we need is a way to do it *not* solely with hoisting.

No matter what, a parent package.json’s node_modules must always be requireable because that’s how node works.

accepted/0000-pure-mode.md
VincentBailly: I know, but this hoisting is sharing 95% of the time. So should we call this new mode "100% sharing-mode"?

accepted/0000-pure-mode.md
VincentBailly: My assumption is that the fact that a package can require its siblings is a side effect of the package managers implementation, not a requirement described anywhere. Or is it specified somewhere?

pnpm and yarn PnP do work without fulfilling this expectation so I would guess that most packages are not relying on it. For the ones that do, it would be very easy to fix by declaring a peerDependency to properly express this expectation.

accepted/0000-pure-mode.md
ljharb: Hoisting is sharing *all* of them. The new mode I think we need is sharing only peers and siblings, which imo is the only thing everyone actually wants.

That parent package's deps are requireable in children isn't hoisting, it's just a natural artifact of the way node's always worked, and no workspace mode should interfere with that.

accepted/0000-pure-mode.md
fritzy: I understand the analogy, and obviously this is pure semantics.

* pure can have negative cultural and historic connotations related to race and religion
* pure implies a value judgement
* the pure function analogy isn't immediately obvious to the reader and wouldn't significantly preload accurate assumptions if they did

accepted/0000-pure-mode.md
ljharb: Absent workspaces, a package can not require its siblings' deps, but it can always require its parent's deps.

PnP has to patch a number of parts of the ecosystem (in its pnp.js file) to ensure things don't break, and both PnP and pnpm have nonzero bugs caused by their approach. I think that it would be a very bad idea for npm to apply this approach, *especially* only on workspaces.

accepted/0000-pure-mode.md
VincentBailly: Makes sense, I guess we have to find a better name then.

fritzy:
I think we're conflating two things here: workspace hoisting to root and tree hoisting within an individual package/workspace.

**Let's create a separate RFC for workspace hoisting and treat these proposals separately on their own merits.**

accepted/0000-pure-mode.md
fritzy: Simply "unhoisted mode" might be sufficient.

isaacs:
> Let's create a separate RFC for workspace hoisting and treat these proposals separately on their own merits.

Strong agreement.  That's exactly what https://github.com/npm/rfcs/pull/375 was intended to do, but didn't get finished.  Wanna pick it up?

VincentBailly:
> I think we're conflating two things here: workspace hoisting to root and tree hoisting within an individual package/workspace.
> 
> **Let's create a separate RFC for workspace hoisting and treat these proposals separately on their own merits.**

Thanks @fritzy for pointing this out. This RFC aims at removing sharing between workspaces, this is what is described in the motivations. The fact that this RFC also suggests to not hoist transitive-dependencies is the result of the suggested implementation but not a goal in its own. I don't see the need to create an RFC about workspace hoisting because this is what this one is meant to be. If we find a way to not share dependencies between workspaces without removing the hoisting within a workspace, it would also address the problem raised by this RFC.

ljharb:
Sharing between workspaces is necessary for peer deps and for sibling packages; i'm not clear on why anyone would put projects into a monorepo in the first place unless sharing was needed.

VincentBailly:
> Sharing between workspaces is necessary for peer deps and for sibling packages; i'm not clear on why anyone would put projects into a monorepo in the first place unless sharing was needed.

I don't think there is a contract anywhere that a package/workspace should be able to import the dependencies of its siblings. But I am far from an authority here, @isaacs can you comment on this?

fritzy:
@VincentBailly we already have https://github.com/npm/rfcs/pull/375 (@isaacs pointed me to it yesterday). I understand that your primary motivation was workspaces, but I think both efforts are very valuable! I suspect that we'll be able to act on the workspaces specific with less side effects and as a default breaking change, so I think we should focus there. But I don't think we should drop the conversation about a strict/pure/unhoisted mode for package reifying in general. The conversations we've had here on that subject are still valuable and probably still worth pursuing an implementation. There's a lot of valuable conversation in this PR and details in the RFC that apply to packages that we haven't captured anywhere else.

Let's move the workspaces work to the previous RFC instead and bring our new knowledge there, leaving this one intact for strict/pure/unhoisted non-workspace output.

isaacs:
Regarding teasing apart the concerns here:

#375 is primarily about what dependencies are _shared_ between workspaces, from a consumer semantics point of view.  (Ie, so if `workspace-a` does `require('foo')` it will get the exact same object as if `workspace-b` does `require('foo')`: always? sometimes? never? configurable?)

Currently, we do not impose any special constraints or contracts around dependency sharing between workspaces.  Which means, everything gets deduped up to the top, just as if they were normal link deps within the project.

This is desirable in some situations (ie, integration testing between sibling packages, and ensuring that they can operate within the same peer set).  However, it is undesirable in others (introduces a high possibility of shadow deps).

The question of what dependencies _ought_ to be shared between workspaces, and how a user might express explicit intent regarding that sharing, is a question for #375, which does not propose any specific reification strategy.  (That is, we could come to rough consensus on #375, and _probably_ implement whatever we decide today, with our current reification strategy.)

This RFC, #436, is about hoisting as such, as I see it.  Whether you use workspaces or not, unhoisted mode (or whatever we call it, soon to be formerly known as "pure mode") will arrange the files on disk using some kind of hashing and symlink strategy we decide on.  The benefit of this strategy will be to prevent duplication and make shadow deps a non-issue.

While workspaces may be the _motivation_ for it (since large projects will get the most benefit, and monorepos with many workspaces are a common sort of "large project" that we care about), it's going to be the same strategy as if you just had a non-workspaces project and for whatever reason wanted to prevent duplication and shadow deps.  For example, you might have a monolith with a lot of dependencies, which isn't split up into workspaces, but still benefits from unhoisted mode.

> I don't think there is a contract anywhere that a package/workspace should be able to import the its siblings in the absence of declared peerDependency on them. But I am far from an authority here, @isaacs can you comment on this?

There is not, no.  There is the de facto behavior that any installed workspaces _can_ typically `require()` one another, but only because they are reified up to the root.

The contract proposed in #375 is that, _if_ a workspace depends (peer or otherwise) on another workspace within the same project, it will _always_ resolve its dependency to its sibling workspace.  Currently, this is not a guarantee.  If there is a dependency version conflict, then a second copy of the sibling workspace _may_ be duped into the dependent workspace.

isaacs:
Approach we landed on in the RFC call yesterday:

- We seem to be calling it "unhoisted mode"?  Is that the name now?  (Personally, I don't love that name, it's a mouthful, but I suppose it is explicit, somewhat descriptive, and free of the connotations of "pure".)
- Since the unhoisted reification strategy _can_ replicate anything that the hoisted mode can (as well as many other things), we can still provide the same ERESOLVE and dep sharing contracts, by creating the traditional hoisted tree, and then transmuting it into an unhoisted symlink-based tree prior to reification.
- The `./package-lock.json` file will reflect the _hoisted_ (traditional, nested folders, no symlinks) style tree.
- The `./node_modules/.package-lock.json` file will reflect the _unhoisted_ (symlinks, no shadow deps, "pure") style tree that actually lives on disk.
- We'll need a way reconcile _back_ from unhoisted to hoisted mode for cases where we build the idealTree starting with the `node_modules` structure (ie, the "has node_modules, no lockfile" use case).
- `./package-lock.json` will no longer be guaranteed to match `node_modules` structure.  But this is good, because it means that two devs on a project can each use different reification strategies, but still work with the same dependency graph.

accepted/0000-pure-mode.md
bnb: @fritzy I feel like that could have a negative impact on user perception compared to a unique name. Why _wouldn't_ I want my packages hoisted? That sounds like a good thing. Further, "unhoisted" makes the addition of new modes later harder since you'll have hoisted, unhoisted, and <x> where <x> is also conceptually "unhoisted" but that name was already taken.

IMO it's better to not have a negative prefix to describe what this is doing but instead have a direct name.

+1 to "isolated" - of the options here, it feels like the least loaded ("non-flat" suffers from a negative prefix with a new term for a concept that's already defined, "strict mode" is already a JavaScript thing, and "predictable" does have an assertion of value)

arcanis:
> We seem to be calling it "unhoisted mode"?  Is that the name now?  (Personally, I don't love that name, it's a mouthful, but I suppose it is explicit, somewhat descriptive, and free of the connotations of "pure".)

Fwiw Yarn also implemented this feature a few months ago; so far we plan to call it "pnpm linker" once it officially ships, since that's how everyone already knows it and we like the idea of acknowledging the work pnpm did to make this approach more mature in the ecosystem; my image

isaacs:
> Fwiw Yarn also implemented this feature a few months ago; so far we plan to call it "pnpm linker"

I'd personally love calling it "pnpm mode", and we'll definitely call it out in the docs as being inspired by pnpm, but I think that name relies on a lot of JS package manager inside-baseball knowledge that people outside these conversations are unlikely to have.

accepted/0000-pure-mode.md
zkochan: I will throw another idea: hidden-subdependencies

accepted/0000-pure-mode.md
VincentBailly: We have decided to that using pure-mode will not affect the lock file at all, this mode will only affect how packages are laid out on disk but the lock file will stay the same.

accepted/0000-pure-mode.md
VincentBailly: Junctions will be used. We will make an option to opt-in symlinks later if there is a need for it.

accepted/0000-pure-mode.md
VincentBailly: A poll was created to vote on a name

accepted/0000-pure-mode.md
VincentBailly: Making sure that I understand your point well @ljharb :
If packages `foo` depends on `bar` and `cat`, you are saying that `cat` should be able to require `bar` even if no peer dependency to `bar` in declared in `cat`'s package.json?

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,351 @@
...
+
+```
+
+## Performance
+
+Compared to the current npm installation strategy, this proposal reduces package duplication, making the installation process faster. On a prototype, this installation strategy brought down the install time from 6 to 2 minutes on a large monorepo of 500+ workspaces.
+
+## Configuration
+
+The `pure-mode` is a project level setting, it does not make much sense to install some dependencies in `pure-mode` and others in `hoisted-mode`. For this reason, to configure a project in `pure-mode`, a setting has to be added to the `.npmrc` of the project and the lockfile has to be re-generated. A tool may be made available to convert `hoisted-mode` lockfiles to `pure-mode` lockfiles.
```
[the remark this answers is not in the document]
VincentBailly: I think we agree here, I used the term 'project' to mean the entire repo. I rephrased this paragraph so it should not be confusing anymore.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,351 @@
...
+Standard supported by [a few browsers](https://caniuse.com/import-maps) and [deno](https://deno.land/) which makes it possible to implement `pure-mode`.
+
+## Unresolved Questions and Bikeshedding
+
+- Should we use symlinks or junctions on Windows? Both of them have drawbacks:
+
+  - Junctions have to be representated by an absolute path, this means that junctions cannot be committed to git or packed into a package.
+  - Symlinks can only be created in elevated shell [or when Windows is in "developer mode"](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/#LCiVBWTgQF5s7fmL.97).
+
+  - answer: junctions by default and symlinks as opt-in
```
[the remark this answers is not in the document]
VincentBailly: agreed. Updated this part

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,356 @@
...
+
+  - answer: junctions by default. support for symlinks can be added later as an opt-in if we see value for it.
+
+- How much community code will break when the system forbids access to undeclared dependencies? In other words, how much code needs to be fixed to work properly in `pure-mode`?
+  - Regarding packages with missing dependencies in the package.json files. The vast majority of the package owners are happy to fix them and the rest simply need to know that this fix is necessary to make their package work great with npm. So the number of packages with missing dependencies is expected to drop fast after npm release `pure-mode`. We may want to later add a feature to npm which allows users to locally declare dependencies on behalf of packages, though this should just be a stop-gap solution.
+  - There don't seem to be any package out there that depend on the way hoiting work as a feature. If this is the case, it would be easy to argue that these packages should only rely on the npm and the Node.js contracts/APIs and not on implementation details like hoisting.
```
jcreamer898, line 352:
```suggestion
  - There don't seem to be any packages out there that depend on the way hoisting works as a feature. If this is the case, it would be easy to argue that these packages should only rely on the npm and the Node.js contracts/APIs and not on implementation details like hoisting.
```

ljharb: This line is incorrect - eslint-plugin-import relies on this, does everything in the babel and eslint ecosystems, because these things have non-javascript config files, referencing package names, that are not in the dependency graph of the main CLI.

accepted/0000-pure-mode.md
```diff
@@ -0,0 +1,356 @@
...
+
+  - Junctions have to be representated by an absolute path, this means that junctions cannot be committed to git or packed into a package.
+  - Symlinks can only be created in elevated shell [or when Windows is in "developer mode"](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/#LCiVBWTgQF5s7fmL.97).
+
+  - answer: junctions by default. support for symlinks can be added later as an opt-in if we see value for it.
+
+- How much community code will break when the system forbids access to undeclared dependencies? In other words, how much code needs to be fixed to work properly in `pure-mode`?
+  - Regarding packages with missing dependencies in the package.json files. The vast majority of the package owners are happy to fix them and the rest simply need to know that this fix is necessary to make their package work great with npm. So the number of packages with missing dependencies is expected to drop fast after npm release `pure-mode`. We may want to later add a feature to npm which allows users to locally declare dependencies on behalf of packages, though this should just be a stop-gap solution.
```
bnb, line 351: I've found it's often not productive to assume how maintainers are willing to maintain their packages. While I generally don't disagree, I think it would be helpful to not codify assumptions.

```suggestion
  - Regarding packages with missing dependencies in the `package.json` file. In many cases, package owners are willing to fix such issues. If there are a significant number of issues encountered around this, we can likely expect the of packages with missing dependencies is expected to drop fast after npm release `pure-mode`. We may want to later add a feature to npm which allows users to locally declare dependencies on behalf of packages as a stop-gap, if existing solutions to this are not enough.
```

accepted/0000-pure-mode.md
VincentBailly: This is a good point, the config files of babel/webpack/eslint... which support simply giving a package name will be broken by the pure-mode. Fortunately, all these tools also allow .js config files which can contain `require.resolve(aPlugin)`, so repos using pure-mode may need to update their config files.

ljharb: That would be forcing people to use JS for config, which is not universally considered a good practice.

The tools allow logicless config, and it is simply *not viable* to dismiss those projects with "psh, just change your config files or you can't use this new mode".

VincentBailly: Developers will be able to keep using their tools and their config files by adding the packages referenced in the config files as dependencies in the top level package.json. This is because dependencies of the top-level package.json are globally accessible within the repo. This is simply not the recommended approach.

accepted/0000-pure-mode.md
VincentBailly: Discussed during the RFC call and @ljharb said that this is not a concern for him anymore.

accepted/0000-pure-mode.md
VincentBailly: Discussed during the RFC call and @ljharb said that this is not a concern for him anymore.

accepted/0000-pure-mode.md
VincentBailly: Discussed during the RFC call and @ljharb said that this is not a concern for him anymore.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Forgetting to declare dependencies
+
+When workspaces can successfully use code of a package without having a dependency on it, people forget to declare their dependencies. This leads to situations where updating the dependencies of one workspace breaks a seemingly unrelated workspace.
+
+It is worth noting that static code analysis tools can help significantly reduce the frequency of these mistakes. While static code analysis tools are probably sufficient for small projects, they are not for large projects. "isolated-mode" not only helps get the dependencies right but also _proves_ that the dependencies are correct. This proof is essential to build trustable build systems which use caching, scoped installations.
```
ljharb, line 34: Why are they not for large projects? I’ve used many extremely large projects with eslint-plugin-import and it works just fine.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Duplication
```
ljharb, line 36: Is this something orthogonal? iow, this seems like something worth solving for both modes, not just for this one.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+Certain tools implement their own module resolution algorithm instead of the one provided by Node.js. One of the motivation for a tool to implement its own resolution algorithm is that it can add more features to it.
+
+Crawling the `node_modules` folder is such a feature used by certain build tools. This feature means that the mere presence of a package in a `node_modules` folder will have an impact on the output of a workspace. This means that any modification of a project's `node_modules` folder is possibly a breaking change to every workspaces in this project, regardless of the dependency graph.
+
+For example, the TypeScript compiler by default includes to its compilation _every_ package in the folder `node_modules/@types` without any explicit import statement needed.
+
+## Rationale
+
+The [import-maps](https://github.com/WICG/import-maps) standard is a very good tool to communicate the dependency graph to Node.js. But because this standard is not yet implemented in Node.js, npm needs an alternative way to solve the problems stated earlier.
```
ljharb, line 56:
```suggestion
The [import-maps](https://github.com/WICG/import-maps) draft standard is a very good tool to communicate the dependency graph to Node.js. But because this feature is not yet implemented in Node.js, npm needs an alternative way to solve the problems stated earlier.
```
I believe it would require more than one browser implementation to be an actual standard

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+The goal is to provide an accurate dependency graph to Node.js while still relying on Node.js current module resolution algorithm.
+
+We deem valuable to invest the time implementing the `isolated-mode` now knowing that import-maps are coming because most of the implementation will be re-usable to implement a `import-maps-mode`.
```
ljharb, line 60:
```suggestion
We deem it valuable to invest the time implementing `isolated-mode` now, knowing that import-maps are coming, because most of the implementation will be re-usable to implement an `import-maps-mode`.
```

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+There are already two competing solutions out there which implement a `isolated-mode` ([pnpm](https://pnpm.io) and [yarn](https://yarnpkg.com)). We decided to choose the pnpm approach for the following reasons:
```
ljharb, line 62:
```suggestion
There are already two competing solutions out there which implement an `isolated-mode`: ([pnpm](https://pnpm.io) and [yarn](https://yarnpkg.com)). We decided to choose the pnpm approach for the following reasons:
```

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+- _Works with current ecosystem_, the pnpm `isolated-mode` does not require any modification to Node.js or to the various build tools.
+- _Battle tested_, Microsoft has successfully used pnpm to manage large monorepo for years.
+- _Recommended by Node.js_, this approach is actually the recommended approach by [the Node.js documentation](https://nodejs.org/api/modules.html#modules_addenda_package_manager_tips).
```
ljharb, line 66: This link doesn’t go anywhere relevant - i was very shocked to read the implication that node made any official recommendation here. It’d be great to track down this content, or remove this bullet point if it can’t be located.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+- The symlinks in a `node_modules` folder correspond to the matching workspace's dependencies.
+- Each symlink is given the same name as the dependency it represents.
+- Each symlink points to a location in the package store where the corresponding package is installed.
+
+In the package store:
+
+- The package store is a folder stored in the root/project level `node_modules` folder.
+- The package store is named `.npm`.
+- The package store contains folders, one for each dependency installed. These dependencies can be direct workspace dependencies or transitive dependencies.
+- These folders have a name containing the name of the package, its version and a hash of its content plus the hash of its dependencies.
```
ljharb, line 97: what exactly is “its dependencies”? If it’s already hashing the content why is an additional hash needed?

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Configuration
+
+The `mode` can be enabled by a CLI option when running `npm install` or by a setting in `.npmrc`.
+
+## Compatibility
+
+### Different versions of npm
+
+The `isolated-mode` will use the same lockfile as `hoisted` mode. Developers can use either mode without changing the lockfile. This makes `isolated-mode` backward compatible and easy to implement. The implementation of `isolated-mode` will reside in the reification, it will take the ideal-tree and apply it on disk using symlink instead of hoisting.
```
ljharb, line 302: What happens if i use one reification, and then run npm install with another? Will it be able to trivially go back and forth without incident?

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+  - answer: junctions by default. support for symlinks can be added later as an opt-in if we see value for it.
+
+- How much community code will break when the system forbids access to undeclared dependencies? In other words, how much code needs to be fixed to work properly in `isolated-mode`?
+  - Regarding packages with missing dependencies in the `package.json` file. In many cases, package owners are willing to fix such issues. If there are a significant number of issues encountered around this, we can likely expect the of packages with missing dependencies is expected to drop fast after npm release `isolated-mode`. We may want to later add a feature to npm which allows users to locally declare dependencies on behalf of packages as a stop-gap, if existing solutions to this are not enough.
+  - There don't seem to be any packages out there that depend on the way hoisting works as a feature. If this is the case, it would be easy to argue that these packages should only rely on the npm and the Node.js contracts/APIs and not on implementation details like hoisting.
```
ljharb, line 343: This remains inaccurate; there are such packages.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+  - Some dev-environments don't support symlinks.
+    - AWS Lambdas -> a repository can a be installed with isolated mode locally and on CI but then deployed in hoisted mode.
+    - React native -> There are plugins existing to make react-native work with symlinks
+  - If a package is missing a dependency, it can be temporarily fixed (while waiting for the package owner to fix this issue) by declaring this missing dependency as top level dependency of the repository.
```
ljharb, line 347:
```suggestion
  - If a package is missing a dependency, it can be temporarily fixed (while waiting for the package owner to fix this issue, or if it’s working as designed) by declaring this missing dependency as top level dependency of the repository.
```

accepted/0042-isolated-mode.md
VincentBailly: Static analysis tools are not enough guarantee to be able to have a build system reliably do caching  (not re-building everything when only a small part of the repo changes). This is because a dependency of one workspace can break another unrelated workspace. Static code analysis does not remove this risk.

ljharb: In my experience, static code analysis tools are sufficient to force all dependencies to be explicitly declared, in large as well as small projects, which seems like it obviates the problem you're describing.

accepted/0042-isolated-mode.md
VincentBailly: it should be able to do the right thing, meaning change from hoisted to isolated and back with no leftover from the previous installation.

accepted/0042-isolated-mode.md
VincentBailly: one example that we hit is a workspace adding a dependency to `@types/jasmine` and that breaks all the other workspaces which use `@types/jest`. Because `@types/jasmine` will be hoisted and included in the compilation of all the workspaces, making them fail. This cannot be caught by code analysis tools because the dependencies are actually correctly declared.

accepted/0042-isolated-mode.md
VincentBailly: the doc changed, this is the new link, I will update it, in the meantime the link is this: https://nodejs.org/api/modules.html#modules_package_manager_tips

ljharb: oof, ok thanks - it even suggests using NODE_PATH, which is SUPER deprecated and not supported at all in ESM.

accepted/0042-isolated-mode.md
ljharb: ok, that makes sense - maybe instead of "they are not", a more nuanced sentence could be used, since it's really not the size of the project that causes the issue, it's "there's more than one child workspace".

It also might be unique to packages with TS types, or to these specific packages which have colliding type definitions - do you have any examples of this happening with JS?

VincentBailly: another example:
- `workspace 1` depends on `foo`
- `foo` requires `yargs` but does not declare it as a dependency
- `workspace 2` depends on `yargs`
I all works fine, foo can require `yargs` because it is hoisted.

- now `workspace 2` get rid of `yargs`
- `workspace 1` is broken because `foo` cannot import `yargs` anymore

In this situation all the dependencies are correctly declared in my repo, so it would not be caught by code analysis.

VincentBailly: > maybe instead of "they are not", a more nuanced sentence could be used, since it's really not the size of the project that causes the issue, it's "there's more than one child workspace".

It is not limited to repositories using workspaces, in the example above, if you replace `workspace 2` by an external package called `bar` and you make `workspace 1` take a dependency on `bar` you get the same issue.

It is actually not limited to the size of the repo, but the likelihood of this being an issue increases with the complexity of the dependency graph, to a point a which it becomes unmanageable passed a certain scale (which we have passed in internal repos in Microsoft).

accepted/0042-isolated-mode.md
VincentBailly: While the NODE_PATH advice may be outdated, the advice related symlinks is probably not, the reasons for writing this in the NodeJs doc (presumably) still stand.

ljharb: I suspect the reasons are that the author preferred this approach, and nobody reviewing the PR noticed that it’d be inappropriate for node to make such statements ¯\\\_(ツ)\_/¯

accepted/0042-isolated-mode.md
ljharb: Sure - but the likelihood isn’t zero in a small repo either, so the size of the repo doesn’t flip a binary, it instead increases a probability - and it’d be great for the prose here to reflect that.

accepted/0042-isolated-mode.md
VincentBailly: A package can be installed more than once in the store, this happens when a package has a peer dependency and is depended upon by two packages which provide a different version of this peer dependency (see one example in the example section). In this scenario, the folder of the package in the store should be different for the two different installations of that package. These packages will have the same content but different node_module folder, using a hash containing the dependencies of a package will make sure that these packages will not have a name collision.

accepted/0042-isolated-mode.md
VincentBailly: @ljharb can you clarify what would be worth solving for both modes?

ljharb: In the default mode, could you not still use symlinking to share package versions across projects and save disk space?

accepted/0042-isolated-mode.md
ljharb: If it’s the same version, then it has the same dependencies (since that’s the “dependencies” field in package.json), although it might have different dependents. If you mean it’s a hash of the node_modules folder, which might differ, that makes sense.

accepted/0042-isolated-mode.md
VincentBailly: It is difficult to use symlinks without fully implementing a store like proposed in this RFC. The reason is that when you require() or require.resolve(), you always get back a real path, so you look the context of which you come from (unless you have that context encoded in the store like this RFC suggests).

Another option for using symlinks is to use the preserve symlink node option, but it comes with bad implications like bad perf because a module can be instanciated many times.
 https://nodejs.org/api/cli.html#cli_preserve_symlinks

accepted/0042-isolated-mode.md
VincentBailly: In "dependencies" I really meant "dependencies+peerDependencies" but we really only need peerDependencies, I will change that.

VincentBailly: updated now, does it make more sense now?

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
+# Isolated mode
+
+## Summary
+
+This RFC is a proposal to add a new opt-in installation mode.
+
+This mode aims at preventing workspaces from over-sharing their dependencies. With this mode, two workspaces will share a dependency only if both have declared it in their package.json.
+
+This RFC aims at improving projects with workspaces, but the suggested implementation will also happen to bring value to projects not using workspaces. Because `isolated-mode` is valuable for any project, it will also be available for projects not using workspaces.
```
fritzy, line 9: This is bit repetitive.

While this RFC aims at improving projects with workspaces, `isolated-mode` could be valuable for any project, whether or not it uses workspaces.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Motivation
+
+The introduction of workspaces in npm v7 brought the ability to split large code-bases into smaller pieces of code with declared dependencies between them.
+
+Several build tools took advantage of these declared dependencies to split the builds into multiple steps (one per workspace). These build tools take advantage of the declared dependencies between workspaces to only rebuild the workspaces that have not changed since the last build. With the current way npm works, any dependency change to any workspace may affect any other workspace. This means that the build systems need to rebuild the entire project every time a single workspace modifies a dependency. This is inefficient and does not scale.
+
```
fritzy, line 16: We should mention that workspaces in npm v7 facilitate the "monorepo" pattern (since you mention it later).

The motivations are:
* monorepo performance
* monorepo dependency isolation (prevent side effects when changing dependencies of one workspace to another)

I think this could be a bit more direct and clearer in this section specifically.

fritzy (commented):
This is very good. I just added some comments to improve clarity and brevity.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+## More on the problem
+
+Packages and workspaces declare their dependencies to npm by using special fields in their `package.json`. The Node.js runtime is unaware of these special fields, instead it uses a [module resolution algorithm](https://nodejs.org/api/modules.html) to determine the meaning of a module importing another module. npm bridges this gap by converting the declared dependency graph into a folder structure which makes sense to Node.js. This operation is call `reification`.
+
+When converting the dependency graph to a folder structure, `hoisted-mode` loses information. Multiple different dependency graphs can be converted to an identical folder structure. Since the folder structure is all that Node.js uses to resolve modules, Node.js misses some information about dependencies. This loss of information is most frequently visible through the fact that workspaces are able to successfully import the dependencies declared by other workspaces.
```
isaacs, line 28:
```suggestion
When converting the dependency graph to a folder structure, `hoisted-mode` loses information. Multiple different dependency graphs can be converted to an identical folder structure. Since the folder structure is all that Node.js uses to resolve modules, Node.js misses some information about dependencies. This loss of information is most frequently visible through the fact that packages in the tree are able to successfully import packages that they have not explicitly declared as dependencies.
```

accepted/0042-isolated-mode.md
isaacs: We do use symlinks in some cases to resolve otherwise impossible scenarios.  For example, in the "pathological nesting loop" scenario, we use symlinks to avoid an infinite regress.  The simplest formulation of that scenario is:

```
x@1 -> (y@1)
x@2 -> (y@2)
y@1 -> (x@2)
y@2 -> (x@1)
```

When reifying, the first time we notice that a package is descended from a package of the same name that does not match the dep being installed, which is in turn descended from a package of the same name which _does_ match the dep being installed, we place a symbolic link instead of unpacking it a second time.

So instead of this:

```
root
+-- x@1
+-- y@1
    +-- x@2
    +-- y@2
        +-- x@1
        +-- y@1
            +-- x@2 ...
```

we have this instead:

```
root
+-- x@1
+-- y@1
    +-- x@2
    +-- y@2
        +-- x@1
        +-- y -> root/node_modules/y
```

I have explored a "dedupe harder" mode.  In this mode, we walk the nodes in the tree, and any that match another instance in the tree are replaced with a symbolic link to the other copy.  It's not very complicated code to deal with, but it doesn't save much in disk space or install time in the experiments we ran, and it resulted in a package tree with symbolic links that were somewhat haphazard for a human who might go poking around in node_modules.

It may be worth exploring again in the context of a massive monorepo, if we want to get the benefits of aggressive deduplication _without_ dependency isolation.

For that reason, I'd suggest the opt-in here be a `--mode=(default|isolated|...)` rather than a boolean, so that we can expand the set of reification modes in the future if we choose.

isaacs: In fact, we're likely to eventually provide a `--mode=importmap` reification strategy, which would potentially deliver the value of `--mode=isolated`, but without any symlinks at all.

accepted/0042-isolated-mode.md
isaacs: I can confidently report that yes, the author preferred that approach. :joy:

https://github.com/nodejs/node/commit/ff39d413a3c1#diff-5ae798d70a8dbd897d490bd7b453d5731ad3b49e416def53ee2ef0d11ca320b3R274

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+- The symlinks in a `node_modules` folder correspond to the matching workspace's dependencies.
+- Each symlink is given the same name as the dependency it represents.
+- Each symlink points to a location in the package store where the corresponding package is installed.
+
+In the package store:
+
+- The package store is a folder stored in the root/project level `node_modules` folder.
+- The package store is named `.npm`.
+- The package store contains folders, one for each dependency installed. These dependencies can be direct workspace dependencies or transitive dependencies.
+- These folders have a name containing the name of the package, its version and a hash of its content plus the hash of its resolved peer-dependencies.
```
isaacs, line 97:
```suggestion
- These folders have a name containing the name of the package, its version and a hash of its content plus the hash of its resolved peer dependency set.
```

isaacs: With a dependency graph like:

```
x ->PEER(y)
y ->PEER(z)
z ->()
```

You effectively have to treat _both_ `y` and `z` as the "peer set" of `x`, because if `x` depends on something with a conflicting peerDep on a different version of `z`, then that package cannot be installed alongside `y`.

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Compatibility
+
+### Different versions of npm
+
+The `isolated-mode` will use the same lockfile as `hoisted` mode. Developers can use either mode without changing the lockfile. This makes `isolated-mode` backward compatible and easy to implement. The implementation of `isolated-mode` will reside in the reification, it will take the ideal-tree and apply it on disk using symlink instead of hoisting.
+
+### Projects not using workspaces
+
+Non-workspace projects _will_ be able to use the `isolated-mode` reification. This is useful as a library author, using `isolated-mode` gives me the guarantee that my library will work as intended when consumed by projects which use `isolated-mode`.
```
isaacs, line 306:
```suggestion
Non-workspace projects _will_ be able to use the `isolated-mode` reification. This is useful as a library author, using `isolated-mode` gives me the guarantee that my library will work as intended, without unexpected run-time errors caused by undeclared dependencies.
```

accepted/0042-isolated-mode.md
```diff
@@ -0,0 +1,347 @@
...
+- The folder representing the package itself contains the code of that package. 
+- The folders representation the dependencies of the package are symlinks to the corresponding package in the store.
+- The dependencies of each packages are installed exactly the same way as workspaces' dependencies are installed.
+
+This structure may seem more complicated than necessary. The reasons for this complexity are:
+
+- to make it possible for a package to import itself using its own name. For example, the following code `require.resolve("foo");` in `node_modules/.npm/foo@1.0.0-1234/node_modules/foo/bar.js` will correctly return `node_modules/.npm/foo@1.0.0-1234/node_modules/foo/index.js`.
+- to prevent creating circular symlinks, even when packages have circular dependencies.
+
+#### Peer dependencies
```
isaacs, line 110: I think we can probably avoid a lot of this peerDep/edge case analysis if we specify that we are going to implement the isolation mode as a post-buildIdealTree conversion step.

That is, even though there are dependency graph shapes that may be enabled by isolated-mode, which are not enabled by hoisted mode, some some of these may be desirable, we are going to instead favor interoperability and reducing the degree of change in the dependency contracts enforced by npm.

So, whereas hoisted mode does the following:

1. Calculate hoisted ideal tree to satisfy dependency graph.
2. Reify ideal tree.
3. Save ideal tree to package-lock.json

instead, npm will:

1. Calculate hoisted ideal tree to satisfy dependency graph.
2. Capture hoisted tree lockfile metadata.
3. Convert hoisted tree into isolated tree.
4. Reify isolated tree.
5. Save hoisted tree lockfile metadata to package-lock.json

Thus, any situation not currently resolvable by hoisted-mode will not be possible with isolated-mode.  However, this ensures that any project may be reified using _either_ hoisted or isolated mode, or switched between them, without any changes to the project lockfile.

Anything that causes an `ERESOLVE` error in hoisted mode will also cause an `ERESOLVE` error in isolated mode.

Then the task is to document how that conversion process is accomplished.  Imo, that ought to end up in this RFC document, even if it happens after some implementation work has been done (which is probably a good idea, since we're bound to find some edge cases we haven't thought of!)

accepted/0042-isolated-mode.md
fritzy: I added some bikeshedding comments on this.

isaacs:
Landed (with some formatting and our planned implementation approach within `@npmcli/arborist`) on 88542dff62b450fba8833ce895606cc01716839f.

Thank you all for getting this proposal to this point, _most_ especially @VincentBailly, but also everyone else who contributed edits and discussion.  This is a big one, and I think we whittled it down to the point where it provides the key pieces of value intended, while also being straightforward and feasible to implement, with a minimum of disruption for our users.  I'm looking forward to getting this shipped!

PR #437, opened 2021-08-23
RFC: Robust Lifecycle Scripts

fritzy:
Proposal to implement more robust lifecycle scripts with reify context.

See issue npm/cli#3042

### DIFF
--- a/accepted/0000-robust-lifecycle-scripts.md
+++ b/accepted/0000-robust-lifecycle-scripts.md
@@ -1,0 +1,71 @@
+# {{TITLE: a human-readable title for this RFC!}}
+
+## Summary
+
+Lifecycle scripts need to have more context passed to them about why an event is being invoked. Due to the way packages are deduped and hoisted, not all use cases are properly covered by simply event names.
+
+## Motivation
+
+Lifecycle scripts such as `preinstall` `install` and `postinstall` are very commonly used, but difficult to reason about due to the way package trees are installed and updated. In the case of `install`, a package could be directly installed as a dependency by a user, installed as a global, indirectly installed as a dependency, or added when dependency versions changed to the point where a dependency could no longer be shared between two packages as a single version. Worst still, `uninstall` could be fired when the package won't be removed at all due to one dependant no longer requiring it, but another still needing to keep it.
+
+An `uninstall` script could assume that the directory will be removed or not used in the future, and put the package into a bad state if it was merely deduped, for example. By providing scripts with more context, like whether a new directory is being made or destroyed, which depenants caused the installation, etc, we could properly cover all use-cases and remove confusion that would lead to destroyed data, bugs, and frustration.
+
+Future versions of npm could use different reify engines and methods, potentially leading to more needed context for these scripts.
+
+The primary reason for this RFC is the lack of `uninstall` lifecycle scripts in npm v7, and how difficult it would be to re-add them in a clear way.
+
+See issue: #3042
+
+## Detailed Explanation
+
+Lifecycle scripts need to be re-engineered and the old approach should be deprecated.
+
+`package.json` will include a new index called `"lifecycle"`
+
+```js
+{
+  "lifecycle": {
+    "install": {
+      "pre" "bin/life_install.sh --phase=pre ${dependant} ${depenant_path} ${reify-reason}"
+      "pre" "bin/life_install.sh --phase=current ${dependant} ${reify-reason}"
+      "pre" "bin/life_install.sh --phase=post ${dependant} ${reify-reason}"
+    },
+    "publish": {
+      "pre": "prepublish" // default
+    }
+  }
+}
+```
+
+By default, the lifecycle script sections will reference the old reserved scripts, maintaining compatibility.
+
+## Rationale and Alternatives
+
+We could solve this problem by simply adding more lifecycle scripts, to an extent.
+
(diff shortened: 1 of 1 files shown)


accepted/0000-robust-lifecycle-scripts.md
```diff
@@ -0,0 +1,71 @@
...
+## Detailed Explanation
+
+Lifecycle scripts need to be re-engineered and the old approach should be deprecated.
+
+`package.json` will include a new index called `"lifecycle"`
+
+```js
+{
+  "lifecycle": {
+    "install": {
```
wraithgar, line 28: I think having discrete points in the reify process is better than trying to support the pre/post paradigm like this.  It was an inadequate model and confuses people a lot (i.e. "pre" means "it runs before the x scripts" not "it runs before x")

I'd prefer not having pre/post be baked in like this. If we have a need for something to run at a time, we put an event there and name it accordingly.

accepted/0000-robust-lifecycle-scripts.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+Lifecycle scripts such as `preinstall` `install` and `postinstall` are very commonly used, but difficult to reason about due to the way package trees are installed and updated. In the case of `install`, a package could be directly installed as a dependency by a user, installed as a global, indirectly installed as a dependency, or added when dependency versions changed to the point where a dependency could no longer be shared between two packages as a single version. Worst still, `uninstall` could be fired when the package won't be removed at all due to one dependant no longer requiring it, but another still needing to keep it.
+
+An `uninstall` script could assume that the directory will be removed or not used in the future, and put the package into a bad state if it was merely deduped, for example. By providing scripts with more context, like whether a new directory is being made or destroyed, which depenants caused the installation, etc, we could properly cover all use-cases and remove confusion that would lead to destroyed data, bugs, and frustration.
```
isaacs, line 11:
```suggestion
An `uninstall` script could assume that the directory will be removed or not used in the future, and put the package into a bad state if it was merely deduped, for example. By providing scripts with more context, like whether a new directory is being made or destroyed, which dependents caused the installation, etc, we could properly cover all use-cases and remove confusion that would lead to destroyed data, bugs, and frustration.
```

wraithgar:
Something that was talked about in previous discussions about this was having some way to at least give people a fighting chance of making this an iterative change.  Namely, that these could reference scripts.  I don't think that's really the right solution but we should definitely keep that on our radar as we build examples.   For instance making the lifecycle script entries call `npm run xxx`.

This does bring up the point of if these should be exclusive. i.e. if this object is populated at all, then the original lifecycle scripts are not ran.

accepted/0000-robust-lifecycle-scripts.md
```diff
@@ -0,0 +1,71 @@
...
+
+`package.json` will include a new index called `"lifecycle"`
+
+```js
+{
+  "lifecycle": {
+    "install": {
+      "pre" "bin/life_install.sh --phase=pre ${dependant} ${depenant_path} ${reify-reason}"
+      "pre" "bin/life_install.sh --phase=current ${dependant} ${reify-reason}"
+      "pre" "bin/life_install.sh --phase=post ${dependant} ${reify-reason}"
```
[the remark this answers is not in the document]
isaacs: If this is running during the reserved phase, then we're in trouble.

Before diffing, we don't know what will actually be done.  Before unpacking, we don't have the scripts on disk to run.  After unpacking, it's arguably not "pre"-install.  But a userland script can modify the folder tree (or even the dependency graph, if it writes to package.json files), so it invalidates the work already done.  We'd have to keep reloading the trees and recalculating the diff if any changes were made.  If we _don't_ do that, then we have security problems.  Better to just keep the "pre" to "here's what we were asked to do, we are about to start getting ready to figure out how to do it", and the "post" is "here's what we did, and it's all the way done now".

So something like:

```json
{
  "lifecycle": {
    "beforeTreeChange": "_loadTrees has not yet been called",
    "afterTreeChange": "_reifyPackages has completed",
    "beforeSave": "have not yet written to package.json and package-lock.json",
    "afterSave": "completely wrote lockfile and manifest"
  }
}
```

If `beforeTreeChange` and `afterTreeChange` exposed the various arguments that we currently pass into `arborist.reify()` (add, rm, update, etc.) then people could use that to decide whether this is an uninstall they care about.  If `afterTreeChange` also had a way to get the diff somehow (though I'm not sure the best way to do this), then they could likewise pick the replacement for `scripts.uninstall` that most suits their need.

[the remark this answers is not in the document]
fritzy: It seems like we'd be exposing implementation pretty directly, which is subject to change. For example, a module-map based solution might not use trees at all. Is there a way we could be more future-proof and still be accurate about the lifecycle? Do package maintainers need all of that granularity?

accepted/0000-robust-lifecycle-scripts.md
ljharb: imo "it runs before x" is the semantic every single person wants, and the other semantic is "silly sugar for `&&`".

bnb:
Would this RFC be a reasonable one to also include enhancements to the lifecycle flags, or should that be a separate concern?

accepted/0000-robust-lifecycle-scripts.md
Jirin32: >

Jirin32 (approved):

PR #486, opened 2021-11-03
Resolved registry overrides

everett1992:
Create a mechanism that allows users to create package-locks that can be used with different custom registries.

[link to rfc](https://github.com/everett1992/rfcs/blob/main/accepted/0000-resolved-registry-override.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "templateOSS": {
     "//@npmcli/template-oss": "This file is partially managed by @npmcli/template-oss. Edits may be overwritten.",
-    "version": "4.29.0",
+    "version": "4.30.0",
     "windowsCI": false,
     "macCI": false,
     "allowPaths": [
@@ -56,6 +56,6 @@
   },
   "devDependencies": {
     "@npmcli/eslint-config": "^6.0.0",
-    "@npmcli/template-oss": "4.29.0"
+    "@npmcli/template-oss": "4.30.0"
   }
 }


everett1992:
We can take this off the agenda tho I'd appreciate implementation input from someone familiar with package-lock and arborist.

I think `metaFromNode` in lib/shrinkwrap is responsible for getting data to write to a lockfile.
https://github.com/npm/arborist/blob/main/lib/shrinkwrap.js#L301-L304

How do I tell if a node was resolved from a registry or a literal http dependency?

everett1992:
I've drafted this RFC in npm/arborist#352. Has there been more discussion including registry in lockfile? I think omitting resolved is the most comprehensive option because it supports switching between registries that host tarballs at different paths. But that makes it also more expensive than other options because npm must re-resolve the tarball url from package manifest. So I'm considering alternatives

Maybe npm could guess that a dependency without `resolved` is hosted at `<registry>/<name>/-/<name>-<version>.tgz`, tho that would be even slower when tarballs are hosted at a different path.

Maybe we implement this RFC and an option that supports switching between registries that host tarballs at the same path without paying the resolve cost. @isaacs mentioned wanting to use a sigil to mean 'the current registry', such as `resolved: ${registry}/npm/-/npm-8.1.0.tgz` but had issues implementing that without requiring a new lockfile version or affecting other npm clients. But we already have a sigil that means 'the current registry' - `https://registry.npmjs.org`! We could record registry.npmjs.org as the resolved url even when using a different registry. Then when npm reads the package-lock it will replace registry.npmjs.org with the current registry.

We need to replace the registry part of the resolved url with `https://registry.npmjs.org` but arborist doesn't record the registries it used. Arborist knows the resolved url of a node, but not where the registry ends and the path begins.

> `http://private.npmjs.org/a/b/c.tgz` could be split more than one way
> | registry | path | recorded |
> | ------------------------------------|--------|--|
> | `http://private.npmjs.org/a/b` | `c.tgz` | `https://registry.npmjs.org/c.tgz` |
> | `http://private.npmjs.org/a` | `a/b/c.tgz` | `https://registry.npmjs.org/a/b/c.tgz` |

We could work backward from the package name and registry config. We need the name because scopes can use different registries. `Node#name` may be the folder name, an alias or won't include the scope. `Node#pkg.name` is not guaranteed to be set. I think we'd have to use edgesIn specs or read package.json.

everett1992:
@darcyclarke would you add this to the agenda for the next RFC meeting? The [PR](https://github.com/npm/cli/pull/4264) implementing this RFC has a Needs Discussion tag.

PR #488, opened 2021-11-05
Make npm install scripts opt-in

tolmasky:
Install scripts that can run just about anything by default pose some pretty serious security considerations, and these are inreasingly moving out of the theoretical realm and becoming actively exploited. See for example here: https://therecord.media/malware-found-in-coa-and-rc-two-npm-packages-with-23m-weekly-downloads/.

At the same time, we have developed better techniques for many of the most common use cases for install scripts in the many years since npm originally included then. In particular, [N-API](https://nodejs.org/api/n-api.html) offers a compelling alternative to binary packages that are built on the users' computer. However, even before this, many packages are choosing to just pre-build for multiple platforms ahead of time to handle most of the common installation targets and make the install process easier on the user in general.

Instead of by default always running the install scripts (`preinstall`, `install`, `postinstall`, `prepublish`, `preprepare`, `prepare`, `postprepare`) if they are present during the install process, provide flags to require users to explicitly allow them to run, either whoelsale as "one big switch", or on a package by package (and optionally version by version) basis. Also provide matching `npm config` options to do the same globally and permanently instead of on every install.

### DIFF
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -1,0 +1,49 @@
+# Transition to a model where install scripts must be explicitly allowed during the install process.
+
+## Summary
+
+Instead of by default always running the install scripts (`preinstall`, `install`, `postinstall`, `prepublish`, `preprepare`, `prepare`, `postprepare`) if they are present during the install process, provide flags to require users to explicitly allow them to run, either whoelsale as "one big swtich", or on a package by package (and optionally version by version) basis. Also provide matching `npm config` options to do the same globally and permanently instead of on every install.
+
+## Motivation
+
+Install scripts that can run just about anything by default pose some pretty serious security considerations, and these are inreasingly moving out of the theoretical realm and becoming actively exploited. See for example here: https://therecord.media/malware-found-in-coa-and-rc-two-npm-packages-with-23m-weekly-downloads/.
+
+At the same time, we have developed better techniques for many of the most common use cases for install scripts in the many years since npm originally included then. In particular, [N-API](https://nodejs.org/api/n-api.html) offers a compelling alternative to binary packages that are built on the users' computer. However, even before this, many packages are choosing to just pre-build for multiple platforms ahead of time to handle most of the common installation targets and make the install process easier on the user in general.
+
+## Detailed Explanation
+
+I propose the following flags and changes in behavior:
+
+### npm install
+1. By default, no scripts are run during the `npm install` lifecycle: https://docs.npmjs.com/cli/v7/using-npm/scripts#npm-install
+2. The `--allow-scripts` flag is added to `npm install` which allows users to enable scripts for a specific package at a specific version, which is the only truly safe thing combination (especially given a `package-lock.json` file that is ensuring the integrity of the tarball in question):
+    - `npm install --allow-scripts canvas@1.0.0`: Install the contents of `package.json`, only allowing `canvas` version 1.0.0 to run install scripts. Perhaps it should warn if no install scripts are present?
+    - `npm install canvas@1.0.0 --allow-scripts canvas@1.0.0`: Installs `canvas@1.0.0` to the current `package.json`, and allows it to run install scripts. Perhaps it should warn if no install scripts are present?
+    - `npm install canvas@1.0.0 --allow-scripts canvas@1.0.0 --allow-script ramda@2.0.0`: Installs `canvas@1.0.0` to the current `package.json`, and allows it to run install scripts, as well as allowing ramda@2.0.0 to run install scripts. Perhaps it should warn if no install scripts are present?
+3. The `--disallow-scripts` option should also be added that does the opposite of `--allow-scripts`. This is necessary since later I will discuss global settings for allowing packages to use install scripts, and so this lets you ignore that setting if you so choose.
+4. The `--allow-scripts-unsafe` option is added for more convenient, but more dangerous, handling of install scripts:
+    - `npm install --allow-scripts-unsafe canvas`: Install the contents of `package.json`, allowing any version of `canvas` to run install scripts.
+    - `npm install --allow-scripts-unsafe "canvas@1.0.0-2.0.0"`: Install the contents of `package.json`, allowing any version of `canvas` in the semver range 1.0.0 to 2.0.0 to run install scripts. While strictly safer than the unranged version above, it still runs the risk of new vulnerable packages being snuck in in that range. For example, if an account is compromised, and a patch version is published to every major version that is present. 
+    - `npm install canvas --allow-scripts-unsafe canvas`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts.
+    - `npm install canvas --allow-scripts-unsafe canvas --allow-unsafe-scripts ramda`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts for canvas and any version of ramda it encounters.
+
+### npm config
+We should add the ability to configure npm globally to allow packages to make use of scripts. This can be useful for packages you install a lot, or for packages you are working on:
+
+1. `npm config set allow-scripts-canvas@1.0.0 true`: This will allow canvas version 1.0.0 to install scripts whenever `npm install` is used.
+2. `npm config set allow-scripts-unsafe-canvas true`: This will allow any version of canvas to install scripts whenever `npm install` is used. It should also probably warn on each case.
+
+## Rationale and Alternatives
+
+The vast majority of packages do not use install scripts, and they provide an easy vector to immediately affect thousnads of users if you are able to take control of a package author's account. While not a perfect solution, this would dramatically reduce that capability, and is probably the correct default instead of allowing any 20-levels-deep subdependency to run any artbitrary script on your system.
+
+## Implementation
+
+I imagine this should hopefully not be harder than adding some if statements around the right lifecycle sections of install.
+
+## Unresolved Questions and Bikeshedding
+
+Two open questions for me:
+
+1. Having npm ship with a grandfathered list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags. 
+2. I think the automatic node-gyp stuff should be under this umbrella as well. That is to say, node-gyp would only run when gypfiles are detected if the `allow-scripts` flag allows it, since you can also techncially put whatever you want in there I believe, and not doing this would simply mean that any malicious behavior would be transitioned there.
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -15,10 +15,10 @@ From our calculations, only around 0.6% of packages on npm currently use this fe
 
 |     | Count | Percentage |
 | --- | --- | --- |
-| Packages with preinstall script | 2,826 | 0.14% |
-| Packages with install script | 2,138 | 0.11% |
-| Packages with postinstall script | 7,155 | 0.36% |
-| Packages with gyp | 2,769 | 0.14% |
+| Packages with preinstall script | 1,660 | 0.08% |
+| Packages with install script | 2,077 | 0.1% |
+| Packages with postinstall script | 6,989| 0.35% |
+| Packages with gyp | 2,744 | 0.14% |
 | Total Packages on npm | 1,996,434 | 100% |
 | **Total Packages with scripts on npm** | 12,035 | **0.6%** |
 
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -50,7 +50,7 @@ I propose the following flags and changes in behavior:
 3. The `--disallow-scripts` option should also be added that does the opposite of `--allow-scripts`. This is necessary since later I will discuss global settings for allowing packages to use install scripts, and so this lets you ignore that setting if you so choose.
 4. The `--allow-scripts-unsafe` option is added for more convenient, but more dangerous, handling of install scripts:
     - `npm install --allow-scripts-unsafe canvas`: Install the contents of `package.json`, allowing any version of `canvas` to run install scripts.
-    - `npm install --allow-scripts-unsafe "canvas@1.0.0-2.0.0"`: Install the contents of `package.json`, allowing any version of `canvas` in the semver range 1.0.0 to 2.0.0 to run install scripts. While strictly safer than the unranged version above, it still runs the risk of new vulnerable packages being snuck in in that range. For example, if an account is compromised, and a patch version is published to every major version that is present. 
+    - `npm install --allow-scripts-unsafe "canvas@1.0.0-2.0.0"`: Install the contents of `package.json`, allowing any version of `canvas` in the semver range 1.0.0 to 2.0.0 to run install scripts. While strictly safer than the unranged version above, it still runs the risk of new vulnerable packages being snuck in in that range. For example, if an account is compromised, and a patch version is published to every major version that is present.
     - `npm install canvas --allow-scripts-unsafe canvas`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts.
     - `npm install canvas --allow-scripts-unsafe canvas --allow-unsafe-scripts ramda`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts for canvas and any version of ramda it encounters.
 
@@ -72,5 +72,5 @@ I imagine this should hopefully not be harder than adding some if statements aro
 
 Two open questions for me:
 
-1. Having npm ship with a grandfathered list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags. 
+1. Having npm ship with a legacy list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags.
 2. I think the automatic node-gyp stuff should be under this umbrella as well. That is to say, node-gyp would only run when gypfiles are detected if the `allow-scripts` flag allows it, since you can also techncially put whatever you want in there I believe, and not doing this would simply mean that any malicious behavior would be transitioned there.
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -2,13 +2,41 @@
 
 ## Summary
 
-Instead of by default always running the install scripts (`preinstall`, `install`, `postinstall`, `prepublish`, `preprepare`, `prepare`, `postprepare`) if they are present during the install process, provide flags to require users to explicitly allow them to run, either whoelsale as "one big swtich", or on a package by package (and optionally version by version) basis. Also provide matching `npm config` options to do the same globally and permanently instead of on every install.
+Instead of by default always running the install scripts (`preinstall`, `install`, `postinstall`, and auto-detected `gyp` files) if they are present during the install process, provide flags to require users to explicitly allow them to run, either whoelsale as "one big swtich", or on a package by package (and optionally version by version) basis. Also provide matching `npm config` options to do the same globally and permanently instead of on every install.
+
 
 ## Motivation
 
 Install scripts that can run just about anything by default pose some pretty serious security considerations, and these are inreasingly moving out of the theoretical realm and becoming actively exploited. See for example here: https://therecord.media/malware-found-in-coa-and-rc-two-npm-packages-with-23m-weekly-downloads/.
 
 At the same time, we have developed better techniques for many of the most common use cases for install scripts in the many years since npm originally included then. In particular, [N-API](https://nodejs.org/api/n-api.html) offers a compelling alternative to binary packages that are built on the users' computer. However, even before this, many packages are choosing to just pre-build for multiple platforms ahead of time to handle most of the common installation targets and make the install process easier on the user in general.
+
+From our calculations, only around 0.6% of packages on npm currently use this feature (see table below), meaning it should hopefully both be possible to transition many of these packages away from relying on script, while simultaneously making this transition as frictionless as possible for users.
+
+|     | Count | Percentage |
+| --- | --- | --- |
+
+| Packages with preinstall script | 2,826 | 0.14% |
+| Packages with install script | 2,138 | 0.11% |
+| Packages with postinstall script | 7,155 | 0.36% |
+| Packages with gyp | 2,769 | 0.14% |
+| Total Packages on npm | 1,996,434 | 100% |
+| **Total Packages with scripts on npm** | 13,406 | **0.6%** |
+
+## What Makes Install Scripts Special vs. Other Security Issues?
+
+It's easy to dismiss the issue with install scripts due to the expectation that you will be running the code you download from npm anyways, but as it turns out this isn't necessarily the case, and the threat vector is considerably more subtle. Outlined below are a number of scenarios to consider that show how install scripts do in fact represent a separate security threat:
+
+
+1. If you are using npm for purely frontend work, install scripts open up a completely unnecessary attack surface. It might be easy to forget here in node land, but many teams that use completely different languages for their backend work only fall precisely in this category. It doesn't make sense to be exposing Ruby or Python devs to install scripts, when otherwise the code they'd install would never run on their own systems and be confined completely to the browser sandbox.
+
+2. Install scripts may run under different privileges than the node apps run in. Do a search for "--unsafe-perm" and you'll see that for better or worse this is a suggested "fix all" for various npm install woes, which thus exacerbates this problem.
+
+3. Install scripts run in different environments than node apps. As mentioned above, install scripts may be uniquely positioned to be the only attack vector on build machines, that otherwise wouldn't run code at all (some companies for example separate build machines and CI machines, etc.) This is made worse by the fact that it is not uncommon to have builds and CI take place automatically in response to a pull request, and thus all it takes to infect the machine is issuing a pull request with the package.
+
+4. Install scripts allow for "typo squatting" where you can register names close to popular packages such that if you accidentally type "npm install canvsa" the install script immediately kicks off before you can even realize and hit ctrl-c and its already spawned a process to start doing something malicious.
+
+5. There is also the fact that install scripts can be easier to introduce precisely because its not code that needs to be imported and run to attack. As such, it is possible to sneak a package into the package-lock.json as part of a larger change, since github will often automatically hide package-lock diffs, and no one really takes the time to read them since they're not really meant to be human-readable. Compare this with was malicious code that needs to be run, where you have a higher chance of noticing it in a diff since the actual code is where people usually pay attention when reviewing code and they'd ask "why are we importing this library we didn't need before?". Again, it doesn't make it impossible, it's just much easier with scripts.
 
 ## Detailed Explanation
 
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -20,7 +20,7 @@ From our calculations, only around 0.6% of packages on npm currently use this fe
 | Packages with postinstall script | 7,155 | 0.36% |
 | Packages with gyp | 2,769 | 0.14% |
 | Total Packages on npm | 1,996,434 | 100% |
-| **Total Packages with scripts on npm** | 13,406 | **0.6%** |
+| **Total Packages with scripts on npm** | 12,035 | **0.6%** |
 
 ## What Makes Install Scripts Special vs. Other Security Issues?
 
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -15,7 +15,6 @@ From our calculations, only around 0.6% of packages on npm currently use this fe
 
 |     | Count | Percentage |
 | --- | --- | --- |
-
 | Packages with preinstall script | 2,826 | 0.14% |
 | Packages with install script | 2,138 | 0.11% |
 | Packages with postinstall script | 7,155 | 0.36% |


WebReflection:
> Install scripts that can run just about anything by default pose some pretty serious security considerations

This is nothing new in the software world, to be honest, when you install pretty much any software in Windows it asks you if it's OK to give this software the rights to configure the OS on your behalf and everyone is like "*sure thing, whatever mate, let me move on*" ... and here there's no difference ... **except** we miss the *TRUST* around packages.

What is trust? The fact a package has every allowed publisher **behind 2FA** to start with, so that publishers are accountable.

Every example to date shows that everyone can steal and publish on behalf of everyone else any kind of crapware, and yet the proposal here is to "*provide yet another flag*"?

What is a flag good for, when it can become a configurable default? It would solve **nothing** around the proposed issue.

If there's anything *npm* should do, in my opinion, is to **promote 2FA modules** any way you can:

  * `npm info module` shows a special star, flag, check, about security/accountability because of 2FA
  * the site itself down-votes any search if the module is not 100% behind 2FA (or its developer/s)
  * a warning is shown on install ... **or** ... no pre/post install script is allowed *if* the module has not 100% 2FA publishers

Now this is thinking in a way that will stop causing the same accident over and over, where developers reputation also could play a role, so that **trust** become the way to move forward, as opposite of limiting all *npm* functiomalities because "*evil happens*".

Thanks for listening, and I hope yet another flag won't be the only move here 👋

tolmasky:
> This is nothing new in the software world, to be honest, when you install pretty much any software in Windows it asks you if it's OK to give this software the rights to configure the OS on your behalf and everyone is like "sure thing, whatever mate, let me move on" ... and here there's no difference ... except we miss the TRUST around packages.

This is actually a very different situation, as you are comparing the behavior of non-technical users installing non-technical software, vs. developers who do generally care (if allowed to) about the software they use. The other critical difference this analogy misses is that a very large portion of package installs are done not by humans, but in an automated fashion. So it is not a bunch of people saying "OK whatever", it is a fleet of CI machines repeatedly pulling packages and by default running scripts that could compromise them. A package going from not running an install script to running an install script between CI runs should IMO absolutely fail the test: there is a very bizarre change in side effects taking place, and test failures are an expected part of test suites (its the whole reason they exist), and at that point you can issue another commit that adds the explicit approval of the package, or, as with what happened earlier today, say "wait a minute, none of these should be installing", and avoid having a bunch of junk infect your CI machines.

Another important piece of context is that vast majority of people will actually see no change, as a tremendously small amount of packages on npm today use this feature. If this was some critical package.json field that 10% of packages used or something, then that would be one thing, but the reality is that when we last checked every package (literally every package on npm) at RunKit, the percentage was vanishingly small.

Additionally, many of the packages that *do* use this legitimately now have better alternatives. A clear example is transpilation that happens on install instead of on publish (which additionally means we're wasting a bunch of energy retranspiling the same code over and over instead of shipping the "built" package). As mentioned in the PR, N-API also provides a better solution to actual binary packages.

> What is a flag good for, when it can become a configurable default? It would solve nothing around the well known issue.

If you read the actual configuration options provided, you will see that there isn't a blanket option, only allowing you to globally allow "safely" a specific version of a package, and "unsafely" any version (or range) of a package. In other words, if you encounter that one random package that you know and love that has an install script, you can enable that one "from now on", which makes sense, you are saying "I've checked this package version, I'm good with it". This is very different than saying "I don't care about scripts running just do whatever with anything".

Thinking about the distinction between "automated" installs vs. "user drive installs" also informs the thinking around configuration vs. flags, and why bother are provided to some extent: it makes total sense that one would have a different risk model for an individual developer's computer vs. a CI or production machine. If a developer accidentally installs something that starts mining on their computer, at the end of the day that may be no big deal. However, that can be a huge profit loss if it manages to install on a production machine. So it makes a lot of sense to use the most explicit flags in the repo, but for individual users to be able to globally turn on certain packages for convenience.

Incidentally, you can even imagine a model where when running `npm install` interactively it can ask you about packages instead of failing (the same way apt asks you about disk usage), that way this process has even less friction.

> Every example to date shows that everyone can steal and publish on behalf of everyone else any kind of crapware, and yet the proposal here is to "provide yet another flag"?

The proposal is to make unexpected behavior default-off. It's like getting angry that browsers now require a iframe tags to explicitly enable certain features -- no, it should have always been that way, and the people working on it (developers, not end users) are perfectly equipped to understand and update their code. Except again, most people will not even need to use this flag, as the use legitimate use cases are small, and only growing smaller. Again, to the point where if this were implemented with a grace period to update packages, very quickly there would be very few packages that "need" this flag. I would argue that this would for example take much less time than the ESM transition for example. Also, as mentioned in the PR, if the flag is really that bothersome, you can globally allow packages to install.

> Now this is thinking in a way that will stop causing the same accident over and over, where developers reputation also could play a role, so that trust become the way to move forward, as opposite of limiting all npm functiomalities because "evil happens".

Trust doesn't fundamentally solve the problem of purposefully malicious actors. If someone steals your 2FA keys and publishes a malicious package, no amount of "ratings" or "stars" in npm info"can account for that, especially not *before hand*. There are two issues at play here: *accountability* (how to deal with things once they go wrong), and *security* (how to prevent things from going wrong in the first place). There doesn't exist an accountability system on Earth that would make me feel comfortable with giving browser JavaScript access to my filesystem by default. I want it sandboxed, no matter what other measures are taken.

WebReflection:
> you are comparing the behavior of non-technical users installing non-technical software, vs. developers who do generally care

besde the fact we don't have data around developers that really cares VS developers that copy and paste whatever they find on SO or elsewhere, this proposal would break [genuine software like this](https://github.com/WebReflection/geo2city#readme) which needs mandatory `postinstall` and everything not having it allowed by default would just break.

A breaking change like this in *npm* needs to take into accountability  all use cases to date (Electron based Apps too), not just CI, but I'll read the rest later, as I think the first line is already a no-go to me, sorry.

wesleytodd:
While this is a laudable goal, this is not the first time this has been brought up, and very little has changed now compared to before. This would be a MAJOR breaking change for the ecosystem and despite agreeing that opt-in would have been a better default, all it would do is shift the attack to the next easiest way to compromise the supply chain (`post-install` scripts currently being the easiest right now in the `npm` ecosystem). There are no easy fix-all's when it comes to running third-party code, and I am still not sold on the juice being worth the squeeze on this one.

IIRC, the last time this was brought up it was also brought up that the current config is a "all or nothing" thing (although I seem to recall some changes to that recently, so I might be wrong now). IMO, a much better starting point for this discussion would be incremental improvements which would not break the entire ecosystem all at once. I have not been following the RFC calls or PR's much recently, but I am sure someone who has could point to a few others in this space which have been discussed.

WebReflection:
I am proposing "**2FA or nothing**" thing instead, which seems way more sensible to me, still as breaking change.

Can anyone please read my suggestion [as an long time needed extra](https://github.com/npm/rfcs/pull/488#issuecomment-962100905), not just as a blocker? Thank you!

wesleytodd:
> I am proposing "2FA or nothing" thing instead, which seems way more sensible to me, still as breaking change.

Have you opened an RFC for this?  I would love to see a `--require-2fa` config which would fail if any packages were not published with 2FA as long as it had a good set of tools for ignoring things.  Seems like a better way to respond to a credential theft as is the source of these recent issues.

tolmasky:
> A breaking change like this in npm needs to take into account all use cases to date (Electron based Apps too), not just CI, but I'll read the rest later, as I think the first line is already a no-go to me, sorry.

That's certainly unfortunate, considering that had you read the RFC before this, it explicitly discusses a strategy of having a warning transition period to give everyone time to update (Electron apps can trivially take this time to add the flag and avoid any hiccups whatsoever, completely transparently to their users. I'm also not even sure this is an issue since Electron apps can ship "built" and don't have to necessarily themselves npm install on users' machines, but again, that would itself be mitigated completely with a one-line change). Anyways, your goal of showing me disrespect has succeeded, by declaring that a huge post that tried to in good faith address your comments isn't even worth reading since you didn't like the opening sentence for a reason that was covered in the original RFC. This however does not prevent you from continuing to criticize.

wesleytodd:
Hot button issues with folks passionate in the ecosystem can get tough to navigate, but I recommend taking a moment to listen. There will always be someone who disagrees in any sufficiently populated OSS community, and that is a good thing. In this case, I think this issue has made SOOO many rounds, and this solution has been recommended and talked through before, that it might lead to folks who have been involved in them in the past to jump to conclusions.

I think we can all agree that you did a great job writing up the RFC @tolmasky, and I highly doubt that going back in time and being able to change things before it was so entrenched, anyone would disagree with you. I would recommend letting folks comment on the RFC, and then attending the next RFC call. The folks who regularly attend are great, and really open to discussion.

peey:
> Install scripts that can run just about anything by default pose some pretty serious security considerations

Since this presupposes that the package you're installing might have malicious code, how do you propose to handle the threat posed by running your own trusted code which `import`s from these untrusted libraries?

tolmasky:
> While this is a laudable goal, this is not the first time this has been brought up, and very little has changed now compared to before. This would be a MAJOR breaking change for the ecosystem and despite agreeing that opt-in would have been a better default, all it would do is shift the attack to the next easiest way to compromise the supply chain (post-install scripts currently being the easiest right now in the npm ecosystem). There are no easy fix-all's when it comes to running third-party code, and I am still not sold on the juice being worth the squeeze on this one.

A number of things here, hopefully some of which ew at RunKit can be helpful with. We have to regularly analyze the usage of `install` scripts on RunKit, since we try to install every combination of every version of every package on npm, specially if we can join forces with npm's side of the data. I feel fairly confident that this can be done in a non "break the world" fashion, especially because I think many of the remedies can be handled by authors vs. the developers using the packages. So, in particular:

1. We should of course no matter what have a transition period where npm simply warns about install scripts.
2. I think there are also some "common sense" *user-driven* heuristics that could mitigate many of the problems completely. For example, if a `package-lock.json` is present then you can don't need to require the flag. This is because there is no danger of picking up a random patch update in this case, you are only installing items you've already installed in the past, and even checking against the checksum. If you wanted to be super safe, you could only apply this rule for package-lock.jsons generated before the "opt-in" transition (this can be determined by the presence of a property in the package lock) This means that you are truly catching the most egregious cases of installing unlocked packages that may pick up unintended script additions.
3. I think we could fairly reasonably establish the "affected radius" of users by combining download data and dependency-chain info. We could thus establish a fairly good metric of "safety" to reach before pulling the trigger on anything that would have any actual end-user affects.
4. During this period we can assist package authors in getting their packages updated. We've also done a lot of work in trying to group "install script usage", so we could fold in some "declarative" alternatives for a lot of the "piggy-backing off of install just because its the only way to do it" use cases. For example:
    - Providing an official way to ask for donations, so that doesn't have to go through install (I can't remember if this was already possible, apologies if this is already the case, but if so, then certainly letting people know that if in a year they don't switch to the official declarative support.md format or whatever their packages will require an install flag will probably go a long way to getting everyone transitioned).
    - Providing a system for declaring and installing "outside of npm" dependencies, either as a check, or as an install. We've actually implemented this ourselves for our own packages internally, something like "aptDependencies" and "aptRequirements", where npm is responsible for either making sure you have certain dependencies or fetching them. This would also just be a great ecosystem move to further "surface" the dependency structure of npm, and make it way easier for people to convey that their packages require certain tools to be available, etc.

With regard to the "bar" we're trying to reach with this strategy, I think the goal is of course not to fix everything, which isn't realistically possible. I do however think we can make reasonable progress to make "passive" attacks considerably harder to execute (where packages don't even need to be run for an attack to be successful).

madjam002:
Seems like a lot of work for the entire ecosystem when all malicious authors would have to do is just switch to running malware on first run via require/import instead which would probably happen as soon as this was implemented.

Perhaps a flag instead to refuse to install packages that have been published in a recent time period or at least flag them interactively? That way each individual user or organisation could configure their preference on how long packages have to have been in the wild before they even get downloaded onto their system.

rauchg:
Very supportive of this proposal. .*(install|prepare) scripts not only decrease security, they slow down installations, complicate caching and remove reproducibility by introducing side effects, and pose challenges for cross-platform compatibility.

These scripts can't be part of the "happy path" of npm. They need to be painful to use and consume and phased out of the default path of installation.

I thoroughly agree with the sentiment that the ecosystem having developed an excellent alternative for installing pre-built binary addons targeting multiple platforms means "this time is different" (I'm cognizant this has been proposed many times before)

peey:
> For example, if a package-lock.json is present then you can don't need to require the flag. This is because there is no danger of picking up a random patch update in this case, you are only installing items you've already installed in the past, and even checking against the checksum.

I like this idea. I think defaulting to trusting the scripts when a package-lock is present and not trusting them when it's absent is sensible.

@tolmasky I'd also encourage you to file a proposal with yarn and pnpm. npm has been conservative about accepting innovations unless they're already popularized by competing package managers.

tolmasky:
> Since this presupposes that the package you're installing might have malicious code, how do you propose to handle the threat posed by running your own trusted code which imports from these untrusted libraries?

This is a great question and I touched on it briefly in the response above, but I think it's worth addressing clearly, if only to establish expectations: this isn't meant to target those threats (but there are other ideas out there that do!). The goal of this RFC is explicitly to handle a unique class of "passive" threats that exist by simply being present in the dependency chain. This can make them particularly nefarious *precisely because* they don't need to be imported or run. This fact can make them easier to sneak into codebases (since people often don't review "metadata" files as thoroughly as code files, so the addition of an innocuous package may not raise any eyebrows, or the even harder to find two step approach of initially getting a "safe" package in, only to then modify *its* internal dependencies afterwards), and can also make these attacks "resistant" against mitigations targeting imported code, since something like code isolation won't help you if the problem is happening during the install. Ultimately, these different kinds of attacks will require different strategies, but I think the `install script` trick has made the barrier to entry to create these kinds of hard-to-spot "passive" attacks very low.

mattisdada:
I think this is a great proposal and have no real issue with it. But, an alternative that would move the burden from users back to NPM is that packages with install must be reviewed and approved by NPM for every publish and probably require an additional cost for said review and 2FA. 

I like this proposal more, but if people really don't want breaking changes, this would not cause any breaking changes for end users and motivate publishers to use alternative methods... Just a thought

dominykas:
I'm not sure I agree with the "next goal post" thinking. We do not have an easy way to evaluate package contents without installing them (side note: something to fetch the tarball without installing would be super useful as well). Yes, you can run every install with --ignore-scripts, but that's crap DX.

A blanket "always ignore _install_ scripts" as opposed to the currently available "always ignore _all_ scripts" (which does not allow anything npm run, last I checked) would be useful.

It doesn't have to start as opt-in. A more comfortable opt-out would be better than what we have now.

And yes, 2fa would be nice. I wish we could revive the staging discussions (#92, one approach for 2fa for automation). I also have https://github.com/nodejs/package-maintenance/pull/282 parked for a very long time.

Not going to hold my breath, though, see you back here in a year. Notice how it's always around budget planning season that these things are more common? Maybe someday the budget for securing things will _actually_ happen...

peey:
>  The goal of this RFC is explicitly to handle a unique class of "passive" threats that exist by simply being present in the dependency chain. This can make them particularly nefarious precisely because they don't need to be imported or run. ... This fact can make them easier to sneak into codebases (since people often don't review "metadata" files as thoroughly as code files,...

I see. If I understand correctly, the attack vector here is code which is contributed by an attacker who does not have commit rights to repo or publishing rights to the package, and then the contribution is merged by a maintainer. 

Do you have any recent examples of this so it's easier to understand how such an attack might be carried out? In the examples you have given (coa, rc) the attacker had access to package developer's account (publishing rights).

totten:
A slightly biased/tangential anecdote: PHP's `composer` went the other way from the start. They never allowed library-packages to run their own installation steps. (Only root-level/consumer-packages could define install steps.) It was a recurring request for a few years (eg https://github.com/composer/composer/issues/1193), and (from my POV, as a downstream dev) I felt it held `composer` back from doing some useful things. But I do appreciate that it reduces the attack-surface and has probably discouraged some supply-chain attacks.

Similar proposals were raised to allow installation logic with some kind of whitelist. The ship for that probably sailed a few years ago, but... eventually (*late last year*) I implemented a version of it. So you can see an example of such behavior [here](https://github.com/civicrm/composer-compile-plugin/blob/master/doc/site-build.md):
 
Screenshot

For installers/upgraders, it did introduce an extra prompt.  I was a little concerned about potential for user-pushback. (*Available signals suggest ~1000 of our deployments do `composer`-based administration.*) So far nobody has complained about the extra workflow step being onerous.

lewisrobbins:
@totten Doesn't that have the same problem as Windows' UAC prompt -- every single user just clicks 'yes' without a second's thought. Appreciate `composer` has more options that just `yes` and `no` but I think the point still applies.

MylesBorins:
Hey all. Myles from the npm team here. I've been trying to think through how we get to a place where we can disable install / postinstall scripts from running by default.

The biggest gap imho, as others have pointed out, is lazy loading platform specific binary assets. This is something I think we should figure out how to support in the client in a static and consistent manner.

I think with that behavior in place we could safely change defaults in a Semver Major.

I'm just stepping away for the weekend and plan to come back to this when back in the office next week.

It's exciting to see all the momentum for this type of change, even just 2 years ago I don't think we'd see such eide support for such a large breaking change

mirek:
Wouldn't signing npms on publish with 2FA protected allowed public keys solve this class of problems?

totten:
> Doesn't that have the same problem as Windows' UAC prompt -- every single user just clicks yes without a second's thought. Appreciate composer has more options that just yes and no but I think the point still applies.

Good point. I don't have data for a systemic answer. Personally, I find it analogous to [TOFU](https://en.wikipedia.org/wiki/Trust_on_first_use) in SSH. During the first invocation, you don't really know what you to expect, and anyone in a rush will blindly accept the suggestion. But then the message goes away (*decision is saved into JSON*); for most day-to-day tasks (`git pull && composer install` or `git clone && composer install` or even `composer update`), the validation is invisible. The message should only pop-up if something changes unexpectedly, eg

* (SSH) If a host starts advertising a different public-key, then that is suspicious. The client prompts for confirmation.
* (Package manager) If a package wants to increase its permissions, then that is suspicious. The client prompts for confirmation.

So trying to apply to the `rc` example - any downstrams newly adopting `rc`@1.2.9 would be liable to blind-acceptance. But all existing downstream upgrading from `rc`@1.2.8=>1.2.9 would get an unusual prompt.

Perhaps your question ultimately turns on "*how frequently do SSH hosts change their keys*" or "*how frequently do packages try to increase their permissions*"?

ivarprudnikov:
For those who mentioned 2FA, there is still an issue with the automated publishing process that would usually use an access token. A token can be leaked, thus 2FA does not solve everything.

mirek:
Another alternative could be to specify on `npm i` minimum duration packages must have been publicly available. Dependency tree will work fine as any dependency must be available at least as long as dependent root. User facing api is minimal, ie. install packages that have been live for at least 3 days only.

Okeanos:
> Perhaps a flag instead to refuse to install packages that have been published in a recent time period or at least flag them interactively? That way each individual user or organisation could configure their preference on how long packages have to have been in the wild before they even get downloaded onto their system.

I am not sure I understand how this would solve this issue? Wouldn't that mean that in case of (critical) security patches everybody using this flag would've to disable it or else run insecure software? Another question I am asking myself here is: if this flag is opt-in instead of opt-out … the attack surface of install scripts isn't really reduced because a lot of people wouldn't know about the flag or use it. Additionally, if everyone waited for say, a week, before installing it would technically increase the window of time in which malicious scripts etc. can be found. But by whom? Who'd do the auditing of this? I currently imagine this would only postpone the inevitable and recreate a status quo "7 days later" (using the example wait window).

I am a casual user of npm/yarn and wouldn't like this burden shifted to me if a better alternative (e.g. like SSH TOFU as previously mentioned) appears to be feasible. 2FA publishing also sounds like a good idea to me, however, as an additional improvement not as an alternative to requiring explicit consent for install scripts by default.

ngudbhav:
What if npm maintains an anti-malware database, just like a regular antivirus? Any install can then be directly halted if the package is marked as unsafe.
To make it a non breaking change, anti-malware can be used as a separate package. npm can issue a warning if installation is triggered without this anti-malware package.
Regular definition updates headache can be removed if the database is maintained online.

A single source of truth for all the vulnerabilities. This can help notify people via tweet/mail about recent alerts.
Organisations can ping the database regularly to find if the package is vulnerable or not.

To make it more streamlined, a simple set of rules like 2FA publish, author's information, last publish source can be kept.

WebReflection:
@tolmasky 

> your goal of showing me disrespect has succeeded

quite the opposite, I wanted to clarify eventually one point after the other, no disrespect meant, not sure where you got that

> by declaring that a huge post that tried to in good faith address your comments isn't even worth reading

I've never said that, I am in another country on data roaming due my father in law (the only one I have) at the hospital due lung cancer complications ... I was trying to read through, but had no time yesterday to come after the entirety of your post, so I've tried to break points down, and promised myself to come back after taking the time, in a place with internet, to read and reply.

Your answer now made it pointless for real though, so I hope this conversation keeps going without shaming other developers intentions out of the blue, thank you.

naugtur:
What's not been discussed here yet is the relationship between what we could do and the actual reason we're talking about it. 

What we need is not banning ALL postinstall scripts. We need to ban UNEXPECTED postinstall scripts.

I tried to remedy this with my recent work - check here
https://dev.to/naugtur/get-safe-and-remain-productive-with-can-i-ignore-scripts-2ddc

If we want to address this on the NPM level, the biggest result we can get without breaking much is preventing packages from introducing postinstall scripts. 

So if you npm install something with a script and there's no package lock, you get asked if you trust it. When lock exists, it should denote if a package has a script to run and if it says not and the latest version has a postinstall script, raise the alarm.
Effectively make a built-in allow-scripts that propagates automatically.

There's very little popular packages with existing postinstall (I know because I have spent a few nights scanning npm for those) and taking over one of them instead of one of countless others is a totally different order of magnitude in probability. 

None of the large impact attacks we saw would be possible with this in place.

TL;DR: First eliminate unexpected scripts showing up all of a sudden instead of all.

tolmasky:
> I was trying to read through, but had no time yesterday to come after the entirety of your post, so I've tried to break points down, and promised myself to come back after taking the time, in a place with internet, to read and reply.

I do apologize if there was a misunderstanding, but in one of your (many) replies to me on Twitter you pointed me to your reply by saying ["I disagreed already with the first line you wrote, I'll read the rest once that's clarified"](https://twitter.com/webreflection/status/1456703065499779081?s=21), which at least to me seems to pretty clearly imply that you expected me to address your concern with the *first sentence* of my post as a precondition to reading any more of it, so I don't think my interpretation was really that off base.

WebReflection:
That tweet thanked you, and it was rushed too. I don’t have much more to add.

tolmasky:
A couple quick updates from compiling questions and comments both here and HN. I'll try to incorporate these into the actual RFC but want to get some notes down here sooner rather than later:

1. I'm trying to compile a list of the various scenarios in which install scripts can be exploited and add those all to the "motivation" section.

2. I think there may be a good argument for adding an entry in the `package.json` file when you use `--allow-scripts`, as someone could otherwise just manually add an entry in the `package-lock.json`, which often goes undetected in PRs since GitHub will hide the diff of large `package-lock.json` changes, and few people actually read those multi-hundred line changes carefully.

mikeal:
> This is nothing new in the software world, to be honest, when you install pretty much any software in Windows […]

It’s worth acknowledging that npm experiences old problems in new ways due to the overwhelming size of the ecosystem. This is the largest package ecosystem in the world, and also the most interdependent, so doing a single npm install will install hundreds or even thousands of packages all written and maintained by different actors.

This is an amazing achievement, we routinely get to harness and combine the effort of thousands of developers with a few keystrokes, but at the scale of this ecosystem what were once marginal risks become all but certain.

That said, this ecosystem also presents a tremendous challenge for a breaking change like this, because even a marginal breaking change is all but certain to enter your terminal. If this isn’t handled with care the ecosystem simply won’t upgrade and this change won’t have the desired effect.

It’s great to see people taking the size of the problem, and the solution, seriously and putting in so much effort. Makes me happy to be in Open Source 😁

avindra:
✅ Support. Scripts should be opt-in. As a developer I am interested first in importing source code. Shenanigans shouldn't be the default case.

bmeck:
I do think erroring out on scripts is potentially fine, but a variety of tasks mostly around compilation are likely to be hard to deal with:

<details>
<summary>
looking at the sheer support matrix of @datadog/native-metrics for example is 31 precompiled C++ modules:
</summary>

```
package/prebuilds/darwin-ia32/node-102.node
package/prebuilds/darwin-x64/node-102.node
package/prebuilds/linux-x64/node-102.node
package/prebuilds/win32-ia32/node-102.node
package/prebuilds/win32-x64/node-102.node
package/prebuilds/darwin-ia32/node-72.node
package/prebuilds/darwin-x64/node-72.node
package/prebuilds/linux-ia32/node-72.node
package/prebuilds/linux-x64/node-72.node
package/prebuilds/win32-ia32/node-72.node
package/prebuilds/win32-x64/node-72.node
package/prebuilds/darwin-ia32/node-79.node
package/prebuilds/darwin-x64/node-79.node
package/prebuilds/linux-ia32/node-79.node
package/prebuilds/linux-x64/node-79.node
package/prebuilds/win32-ia32/node-79.node
package/prebuilds/win32-x64/node-79.node
package/prebuilds/darwin-ia32/node-83.node
package/prebuilds/darwin-x64/node-83.node
package/prebuilds/linux-x64/node-83.node
package/prebuilds/win32-ia32/node-83.node
package/prebuilds/win32-x64/node-83.node
package/prebuilds/darwin-ia32/node-88.node
package/prebuilds/darwin-x64/node-88.node
package/prebuilds/linux-x64/node-88.node
package/prebuilds/win32-ia32/node-88.node
package/prebuilds/win32-x64/node-88.node
package/prebuilds/darwin-ia32/node-93.node
package/prebuilds/darwin-x64/node-93.node
package/prebuilds/linux-x64/node-93.node
package/prebuilds/win32-ia32/node-93.node
package/prebuilds/win32-x64/node-93.node
```

</details>

These cannot be recreated with N-API, WASM, or JS and C++ bindings are needed. Requiring projects to default to pre-build for a support matrix like this seems prohibitive. I think most compilation chains are running arbitrary scripts effectively so they do fall in the bucket of install scripts and this needs to be considered. In particular, if a project wants to run compilation on the installer's machine it needs to actually run those commands. That being said, it is fairly rare to actually have a compiled module in your module graph. Most application installations I have worked with do have a compiled module but the percentage of modules that are compiled is quite small usually.

In general, permitting every single module to run arbitrary scripts is undesirable from this issue thread. Requiring manual vigilance when running the CLI is exactly what we wish to avoid, so we definitely need to not have 2 modes of operation.

1. refuse to run any scripts
2. permit running any scripts

Instead individual modules may be expected to run with dangerous privileges for means like C++ compilation. But this can be confusing. If you install `dd-trace` for example like most applications should, how would a user respond to messages around `@datadog/native-metrics` needing permissions? Additionally, why would a user grant arbitrary script execution rather than the ones `dd-trace` is requesting to have permissions to run scripts. It isn't the module that you attempted to install but `@datadog/native-metrics` is actively being installed on behalf of `dd-trace`. I can imagine multiple UX around this, but in general I don't think a global boolean flag that is needed for every CLI invocation is ideal, and I don't think only having configuration on the package with the scripts is good either.

tolmasky:
> I would recommend letting folks comment on the RFC, and then attending the next RFC call. The folks who regularly attend are great, and really open to discussion.

Is there a schedule for RFC calls? My apologies, couldn't find anything in the main repo other than the meeting notes for the last call.

WebReflection:
@tolmasky

> a very large portion of package installs are done not by humans, but in an automated fashion

Automations are configured by (hopefully expert) humans though, hence idelly configured by people that know what they are doing.

`npm i` on the other hand, is perfomed by pretty much everyone, which is the easy entry-level welcome we used to love about JS so far ... right?


> A package going from not running an install script to running an install script between CI runs should IMO absolutely fail the test ... "wait a minute, none of these should be installing", and avoid having a bunch of junk infect your CI machines.

Like I've said, we're confining *npm* to CI only ... while I care about every otehr use case that's not CI, as *npm* has no boundaries in usage, by definition, isnt't it?


> Another important piece of context is that vast majority of people will actually see no change, as a tremendously small amount of packages on npm today use this feature ...

I can erspectfully yawn around this point ... define people, are those developers, are those the people from SO with copy/paste habit? what kind of people and what data do you have around those peoiple?


> Additionally, many of the packages that do use this legitimately now have better alternatives.

No. `postinstall` is about doing something after, meaning:

 * the files are there
 * the scripts are there
 * everything will work as expected

That's *postinstall* in a nutshell.


> allowing you to globally allow "safely" a specific version of a package

why would I care about about any global "*safe*" thing, which is to my opinion rather yet another seurity attack vector instead?


> Thinking about the distinction between "automated" installs vs. "user drive installs"

There's no thinking there to me: just differnt use cases. Think about only one of thoes, and screw every other.


> you can even imagine a model where when running `npm install` interactively

Your whole complain was about not considering CI, and you your last resort is about running `npm i` interactively?



> The proposal is to make unexpected behavior default-off.

Once again, I've never stated my position was meant to block this proposal, it was all about thinking further all options to make the *npm* ecosystem better.

wesleytodd:
> Is there a schedule for RFC calls? My apologies, couldn't find anything in the main repo other than the meeting notes for the last call.

@tolmasky There are issues created in this repo: https://github.com/npm/rfcs/issues?q=label%3AMeeting+is%3Aclosed

Also I think there is a google calendar link somewhere, maybe @darcyclarke has it? Might also be good thing to add to the readme.

jwalton:
> Since this presupposes that the package you're installing might have malicious code, how do you propose to handle the threat posed by running your own trusted code which imports from these untrusted libraries?

To give you a real world example - I am a very fast typist, but not the most accurate.  Occasionally I mistype something.  On a handful of occasions I've gone to install, say, react, and I've typed:

```sh
$ npm install r
```

and then nudged the enter key.  So now I'm installing (and running scripts from) a package I never even wanted to install in the first place.  Let's go have a look at this package and figure out what I just accidentally installed (and possibly ran scripts from):

https://www.npmjs.com/package/r

Uhh...  That's not very informative.  So now I need to `npm info r` so I can download the tarball and make sure it didn't install a bitcoin miner on my machine.  Forunately, in this particular case I came out OK, but it would be much nicer if `npm install` wasn't a loaded gun with potentially irrevocable harms involved.  If you could `npm install`, and then poke around in a package to see what the source code does, you could theoretically avoid malicious code by being vigilant, and then `npm uninstall` it before it has a chance to do something bad.  (`npx` is another "loaded gun" here, and something you should never use from the command line - use [`npm-run`](https://www.npmjs.com/package/npm-run) instead.)

tolmasky:
> To give you a real world example - I am a very fast typist, but not the most accurate. Occasionally I mistype something. On a handful of occasions I've gone to install, say, react, and I've typed:

Yes, I've had similar problems before, but with typos (like "lodhas" instead of "lodash", etc.). I think install scripts create the opportunity of an entire class of attacks where you can just squat a package out on common typo-ed versions of popular packages hoping for people to make these mistakes.

tolmasky:
> imagine we flip the opt-in so that `--ignore-scripts` or `--no-scripts` or however the flag is called now becomes the default, do you have any idea how to enable just one dependency down the pipe of all others that **needs** a `postinstall` to work?

I think I might be misunderstanding the question because I think that is precisely what the [--allow-scripts]((https://github.com/npm/rfcs/blob/15ec3f83cc19e5425302ad3900d4ccfa2b29bab5/accepted/0000-make-scripts-install-opt-in.md#npm-install) flag in the RFC is for. If you are installing "some-package@latest", which itself has a dependency on "built-package@1.0.3", then you would just do:

```bash
npm install some-package --allow-scripts built-package@1.0.3
```

This will now recursively install some-package and all its dependencies, and only allow built-package@1.0.3 to run its install scripts if it has any (and issue a non-failing warning if it built-package@1.0.3 did not have any scripts, to let you know that something was off about your expectations). This flag can be repeated so you can turn on scripts for as many packages as you want.

WebReflection:
> I think that is precisely what the --allow-scripts flag in the RFC is for

sort of ... I have mentioned *down the pipe* not by accident, as this is overly confusing to me (beside the typo, missing *s* in the second `--allow-script`):

```
npm install canvas@1.0.0 --allow-scripts canvas@1.0.0 --allow-script ramda@2.0.0
```

If I allow a specific version of `canvas`, and canvas needs some dependency to run scripts too, I don't understand why it becomes responsibility of the user to explicitly know upfront, and allow, canvas dependencies too ... I mean, if canvas changes a dependency in a patch (totally fine/allowed if it doesn't technically break semver) should every installer out there also drop `ramda` and put `delta` with a version in there?

To simplify the question: would `npm i thing --allow-scripts-unsafe thing` work for both *thing* and every nested sub/dependency *thing* might have and need to work, in terms of scripting?

If that's the case (I hope so, or it'll be a mess to maintain when you trust, own, know a module) could we have a shortcut for that flag? Thank you!

bmeck:
In general I think the problem with nested dependencies is exactly because there seems to be too much focus on a CLI flag as the solution. Other ways to enable things exist, for example to just show a simple alternative:

1. `npm install --allow-scripts canvas` could only allow the scripts directly in `node_modules/canvas/package.json`
2. `node_modules/canvas/package.json` could define that it expects dependencies to execute scripts:

```
{
  "dependencies": {"nested-dep":"*"},
  "allowDependencyScripts": {
    "nested-dep": ["pre-install"]
  }
}
```

This would keep the trust explicit and well formed with direct paths between all paths causing script execution.

I don't think the example above is a good UX, but it is just trying to show that there are other solutions and perhaps we should focus on the problems we discover with workflows rather than overly focusing on how to argue that a certain workflow is the solution.

tolmasky:
> If I allow a specific version of `canvas`, and canvas needs some dependency to run scripts too, I don't understand why it becomes responsibility of the user to explicitly know upfront, and allow, canvas dependencies too ... I mean, if canvas changes a dependency in a patch (totally fine/allowed if it doesn't technically break semver) should every installer out there also drop `ramda` and put `delta` with a version in there?

This is by design, and gets at the heart of the matter of why all this stuff is so hard to track. The reality is that most of the time a package's dependencies don't change because they themselves make an explicit change, but because these subdependencies (or their subdependencies...) themselves change but still abide by the semver range listed in the parent package's package.json.

For example, if the developer of canvas originally installed `some-dependency`, and it happened to version 1.0.0 and happened to have no install scripts, it still got recorded into the package.json as "1.x.x" by default, and not "1.0.0". Now, when you go to install canvas next week, `some-dependency` may have since been updated to `some-dependency@1.0.1`, which now *does* include an install script, and there's really no way for that "past" canvas to know anything about that, and more importantly the canvas version doesn't reflect that. Because a package version is (unfortunately) not a static artifact, but rather a fluctuating set of possibilities that changes minute by minute defined by changes to its dependencies that are out of its control.

So, we have a few elements working together here:

1. The fact that semver ranges are allowed at all in package.json files at all means that there usually exists no "static" or "canonical" installation of any given package with dependencies, as any sub-sub-sub-dependency could be different at any given install.
2. The fact that npm chooses by default to always record dependencies as ranges and not specific versions means that the vast majority of dependencies fall into this category, and it's not just some edge case.  The clock starts ticking the second you hit publish on your package and the possible "semver valid" installs of any given dependency tree explodes exponentially very quickly.
3. `package-lock.json` could solve this problem, but they are explicitly recommended against at the package level,  which means that everyone that types "npm install canvas" still gets a different version than each other. 
4. semver versions are just human decisions. Anyone can just choose (either willfully or accidentally) to make any change just be a "patch bump", and thus likely to pass the semver range check on every dependent's package.json. There's plenty that could be done here (for example I wish the npm server enforced major version changes if it detected a different license file), but ultimately you can't actually reasonably catch "invalid patch changes", especially malicious ones that deliberately try to hide it.

The end result is that people install "package permutations" that are completely different the one on the package authors' machine, and from each other, due to the "semver range"-first nature of npm and package.json files. *If* we instead had a "semver version" based system (which would be equivalent to everything having package-lock.json files), where your internals can't change without you yourself updating your own version accordingly, then I'd actually be right there with you. You'd actually have a deterministic chain of dependencies, and all of a sudden you can reason about a package version. This is incidentally how NixOS dependencies work. If you a sub-dependency changes, it sparks a rebuild up the dependent-chain, and that "deep update" is represented with a "top level" version change. In other words, "canvas@1.0.0" has only one meaning, not a range of possible meanings that is time dependent.

tolmasky:
> In general I think the problem with nested dependencies is exactly because there seems to be too much focus on a CLI flag as the solution. 

The focus on the CLI flag is ultimately because all this has to start with someone installing the package to start off with. Most people don't hand-write their package.json files, then type "npm install" afterward. If the choice to run scripts is then recorded in the package.json, then things are fine afterward. This is also the reason I mentioned having an interactive mode too -- so you just type "npm install top-level", and as it encounters install scripts it asks you about them, and then records all that in the final package.json. I do however strongly believe that this should be clamped to specific versions...

> I don't think the example above is a good UX, but it is just trying to show that there are other solutions and perhaps we should focus on the problems we discover with workflows rather than overly focusing on how to argue that a certain workflow is the solution.

I agree there should be a package.json component (and I am working on adding this to the RFC, namely auto-generated from the initial install), but I think it has to ultimately be version-specific, or it will suffer the same issue as we currently have: an out-of-band malicious patch update immediately is immediately allowed to run in a dependent's installation as long as a past version was trusted. The live examples we've seen of this exploit specifically target previously installed packages and just ship patch updates. See my above comment on semver ranges for more thoughts on this.

We are currently running our script re-analyzing every package on npm right now, so I hope to very soon provide some meaningful data around all this. I think a lot of this comes from the fear that people are going to be having to manage lists of hundreds of install-script permissions or something, which is what leads to wanting to defer these security decisions "down the line", where as long as A says B can install, and B says C can install, and C says D can install, then D can install even if its a brand new version that just got published 5 minutes ago and no one in that entire dependency chain has actually even had a chance to look at that change and whether they're still comfortable with it.

tolmasky:
> I think an important thing to note here is that multiple versions could be trusted and this does not need to be version specific, it just cannot be a range past what it is installed. E.G. if we trust `foo@1.0.1` and we update to `foo@1.0.2` I don't think trusting `foo@1.0.2` implies to not trust `foo@1.0.1`. Ranges likse `"foo": "1.0.1 || 1.0.2"` do cover both versions that were trusted. Care does need to be taken though not to imply trust on the ranges in between (like pre-releases).

So the current system in the RFC does allow for listing multiple versions of the same package, as name/version pairs are just considered their own specifiers (so, you can do "canvas@1.0.1" and "canvas@1.0.2"). There are two important differences with what you are suggesting from what is currently described though I think:

1. Ranges provide the possibility of "infinite allowed versions", which you touched on. Something like "1.x.x" allows for any future version. But perhaps you meant specifically that only "or" semver ranges are allowed. If the latter is the case, then at least in this aspect they are equivalent and I'd have no problem with that shorthand.
2. The more important difference seems to be that the purpose is to allow deferring these decisions to dependencies, at which point it becomes necessary for the registry to be involved in enforcing the "only versions that were around at time of installation rule", because this isn't something the client can enforce, only the server. So if you list 1.0.1, 2.0.0, 2.0.1, but only 1.0.0 and 2.0.0 exist in the registry, then it should fail since you're "pre-allowing" versions that don't exist. Clearly if more arbitrary semver ranges are allowed this becomes even more complex (in theory somehting like 1.x.x would have to be "flattened" to the versions of 1.x.x that are currently present in the registry, so that future 1.x.x versions aren't allowed).

WebReflection:
> it will suffer the same issue as we currently have: an out-of-band malicious patch update is immediately allowed to run in a dependent's installation as long as a past version was trusted.

I agree with what @bmeck said there: if a module was trusted already, it doesn't mean that its next version should be trusted too (unless that module is behind 2FA, which I keep believing should be part of this *trusting* equation/discussion).

On the other hand, if a module had **no scripts** and suddenly a patch update (or any other version) brings in scripts, then the installation should warn, or fail, because that module introduced a **breaking change** in terms of installation procedure, so that the user, or the CI, should explicitly accept/allow that change and trust it's going to be OK from that time on.

This makes more sense than asking everyone to maintain a long list of sub-dependencies modules, imho, but I also like the package.json approach, so that module authors can explicitly declare themselves what they trust.

Add 2FA details in the process, and see that it could really work well as measure to avoid all the same issues we had to date.

tolmasky:
> On the other hand, if a module had **no scripts** and suddenly a patch update (or any other version) brings in scripts, then the installation should warn, or fail, because that module introduced a **breaking change** in terms of installation procedure, so that the user, or the CI, should explicitly accept/allow that change and trust it's going to be OK from that time on.

For the record, I think this sort of stuff would be very cool, however, it ultimately requires registry changes and backend support, which we have no direct access to (as opposed to the CLI), and that are subject to schedules that are out of our control.

> This makes more sense than asking everyone to maintain a long list of sub-dependencies modules, imho, but I also like the package.json approach, so that module authors can explicitly declare themselves what they trust.

I think this relies on the assumption that this is a highly used feature and that we can't come up with alternatives for the most common use cases. I think both aren't true, but we should hopefully have a more concrete understanding of this beyond our own anecdotal experiences soon enough.

I also think that the "provide good alternatives" is a critical part of this story with its own benefits. I think moving to a world where common things are both easier, not requiring a roll-your-own script, and also more secure (completely in the control of npm), would be a big win in its own right. The goal is not to remove this feature, it is to limit it to legitimate use cases that really would have no other resource, so that they could be properly considered by the end user without being overwhelmed with random little dependencies failing to install simply because they like to present ads in the terminal from their install scripts.

tolmasky:
> I was rather thinking about a `package-lock.json` based approach, where a field in that schema could tell if it was trusted before, or surely if there were "dangerous scripts" to consider, or previously enabled by the current env, if that makes sense.

I *believe* this is handled by the [`npm config`](https://github.com/npm/rfcs/blob/15ec3f83cc19e5425302ad3900d4ccfa2b29bab5/accepted/0000-make-scripts-install-opt-in.md#npm-config) part of the RFC already, where you can set a package/version pair as trusted "from now on" and never be bugged about it again. It also allows you to specify this for "every version" of a package, and merely has the word unsafe in the congif option to alert you that this isn't the best idea to do with everything (but for example could be very handy for your own packages). npm config can be used both by users and companies/machines, so if you have an approved "global list" of packages it is trivial to throw that onto your build machines and not have to do individual flags/package.json/etc. for every individual install. Is this sufficient for what you were looking for, or does this not cover the cases you are thinking of?

WebReflection:
yeah ... that solves for "*my machine*", but it's extra stuff to *copy/paste* and write down in READMEs, when needed.

there are still [new comers](https://github.com/WebReflection/discussions/discussions/64) that have hard time starting with most basic topics, but there are also projects such as [opencollective](https://github.com/opencollective/opencollective-postinstall#readme) that are 100% based on `postinstall` behavior.

I've created long time ago a [lighter alternative too](https://github.com/WebReflection/lightercollective#readme) and I guess time to exploit `postinstall` to inform people there are developers behind what they install is gone.

Anyway, this exchange answered all my questions, thanks and feel free to ping me if needed 👋

tolmasky:
> there are still [new comers](https://github.com/WebReflection/discussions/discussions/64) that have hard time starting with most basic topics, but there are also projects such as [opencollective](https://github.com/opencollective/opencollective-postinstall#readme) that are 100% based on `postinstall` behavior.

So, this I think should be the poster child of what not to use an install script for, especially under the context of "we should be coming up with good alternatives", because we already have! npm has *already* built an official system for requesting support with the [`funding` package.json field](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#funding). Newcomers should be learning the standard, built-in system that works with commands like [`npm fund`](https://docs.npmjs.com/cli/v7/commands/npm-fund), not installing additional packages in their dependency chain that depend on the most general purpose tool just to show to log a message, each implementation of which defines its own custom environment variables to disable. We can't be leaving arbitrary script access in just to support these behaviors that have trivially-implementable standard solutions. And again, give everyone a year to update! Email everyone that has opencollective in their dependency chain and warn them that in *a year* if they don't switch to the 2 line package.json method then people will have slightly more friction installing their package. We have the data to hand-hold everyone through this process.

WebReflection:
> time to **exploit** `postinstall`

by no mean I like that approach and I can't care less if these old ways to show stuff while installing are gone ... it's just yet another example of breaking change that *might* e considered **or**, it'd be great if all modules based on such behavior would be automatically deprecated, including mine (don't care, nobody read it anyway)

danrevah:
That’s a really good step forward. 

My only concern is that we eventually don’t really know what’s running when we’re installing an NPM package, and it would always be the case while having the benefit of using a third party library.

**In an ideal world** I would love to see this proposal taken even step further. In addition of plainly having the developer decide either if to run the post/pre script of a specific library, each library would also have a permissions config, this could even be part of the ‘package.json’ file.

For example:
‘’’
{
  “permissions”: { “file_system”: [“read_files”] }
}
‘’’

That it would require explicit permissions from the user to do specific actions, and would also alert the user when the same package is asking for more permissions during ‘npm install’.

Yes, it’s a big change. IMHO that’s what tackling great challenges requires.

mikl:
Does preventing these scripts really make that much of a difference?

The thing people usually do after installing an npm package is running said package in a Node.js environment, where a malicious script would have more or less the same access to do Bad Things™ as they would have had with an install script.

Unless I’m missing something, only in cases where the npm package is not run in Node.js (ie. dependency management for browser code or npm packages containing fonts or other non-code assets) would the elimination of install-scripts effectively prevent malicious code from running.

Unless we completely rethink Node.js and start limiting what code loaded into the runtime can do (a bit like Deno does), I don’t see how limiting install scripts would make a serious security improvement. It would just be a tiny speed-bump for bad actors.

tolmasky:
> Does preventing these scripts really make that much of a difference?

Hi Mikkel, yes there are a great number of circumstances where this makes a big difference, that have been outlined in various comments here, but I'll quickly resummarize some of them here:

1. Per your own post, if you are using npm for purely frontend work, this opens a completely unnecessary attack surface. It might be easy to forget here in node land, but many teams that use completely different language for there backend work only fall precisely in this category. It doesn't make sense to be exposing Ruby or Python devs to install scripts, when otherwise the code they'd install would never run on their own systems and be confined completely to the browser sandbox.
2. Install scripts may run under different privileges than the node apps run in. Do a search for "--unsafe-perm" and you'll see that for better or worse this is a suggested "fix all" for various npm install woes, which thus exacerbates this problem.
3. Install scripts run in different environments than node apps. As mentioned above, install scripts may be uniquely positioned to be the only attack vector on build machines, that otherwise wouldn't run code at all (some companies for example separate build machines and CI machines, etc.) This is made worse by the fact that it is not uncommon to have builds and CI take place automatically in response to a pull request, and thus all it takes to infect the machine is issuing a pull request with the package.
4. Install scripts allow for "typo squatting" where you can register names close to popular packages such that if you accidentally type "npm install canvsa" the install script immediately kicks off before you can even realize and hit ctrl-c and its already spawned a process to start doing something malicious.
5. There is also the fact that install scripts can be *easier* to introduce precisely because its not code that needs to be imported and run to attack. As such, it is possible to sneak a package into the package-lock.json as part of a larger change, since github will often automatically hide package-lock diffs, and no one really takes the time to read them since they're not really meant to be human-readable. Compare this with was malicious code that needs to be run, where you have a higher chance of noticing it in a diff since the actual code is where people usually pay attention when reviewing code and they'd ask "why are we importing this library we didn't need before?". Again, it doesn't make it impossible, it's just much easier with scripts.

To be clear, there are many more security vulnerabilities outside of this. This has been repeatedly actively exploited though, and the bar is super low (people now just straight up copy old attacks and insert them in new places). When you couple this with the fact that it is infrequently used and many of the use cases could be standardized, it seems like a good direction to merely require these (hopefully increasingly) rare uses to be approved by users. This comment I made is also helpful with respect to this question: https://github.com/npm/rfcs/pull/488#issuecomment-962195445

matthiasg:
> around modules behind 2FA, or provide at least _that_ detail when _info_ is used, or a module surfed through the site.

How would 2FA offer any more trust when the second factor is not at least linked to a telephone number ?

Currently npm asks for an authenticator app code based on [rfc6238](https://datatracker.ietf.org/doc/html/rfc6238) not even a physical key. It can be used on a number of 'anonymous' devices. So it protects a tiny bit more and can mitigate effects from 'hacked' passwords, but the typical attacks are too different. Too readily accepted pull requests, maintainer change, misspellings of package names etc. 

I would hesitate to put too much emphasis on a verification based on 2FA, though I would agree that it might be a base layer.

In general security of the executing code would need to be improved. It should for example be limited to their own package source and each package should also by default not get access to communication features unless specifically allowed by the main author.

I would follow @tolmasky 's arguments and default to use an opt-in to limit the attacks in general and put pressure on the package owners to remove any install time scripts (I haven't used any in years).

ljharb:
Isn't linking 2FA to a phone number pretty insecure, given SIM swaps and whatnot? I'm pretty sure an authenticator app is much more secure than a phone number.

WebReflection:
> I would follow @tolmasky 's arguments and default to use an opt-in to limit the attacks in general 

in multiple times I've stated these two are **not** mutually exclusive, but 2FA is either useful or it's not, and we can just remove it as it provides zero extra guards, as you state, right?

Well no. Password leaks are very common still in 2021, and without 2FA a lot of modules, including mine, in the past, could've been hacked with ease. I added 2FA after last global leak from known site and never looked back. I suggest you, and everyone else, to do the same.

matthiasg:
@WebReflection I am not arguing against 2FA :) It does offer a little extra protection, but the emphasis should be on not running package scripts, not matter whether they use 2FA or something else.

Requiring 2FA or higher validation levels for package dependencies emitted as warnings or errors in general would make sense as an independent setting of course and could put some positive pressure on the ecosystem.

WebReflection:
@matthiasg so we kinda agree then ... me pushing for *any* way to empathize 2FA is being used is because *now* there's literally **zero** knowledge whatsoever about 2FA behind any module, even those with million downloads per week.

I am looking forward to see emphasis around 2FA instead, for that positive pressure on the ecosystem you mentioned.

thescientist13:
@WebReflection 
Not sure if it exists, but would it make sense to spin up / contribute to a new RFC about 2FA?  Could then get into the queue for the weekly NPM calls and continue the thread in parallel to this one.

Maybe it could even be metadata in _package.json_ that could display an icon on npmjs.org like  it does for TypeScript?
Screen Shot 2021-11-10 at 10 04 52 AM

WebReflection:
@thescientist13 I am not familiar with the RFC process in here, but yeah, I guess that makes sense, although self proclaiming a package 2FA enabled with `package.json` is nor really what I am after there ... the TS case is just "*point at some index.d.ts and icon happens*", which is equivalent to zero extra trust to me. I will stop mentioning 2FA though, it's just that this whole topic to me should result into a `--trust` flag, as opposite of `--allow-scripts-unsafe-modulename` specially because `@group/name` are also module names, and a config flag such as `--allow-scripts-unsafe-@group/name` looks super awkward, while an explicit *trust* flag would be clearer, as *trust* is the kind of fix we're proposing here (or better, this is about *not strusting*, but *trust* is the keyword that matters to me).

Will try to find time to check how to contribute with a new RFC around this topic.

thescientist13:
No problem, just also felt it might a good thread worth tugging at in any of its forms.  The README should hopefully provide a comprehensive overview of the process, and then meetings happen weekly and are posted as issues in this repo.
https://github.com/npm/rfcs/issues?q=is%3Aissue+is%3Aclosed

everett1992:
@WebReflection I think you're over indexing on 2FA to I'm in favor of publishing package provenience data, including ownership and access and credential types.

---

> Password leaks are very common still in 2021

This qr code decodes to `otpauth://totp/test-account?secret=4OWWKH4IFOARZABD7K6RQ3JPFBFLHC5I`.

oathtest-qr

If this were a real account npm would store  or be able to generate `4OWWKH4IFOARZABD7K6RQ3JPFBFLHC5I` to validate a one time pass. If the secret is leaked by npm or the user an attacker could generate valid one time passes. npm's 2FA is [rfc6238] a time based extension of [rfc4226] which is a shared secret algorithm. [Section 7.5] covers management of shared secrets. If npm (or any other rfc6238 provider) implements *Random Generation* they have a table of all users 2fa secrets which can be used to generate one time passwords. Unlike passwords these cannot be hashed or slated. 

[rfc6238]: https://datatracker.ietf.org/doc/html/rfc6238
[rfc4226]: https://datatracker.ietf.org/doc/html/rfc4226
[Section 7.5]: https://datatracker.ietf.org/doc/html/rfc4226#section-7.5

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,49 @@
...
+    - `npm install canvas --allow-scripts-unsafe canvas`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts.
+    - `npm install canvas --allow-scripts-unsafe canvas --allow-unsafe-scripts ramda`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts for canvas and any version of ramda it encounters.
+
+### npm config
+We should add the ability to configure npm globally to allow packages to make use of scripts. This can be useful for packages you install a lot, or for packages you are working on:
+
+1. `npm config set allow-scripts-canvas@1.0.0 true`: This will allow canvas version 1.0.0 to install scripts whenever `npm install` is used.
+2. `npm config set allow-scripts-unsafe-canvas true`: This will allow any version of canvas to install scripts whenever `npm install` is used. It should also probably warn on each case.
+
+## Rationale and Alternatives
```
everett1992, line 36: We should add a scope section, your comment here is a good starting place
https://github.com/npm/rfcs/pull/488#issuecomment-964714944

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Implementation
+
+I imagine this should hopefully not be harder than adding some if statements around the right lifecycle sections of install.
+
+## Unresolved Questions and Bikeshedding
+
+Two open questions for me:
+
+1. Having npm ship with a grandfathered list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags.
```
everett1992, line 48: N.B. grandfather clause has a pretty nasty history, I try to use  'legacy'.
> grandfather clause
> a provision in several southern state constitutions designed to enfranchise poor white people and disenfranchise Black people by waiving high voting requirements for descendants of men voting before 1867

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,49 @@
+# Transition to a model where install scripts must be explicitly allowed during the install process.
+
+## Summary
+
+Instead of by default always running the install scripts (`preinstall`, `install`, `postinstall`, `prepublish`, `preprepare`, `prepare`, `postprepare`) if they are present during the install process, provide flags to require users to explicitly allow them to run, either whoelsale as "one big swtich", or on a package by package (and optionally version by version) basis. Also provide matching `npm config` options to do the same globally and permanently instead of on every install.
```
everett1992, line 5: Please clarify if scripts are *ignored* or *fail* the install if they are not allowed to run.

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,49 @@
...
+4. The `--allow-scripts-unsafe` option is added for more convenient, but more dangerous, handling of install scripts:
+    - `npm install --allow-scripts-unsafe canvas`: Install the contents of `package.json`, allowing any version of `canvas` to run install scripts.
+    - `npm install --allow-scripts-unsafe "canvas@1.0.0-2.0.0"`: Install the contents of `package.json`, allowing any version of `canvas` in the semver range 1.0.0 to 2.0.0 to run install scripts. While strictly safer than the unranged version above, it still runs the risk of new vulnerable packages being snuck in in that range. For example, if an account is compromised, and a patch version is published to every major version that is present. 
+    - `npm install canvas --allow-scripts-unsafe canvas`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts.
+    - `npm install canvas --allow-scripts-unsafe canvas --allow-unsafe-scripts ramda`:  Installs the latest version of canvas to the current `package.json`, and allows it to run install scripts for canvas and any version of ramda it encounters.
+
+### npm config
```
everett1992, line 30: The set of packages that can run scripts should be tracked in source control so developers or CI agents can install the projects dependencies without making decisions about which scripts to run. Choosing scripts to run should be done when adding or updating dependencies, not when installing packages that are already dependencies.

I think package.json is the best location for this option. 
 - package-lock.json is ephemeral and may be deleted to get a fresh install.  
 - `.npmrc` may or may not be tracked in source. [^1]

 
 [^1]: If it is checked into source developers need to avoid checking in personal changes. For example a dev needs a proxy for one project but not another. If they set the config in project npmrc they must not commit that change, if they set the config in their user npmrc they must remember to remove it before working on another project. I should write a RFC for multiple project npmrc's.

WebReflection:
@everett1992 about this

> I think you're over indexing on 2FA tho I'm in favor of your suggestion

apparently it's not just me believing is [a better way forward](https://github.blog/2021-11-15-githubs-commitment-to-npm-ecosystem-security/#continued-investments-in-npm-security) ;-)

Screenshot from 2021-11-17 00-25-08

bmeck:
Could we get a STDEV on the total number of transitive dependent packages on packages with scripts?

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,76 @@
...
+## Implementation
+
+I imagine this should hopefully not be harder than adding some if statements around the right lifecycle sections of install.
+
+## Unresolved Questions and Bikeshedding
+
+Two open questions for me:
+
+1. Having npm ship with a legacy list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags.
+2. I think the automatic node-gyp stuff should be under this umbrella as well. That is to say, node-gyp would only run when gypfiles are detected if the `allow-scripts` flag allows it, since you can also techncially put whatever you want in there I believe, and not doing this would simply mean that any malicious behavior would be transitioned there.
```
naugtur, line 76:
```suggestion
2. I think the automatic node-gyp stuff should be under this umbrella as well. That is to say, node-gyp would only run when gypfiles are detected if the `allow-scripts` flag allows it, since you can also techncially put whatever you want in there I believe, and not doing this would simply mean that any malicious behavior would be transitioned there.

## Prior art
- run scripts from allowlist [allow-scripts](https://www.npmjs.com/package/allow-scripts)
- run scripts from allowlist [allow-scripts from lavamoat](https://www.npmjs.com/package/@lavamoat/allow-scripts)
- tools to prevent/detect scripts running when not wanted [preinstall-always-fail](https://www.npmjs.com/package/@lavamoat/preinstall-always-fail) [pentest-my-ci](https://www.npmjs.com/package/@naugtur/pentest-my-ci)
- reviewing which scripts need to run to help build allowlists [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts)
```

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Implementation
+
+I imagine this should hopefully not be harder than adding some if statements around the right lifecycle sections of install.
+
+## Unresolved Questions and Bikeshedding
+
+Two open questions for me:
+
+1. Having npm ship with a legacy list of "known safe" package versions. For example, every currently shipping version of canvas. That way, for important large packages, this is a "from now on" requirement. Alternatively, there could just be a transition period where npm just warns about script, and then afterward we make it actually refuse to run them without the proper flags.
```
ljharb, line 75: a hardcoded package list is a very bad idea - this incentivizes authors to never publish updates, since existing known-good versions are privileged.

naugtur:
If we disable scripts by default, we need means to run compilation etc. for some packages. npm rebuild could do that. 
I have a poc of running arbitrary code if you run npm rebuild on something trustworthy. 
Who should I talk to?

Ilmarinen100:
Since this wasn't well received in the meeting in January and we had at least one more high profile hack through this path:
What if we make the install scripts opt out in a usable way?
Right now `ignore-scripts=false` in `.npmrc` is not practical as it breaks project-local scripts, too. Potentially there could be other solutions like a more fine-grained config key (or value for ignore-scripts), that specifically disables scripts for package installation.
I'm assuming here there is a large group of devs like me that usually do not want to run any install scripts (and that are sufficiently worried about security to turn on such a setting globally or per project).

fengmk2:
https://github.com/styled-components/styled-components/commit/ba9d732ca7da53f2a095e35450ecffd592c6f5ba

mbalabash:
> Install scripts that can run just about anything by default pose some pretty serious security considerations, and these are inreasingly moving out of the theoretical realm and becoming actively exploited. See for example here: https://therecord.media/malware-found-in-coa-and-rc-two-npm-packages-with-23m-weekly-downloads/.

Just wanted to highlight that install scripts are not the only threat and there are cool tools that may help to mitigate those risks. I would suggest seeing [sdc-check](https://github.com/mbalabash/sdc-check) and [scanner](https://github.com/NodeSecure/scanner).

naugtur:
LavaMoat

silverwind:
FYI, pnpm has implemented this in v10: https://socket.dev/blog/pnpm-10-0-0-blocks-lifecycle-scripts-by-default

everett1992:
We will likely fork or patch this feature if it's not handled by npm. Or migrate to pnpm (which would be a huge lift).

I'm personally okay to compromise on scripts being opt-in by default as long as packages can opt-in to an allowlist mode. That makes it a non breaking change for npm. Then informed and careful users can limit their exposure to what seems to be the most common vector for malware in npm. Changing the default can be deferred. 

We've also come a long way since @MylesBorins [comments in 2021](https://github.com/npm/rfcs/pull/488#issuecomment-962247162) and npm has better support for host specific distributions, so there are fewer legitimate cases for postinstall scripts, so maybe changing the default has more appetite?

Another requirement for us will be the ability to share allowlist configs. I like pnpm's [config dependencies](https://pnpm.io/config-dependencies) and ability to configure via [.pnpmfile.cjs](https://pnpm.io/pnpmfile) but equivalents would be out of scope for this RFC. We could manage with a global list that's merged with a per-package list.

naugtur:
No need to fork anything. Use https://www.npmjs.com/package/@lavamoat/allow-scripts

Disclosure: I'm the maintainer.


That being said, if there's a move to change the default in npm I'm happy to contribute with some experience from almost 5 years of running with scripts turned off. And pieces of the implementation too.

BTW I'm working on pinning versions in the allowlist for lifecycle scripts now. With recent events, the malware arriving in a script that's already been approved is becoming more likely than we'd want.

everett1992:
@lavamoat looks like a better option than forking. I will need a way to share allowlists across packages, have you thought about how shared configs would work?

naugtur:
Please make this an issue in our repo and I'm happy to discuss.
Let's not detour this thread too much.

Getting npm to default to no scripts would be great, but not easy.

JamieMagee:
A new revival is now open at #868. Same default-deny direction as this RFC, with the phased migration plan that wasn't in the 2021 version. New surface: an `allowScripts` map in `package.json`, plus `npm approve-scripts` / `npm deny-scripts` commands. Worth following if you 👍'd this one.

owlstronaut:
Closing in favor of #868 . Thank you for all the thought and discussion on this issue.

Issue #463, opened 2021-09-24
[RRFC] Multi-app Monorepo Support


mysterycommand:
# Before Opening Please...

- [x] Search for an existing/duplicate RRFC which might be relevant to your RRFC

## Motivation ("The Why")

Please see detailed description[here](https://github.com/npm/rfcs/discussions/462), but tl;dr I'd like npm to "natively" support installing a package workspace's `npm pack`'d artifact into an app workspace's local node_modules folder for production deployment without a bunch of extra prerelease/registry rigmarole. In a sense treat the monorepo/workspace root as a kind of ephemeral local registry

### Example

A monorepo with three workspaces, two apps "a" and "b", one package "c". Apps "a" and "b" depend on package "c", and are deployed to different locations (Docker, AWS Lambda, Google Cloud Functions, etc). Following best practices, minimizing build tooling, and wanting to ship a lean, production-only, app artifact, I'd like to an app structure like this (the dist folder is just meant to represent some production-ready code, this is just to show a folder structure that expresses the relationship between packages):

```sh
$ tree apps/a --dirsfirst
apps/a
├── dist
├── node_modules
│   └── @my
│       └── c
│           ├── dist
│           └── package.json
└── package.json
```

With this structure I'd zip the apps/a folder and push it to AWS, or `COPY` it into a Docker image in my Dockerfile, etc.

#### Current Behaviour

This structure can be roughly achieved by running these commands:

```sh
npm pack -w @my/c
npm pack -w @my/c --pack-destination=apps/a
npm i my-c-1.0.0.tgz -w @my/a
rm my-c-1.0.0.tgz apps/a/my-c-1.0.0.tgz
```

**Note:**

I'm not sure why it's necessary to pack it twice, but something about running `npm i my.tgz -w @my/a` needs the tarball to exist in both the root and the workspace. Maybe a separate bug about resolving install paths when `-w` is set?

Also worth noting this changes apps/a/package.json's `"@my/c"` dependency from `"^1.0.0"` to `"file:my-c-1.0.0.tgz"` which I have to immediately undo in order to maintain the dev-time hoisted dependency functionality I want.

#### Desired Behaviour

Allow a way to specify that `npm install` and/or `npm ci` should look to the workspaces in the current monorepo before reaching out to a registry. If a match is found call `npm pack` on that workspace and "serve" the resultant tarball (or its contents). Some suggestions for what this'd look like in the cli:

1. `npm i --only=prod --prefer-workspaces -w @my/a`
[...]

PR #493, opened 2021-11-15
docs: add npm copy rfc

everett1992:
`npm copy <destination>` copies the current project's files and dependencies to
`destination`

When deploying a project (`COPY` into a docker image, `zip` into an archive)
non production dependencies and files should be excluded. npm has some commands
that cover this usecase (`prune`, `pack`, `install -g` but they each have
issues, especially when workspaces are used. 

```
# copy app's production dependencies and packaged files to out
npm copy out --production -w app

# copy all workspace production dependencies and packaged files to archive
npm copy out --production --workspaces
```

See [0000-npm-copy.md](https://github.com/everett1992/rfcs/blob/npm-copy/accepted/0000-npm-copy.md)

I've written an implementation here everett1992/cli#1

### DIFF
--- a/accepted/0000-npm-copy.md
+++ b/accepted/0000-npm-copy.md
@@ -1,0 +1,100 @@
+# npm copy <destination>
+
+## Summary
+
+`npm copy <destination>` copies the current project's files and dependencies to
+`destination`.
+
+## Motivation
+
+When deploying a project (`COPY` into a docker image, `zip` into an archive)
+non production dependencies and files should be excluded. npm has some commands
+that cover this usecase (`prune`, `pack`, `install -g` but they each have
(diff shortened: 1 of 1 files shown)


everett1992:
I  think this RFC addresses npm/rfcs#463 Multi-app Monorepo Support

bnb:
Seems like a nice command. Just reiterating what I said in the RFC, want to make sure this works well for non-workspace users in addition to workspaces (which it seems to!).

jeanbmar:
I can't wait for this to land. Packaging workspaces for deployment is a guaranteed headache.

darcyclarke:
Removing `Agenda` label as we have discussed this work & agree we'd like to see it land. We can ratify this doc at any point here.

PR #519, opened 2022-02-02
RFC: Package Distributions

darcyclarke:
- adds RFC for new `distributions` capabilities

[See rendered RFC](https://github.com/npm/rfcs/blob/package-distributions/accepted/0000-package-distributions.md)

### DIFF
--- a/accepted/0000-package-distributions.md
+++ b/accepted/0000-package-distributions.md
@@ -1,0 +1,176 @@
+# Package Distributions
+
+## Summary
+
+Today, maintainers utilize various strategies to distribute platform-specific versions of their software under a singular package namespace. Often, these strategies rely on `install` scripts or `optionalDependencies` with some type of bootloader implementation (ex. [`esbuild`](https://npmjs.com/package/esbuild?activeTab=explore)); borth are not ideal. The `npm` CLI should support a first-class/standard way for maintainers to define conditions in which a package distribution is reified in place of the origin target. 
+
+## Motivation
+
+As we continue progress towards, & focus on, creating reproducible installations/builds, we should actively reduce the **need** for bespoke package distribution implementations & install scripts.
+
+## Detailed Explanation
+
+Introduce a new field called `distributions` which will be utilized by the `npm` CLI to add & then conditionally reify a package in place of the initially named version (if satisfied).
+
+## Rationale and Alternatives
+
+### Optional Dependency Fallback:
+
+The current best practice/work-around is to define all package distributions as `optionalDependencies` & rely on the environment failing to install them - then testing to see which dep successfully installed & use accordingly.
+
+#### Example:
+```json
+{
+  "name": "foo",
+  "optionalDependencies": {
+    "foo-win": "2.x",
+    "foo-darwin": "2.x",
+    "foo-android": "2.x"
+  }
+}
+```
+
+#### foo-android `package.json
+```json
+{
+  "name": "foo-android",
+  "os": [
+    "android"
+  ]
+}
+```
+
+## Implementation
+
+### Goals:
+- Limit the amount of net-new concepts & scope (ie. try to use existing paradigms & building blocks as much as possible)
+- Make this feature opt-in (at least initially)
+- Implement in a way that allows for graceful degredation (ie. fallbacks/polyfills)
+
+### Overview:
+- `distributions` will be an `Array` of objects defining conditions where a different package will be refieid & linked as the original target
+- `package-lock.json` will include _all_ distributions alongside the origin package (as it would for `optionalDependencies` today), but **only one** of these will be **reified on disk** ([as a link](#arborist-reify))
+- `distributions` **cannot conflict** with one another or have different dependencies within the tree (ex. it is not possible to have one distribution with a peer on `react@15` and another on `react@16`)
+- the **initial** conditional fields available will include the existing environment information `npm` already supports (ie. [`platform/os`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#os), [`arch/cpu`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#cpu), [`engines`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines))
+- `npm` will create the _same_ tree when `distributions` are found; the only thing that will change is that the package matching the conditions is reified in place
+
+#### Example
+
+```json
+{
+  "name": "foo",
+  "version": "1.2.3",
+  "main": "index.js",
+  "scripts": {
+    "preinstall": "node-gyp rebuild"
+  },
+  "distributions": [
+    {
+      "engines": {
+        "node": "10"
+      },
+      "platform": "win32",
+      "package": "foo-native-win32-10@1.x"
+    },
+    {
+      "platform": "linux",
+      "arch": "x64",
+      "package": "foo-native-linux-x64@2.x"
+    },
+    "..."
+  ]
+}
+```
+
+### `Arborist.buildIdealTree`:
+
+* Place _all_ distribution targets as optional<sup>(or peer?)</sup> of the `foo` package
+* Any that cannot be placed are not placed (eg, package not found, cannot be placed for conflicting peer deps, etc.), similar to `optionalDependencies`
+* Record status of `idealTree` placed distributions of the main "foo" package, along with their selection criteria
+- the `idealTree` **should not** be platform-specific
+
+### `Arborist.reify`:
+
+* Choose one distribution of the main package, and reify that
+* **Do not** reify the others, or any of their dependencies (unless they are required to meet another dependency in the tree)
+* Reify the main package as a `Link` to the chosen distribution
+
+### For Publisher's:
+
+* Publish the main package with the source, pointing to distribution package specifiers and selection criteria
+* Publish pre-built distributions as needed
+
+This is amenable to publishing distributions post-hoc as a CI build.  If any fail to build and publish, no matter, they will simply fail to install, and fall back to the main package.
+
+### For Consumer's:
+
+* Add a `foo` as a dependency
+* Pre-built distributions will be added to lockfile.
+* `require('foo')` will return the appropriate context-specific distribution if one was found successfully, or the original package if it was not.
+
+> **Note:** There should be no chance of an exponential explosion in lockfile size, as we calculate every possible combination of distribution matrices.
+
+## Use Cases
+
+### Polyfills:
+While this is most useful for slow and costly binary builds, it is also interesting for providing polyfills for node features.
+
+#### Example:
+
+```json
+{
+  "name": "fs-readdir",
+  "version": "fs.readdir() guaranteed to have withFileTypes:true support",
+  "distributions": [
+    {
+      "engines": { "node": "<v10.11.0"},
+      "package": "fs-readdir-polyfill@1"
+    },
+    {
+      "package": "fs-readdir-native@1"
+    }
+  ]
+}
+```
+
+#### fs-readdir-polyfill `package.json`
+
+```json
+{
+  "name": "fs-readdir-polyfill",
+  "version": "1.2.3",
+  "description": "polyfill the fs.readdir withFileTypes using fs.stat"
+}
+```
+
+#### fs-readdir-native `package.json`
+
+```json
+{
+  "name": "fs-readdir-native",
+  "version": "1.2.3",
+  "description": "just export require('fs').readdir"
+}
+```
+
+> Notably, we could just as easily have the `fs-readdir-polyfill` define `fs-readdir-native` as a distribution when `"engines": {"node": ">=10.11.0" }`
+
+## Prior Art
+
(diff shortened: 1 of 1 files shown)


accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
...
+  "scripts": {
+    "preinstall": "node-gyp rebuild"
+  },
+  "distributions": [
+    {
+      "engines": {
+        "node": "10"
+      },
+      "platform": "win32",
+      "package": "foo-native-win32-10@1.x"
```
ljharb, line 73: does this mean that "distributions" only points to additional packages? How will this impact dependency graph analysis tools?

accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
...
+* Reify the main package as a `Link` to the chosen distribution
+
+### For Publisher's:
+
+* Publish the main package with the source, pointing to distribution package specifiers and selection criteria
+* Publish pre-built distributions as needed
+
+This is amenable to publishing distributions post-hoc as a CI build.  If any fail to build and publish, no matter, they will simply fail to install, and fall back to the main package.
+
+### For Consumer's:
```
ljharb, line 105:
```suggestion
### For Publishers:

* Publish the main package with the source, pointing to distribution package specifiers and selection criteria
* Publish pre-built distributions as needed

This is amenable to publishing distributions post-hoc as a CI build.  If any fail to build and publish, no matter, they will simply fail to install, and fall back to the main package.

### For Consumers:
```

accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
...
+  "name": "fs-readdir",
+  "version": "fs.readdir() guaranteed to have withFileTypes:true support",
+  "distributions": [
+    {
+      "engines": { "node": "<v10.11.0"},
+      "package": "fs-readdir-polyfill@1"
+    },
+    {
+      "package": "fs-readdir-native@1"
+    }
```
ljharb, line 131: what condition applies here?

ljharb (changes requested):
Is "engines" the only condition?

What about when someone only wants the "browser" version of the package?

What about the use case of "i want the default package to omit tests, but i want a way to opt in to installing with tests"?

accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
+# Package Distributions
+
+## Summary
+
+Today, maintainers utilize various strategies to distribute platform-specific versions of their software under a singular package namespace. Often, these strategies rely on `install` scripts or `optionalDependencies` with some type of bootloader implementation (ex. [`esbuild`](https://npmjs.com/package/esbuild?activeTab=explore)); borth are not ideal. The `npm` CLI should support a first-class/standard way for maintainers to define conditions in which a package distribution is reified in place of the origin target.
```
styfle, line 5:
```suggestion
Today, maintainers utilize various strategies to distribute platform-specific versions of their software under a singular package namespace. Often, these strategies rely on `install` scripts or `optionalDependencies` with some type of bootloader implementation (ex. [`esbuild`](https://npmjs.com/package/esbuild?activeTab=explore)); both are not ideal. The `npm` CLI should support a first-class/standard way for maintainers to define conditions in which a package distribution is reified in place of the origin target. 
```

accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Prior Art
+
+- Yarn's **Package Variants** Proposal: https://github.com/yarnpkg/berry/issues/2751
+
+## Unresolved Questions and Bikeshedding
+
+- Do we need to feature flag this for `npm@8`? ex. put this feature under a new flag (ex. `--with-distributions`)? Does that limit it's impact/usage?
+- Should we outline best practices? 
+  - ex. a best practice we could recommend for maintainers to ensure they're consumers are using `distributions` properly, & to avoid confusion, is to set `engines` value for `npm` & educate maintainers/consumers on `--engines-strict`
```
styfle, line 166:
```suggestion
  - ex. a best practice we could recommend for maintainers to ensure their consumers are using `distributions` properly, & to avoid confusion, is to set `engines` value for `npm` & educate maintainers/consumers on `--engines-strict`
```

accepted/0000-package-distributions.md
lydell, line 5: Emphasis mine:

> Often, these strategies rely on `install` scripts **or** `optionalDependencies` with some type of bootloader implementation (ex. [`esbuild`](https://npmjs.com/package/esbuild?activeTab=explore))

In the case of `esbuild` it should be **and:** It uses _both_ `optionalDependencies` _and_ a `postinstall` script:

https://github.com/evanw/esbuild/pull/1621

> * There is still a small post-install script but it's now optional in that the `esbuild` package should still function correctly if post-install scripts are disabled (such as with `npm --ignore-scripts`). This post-install script optimizes the installed package by replacing the `esbuild` JavaScript command shim with the actual binary executable at install time. This avoids the overhead of launching another `node` process when using the `esbuild` command. So keep in mind that installing with `--ignore-scripts` will result in a slower `esbuild` command.

https://github.com/evanw/esbuild/blob/049e765d95532055f87b1655196c3dc4505b72cf/lib/npm/node-install.ts#L150-L195

For a tool like `esbuild` – which can compile a small project in a couple of milliseconds – it’s a bit funny that the JavaScript wrapper around the executable can take more time than the compilation itself. If you only run `esbuild` once in a while it doesn’t matter super much of course.

But that extra startup time does add up in the Elm ecosystem. Editor plugins run both [Elm](https://github.com/elm/compiler) and [elm-format](https://github.com/avh4/elm-format/) on save, and there you want an as snappy an experience as possibly, so shaving 100 ms or so definitely counts.

Many people install `elm-format` with npm locally in each project (so that every contributor is on the exact same version, which is very important for formatters). So a JavaScript wrapper around the project hurts!

So – is this something that this RFC seeks to improve on too? (Just checking!) In my opinion it would be awesome to somehow achieve maximum performance and `--ignore-scripts` with no “shenanigans” to pull from package authors!

Jolg42:
Hi there! Really happy to see this RFC and can provide an example from how we currently achieve the distribution of binaries at [Prisma](https://www.prisma.io/)

We support `binaryTargets`, see https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#binarytargets-options 
The list looks like this 
```ts
[
  'darwin',
  'darwin-arm64',
  'debian-openssl-1.0.x',
  'debian-openssl-1.1.x',
  'rhel-openssl-1.0.x',
  'rhel-openssl-1.1.x',
  'linux-arm64-openssl-1.1.x',
  'linux-arm64-openssl-1.0.x',
  'linux-arm-openssl-1.1.x',
  'linux-arm-openssl-1.0.x',
  'linux-musl',
  'linux-nixos',
  'windows',
  'freebsd11',
  'freebsd12',
  'openbsd',
  'netbsd',
  'arm',
]
```

- We build our binaries for each of these "targets"
- You can also build your own engine, for example
	- if you "forked" or want to build it yourself
	- if we don't build for your "target"

The download of the binaries is achieved on `postinstall` of the [`@prisma/engines` package](https://www.npmjs.com/package/@prisma/engines), which fetches the binaries from CloudFront/S3.

Here you can see there are quite a few requirements to replace this with `Package Distributions`
- platform (covered by RFC)
- arch (covered by RFC)
- openssl version (for example, `rhel-openssl-1.0.x` is needed for AWS lambda)
- Escape hatch to provide your own custom-built binaries (not sure how this is covered at the moment)

Happy to be part of a next meeting if you find this interesting 😃

Jolg42:
I forgot to mention that we also need in some cases to install more than one "binaryTarget" locally.

Example:
For AWS Lambda a zip archive needs to prepared with what you want to deploy.
If you are on macOS and are deploying a Lambda like this you then require 2 "binaryTarget", 
- `darwin` if you want to execute the code locally on macOS 
- `rhel-openssl-1.0.x` so that it can be added in the zip archive.

styfle:
There are two features we need for Next.js that `optionalDependencies` doesn't currently provide:

1. Support for libc detection: `glibc` vs `musl`. Today we need a custom install script to run [detect-libc](https://www.npmjs.com/package/detect-libc) and install the appropriate binary. This is necessary for every Linux install.
2. Support for WASM fallback. Today we need a custom install script to run and detect if the current platform fails to match known distributions with prebuilt binaries. This is necessary for CodeSandbox/StackBlitz/etc.

Hopefully `distributions` can solve these problems.

Brooooooklyn:
Besides detecting engines and other conditions automatically, also we need a mechanism to install dependencies for the other platform for cross-compiling:

Refs:

- https://github.com/napi-rs/node-rs/issues/376
- https://github.com/coder/code-server/issues/4804

accepted/0000-package-distributions.md
```diff
@@ -0,0 +1,176 @@
...
+    {
+      "engines": {
+        "node": "10"
+      },
+      "platform": "win32",
+      "package": "foo-native-win32-10@1.x"
+    },
+    {
+      "platform": "linux",
+      "arch": "x64",
```
Brooooooklyn, line 77: Don't forget `libc` here

saquibkhan:
- can check with vercel and esbuild
- can we achieve the same results without this feature
- will this help to remove post install script

Brooooooklyn:
## Feedback from users of next.js

- Installation size is too large because of the libc field is not supported: https://github.com/vercel/next.js/issues/42641
- Platform-specific optional dependencies not being included in package-lock.json when reinstalling with node_modules present

## Feedback from users of NAPI-RS

### Noway to install prebuild packages for another platform.

This is really important for Electron developers and Docker users.

For example, users need to build their Windows ia32 distribution Electron app on Windows x64 and they need to install prebuild packages for windows-ia32 platform on it.
Some users want to installation packages on Linux x64 gnu host so they can reuse the cache on it, then add the installed node_modules into node:lts-alpine (Linux x64 musl) containers to build their production image.
In cross platform testing CI, developers want to install packages on host platform because it’s fast and can be benefited from the cache on the host. Install the packages is extremely slow in the qemu enumerate. https://github.com/Brooooooklyn/snappy/blob/main/.github/workflows/CI.yaml#L428-L444

Yarn 3 is the only solution for this case at present: https://yarnpkg.com/configuration/yarnrc#supportedArchitectures.

There are some issues mentioned about it:
https://github.com/napi-rs/node-rs/issues/376
https://github.com/swc-project/swc/issues/2898
https://github.com/coder/code-server/issues/4804

### Install wasm fallback without postinstall script

Always including the wasm fallback package as dependency will increase the installation size. There is no way to install the wasm fallback package without postinstall script when all the platform specified optionalDependencies are not suitable for the host platform.

What we are doing for this case is to use postinstall script to detect if the native binaries are installed successfully, if not, we will install the wasm fallback package in the postinstall script.

### ~~Install the fallback package contains sources and provide a way to rebuild it~~

**It could be solved by [0049-link-packages-to-source-and-build](https://github.com/npm/rfcs/blob/main/accepted/0049-link-packages-to-source-and-build.md)**

https://github.com/napi-rs/napi-rs/issues/1474

Like wasm package, we also can’t contain the sources for prebuild binary in the package by default due to the installation size concern.
But some developers who are working for Large international companies complain about dependencies on prebuild native binaries; because the compliance policy at their company wanted to audit the source code (and how can we prove that the downloaded binary was built from the code?).
So we need a way to publish the sources & infra codes for the prebuild binaries, but not install for users by default.

ljharb:
@saquibkhan the use case I find valuable here is if I could make the "default" distribution for my package omit tests, source files, docs, etc, but provide some kind of "full" distribution that includes all those things.

The beneficial install bandwith impact of applying this to my 300-400 packages _alone_ would likely save npm millions of dollars a year, conservatively.

nlf:
i've made some updates to this document including several unanswered questions as well as detailing a known risk and adding in some other alternatives

JoshuaKGoldberg:
👋 @darcyclarke we can see that you closed this on Dec 1 but it's not clear why. https://github.com/npm/statusboard/issues/647 says to refer to this PR for information. Is there somewhere that says _why_ this RFC was closed? Thanks!

darcyclarke:
@JoshuaKGoldberg I apologize for the wait but I was weighting a response against leaning into the exact situation I was trying to avoid (ie. being asked for updates on `npm` isuess/prs/features). I left GitHub/npm in 12/22 & am focusing my energy on [`vlt`](https://github.com/vltpkg). I'm still interested in this feature but will be prioritizing its concept/shape as part of `vlt` moving forward. If the team wants to pick this work up independently from what we're doing there, they should fork the PR or create a new proposal (unfortunately, I don't know if this is still the forum where product direction is derived from though). The `statusboard` project you linked is a quasi-open backlog for the team but I don't know how accurate all of the issues in it are today (as I made many of them). [Work is still being tracked there to some extent](https://github.com/npm/statusboard/issues/488) but you'll have to ask the current team for any updates/clarifications on the state of issues/prs/features.

PR #522, opened 2022-02-04
Option for npm diff to ignore cr at eol

oBusk:
An option to make `npm diff` more useful when working on packages developed
across multiple platforms.

This is my first suggestion. Please let me know if it needs clarification or
improvement!

### DIFF
--- a/accepted/0000-diff-ignore-cr-at-eol.md
+++ b/accepted/0000-diff-ignore-cr-at-eol.md
@@ -69,8 +69,12 @@ PR to add this.
 * As mentioned previously 
   [`git diff --ignore-cr-at-eol`](https://github.com/git/git/commit/e9282f02b2f21118f3383608718e38efc3d967e1)
 
+* Gnu diffutils 
+  [`diff --strip-trailing-cr`](https://www.gnu.org/software/diffutils/manual/html_node/diff-Options.html#diff-Options)
+
 ## Unresolved Questions and Bikeshedding
 
-* Option/flag name? `diffIgnoreCRAtEOL`, `diffIgnoreLineEndings`, `diffIgnoreCr`?
+* Option/flag name? `diffIgnoreCRAtEOL`, `diffIgnoreLineEndings`, `diffIgnoreCr`, 
+  `diffStripTrailingCr`?
 * Should `ignoreAllSpace` ignore CR?
 * Should CR be ignored by default?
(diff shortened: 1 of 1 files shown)


ruyadorno:
Thanks for the contribution @oBusk! We chatted about it in our weekly OpenRFC calls (you're able to catch up with the recordings if you click on the linked issues here) and my feedback there is that this was the plan all along 😁 Launch with a minimal set of options enabled and let's wait for the community to surface what other options from `git diff` might be useful in `npm diff`. With that in mind I believe this is a natural RFC for us to accept.

It looks like you already got started on a PR in `jsdiff` to enable that (ref: https://github.com/kpdecker/jsdiff/pull/344) it looks great and will make it very easy to add support to this option in npm.

I believe this is ready to land. So to answer to the questions at the end:

> Option/flag name?

I'd stick with `--diff-ignore-cr-at-eol` as the npm config value and `diffIgnoreCRAtEOL` for internal `libnpmdiff`

> Should ignoreAllSpace ignore CR?

I believe they're two different things but to be honest, we're probably just going to follow the default behavior that `jsdiff` provides, in case it changes there someday we would not do any extra work to avoid it.

> Should CR be ignored by default?

I don't think so, it might be very useful to have that metadata surfaced in some cases and it seems to be a more "correct" default behavior. Anyone else might opt into the option in their `.npmrc` files.

ruyadorno:
Approved and landed in 25e4b087980fa78dd344f8f9f0f67528313bf9d0

Thank you so much @oBusk!

oBusk:
@ruyadorno that's so cool 😊 I had a look at both this and last weeks meeting and I agree with the notion of trying to keep npm diff to match git diff where possible. In the latest meeting it was mentioned a question of what the default in `git diff` is so I tried to have a look in the small [test repo](https://github.com/oBusk/jsdiff-crlf) I setup and gnu diffutils behaved as I expected it, but honestly I can't get `git diff` to show CRLF differences but since the `--ignore-cr-at-eol` flag is not a toggleable option but a opt-in flag, I still assume the default behaviour is to highlight CR vs CRLF differences 🤔

ruyadorno:
> since the `--ignore-cr-at-eol` flag is not a toggleable option but a opt-in flag, I still assume the default behaviour is to _not_ ignore CR vs CRLF differences

yeah, I'm assuming the same!

PR #525, opened 2022-02-08
Stop storing `integrity` for git dependencies

nlf:
This is relevant for
 - npm/cli#2846
 - npm/cli#2971


ljharb:
Would there be value in storing the git sha, since the URL might not contain it?

nlf:
> Would there be value in storing the git sha, since the URL might not contain it?

the sha is already what gets stored in the `resolved` field today

bnb:
will restate what I said in the meeting, half in jest but also half seriously: if this feature isn't working in a consistent way that accomplishes what it says it does, entirely removing it is imo potentially a candidate for a patch release.

derTobsch:
We run into the same problem today and now our ci pipline is stuck, because it does calculate a different hash than any other development machine. If we can help to test this issue, please let us know.

nlf:
npm@8.5.2 no longer compares the stored integrity field to the generated data

PR #528, opened 2022-02-11
RFC: Shared Version Specifications

boginw:
Provide a mechanism for users to specify reusable version specifications for multiple dependencies. For example, many dependencies are published with the same version. This RFC introduces a syntax for specifying the version in a single place and then allowing reusing it across dependencies.

### DIFF
--- a/accepted/0000-shared-version-specs.md
+++ b/accepted/0000-shared-version-specs.md
@@ -1,0 +1,160 @@
+# Shared Version Specifications
+
+## Summary
+
+Provide a mechanism for users to specify reusable version specifications for multiple dependencies. For example, many dependencies are published with the same version. In such cases, it would be handy for a user to specify the version in a single place and then reuse it across dependencies.
+
+## Motivation
+
+Developers sometimes intend for several packages to have the same version. These packages usually compose a framework or some other larger structure. A framework usually has a version, and each of its composing packages uses that version as their package version. For instance, JavaScript frameworks like "Angular" and "React" publish several packages with the same version. In such cases, it would be handy for a user to specify the version in a single place and then reuse it across dependencies.
+
+Furthermore, tools like [Dependabot](https://github.com/dependabot) create Pull-Requests that update dependencies to the latest versions. In some environments, like Gradle, where sharing version specifications is supported, these tools can reason about versions and update several dependencies at once. However, for NPM, these tools cannot infer that framework packages should be upgraded together. Allowing for reusing versions specifications can enable these tools to reason which dependencies should be updated together.
+
+## Detailed Explanation
+
+This proposal includes adding a new field in `package.json` named `versions` to achieve the desired result. This entry maps a name to a version. We would then resolve package specs with the following format `versions:<name>` to the entry under `versions` with the name `<name>`.
+
+### Example
+
+Let us say that we have a framework, aptly named, framework. This framework has two packages, "@framework/core" and "@framework/common". The latest version of the framework is "~1.2.3", and we would like for both of these packages to use that version, then we could specify the version like so:
+
+```json=
+  ...
+  "versions": {
+      "framework": "~1.2.3"
+  },
+  "dependencies": {
+      "@framework/core": "versions:framework",
+      "@framework/common": "versions:framework",
+      ...
+  }
+  ...
+```
+
+We could update both packages when a new version rolls out by updating the entry "framework" under "versions".
+
+## Rationale and Alternatives
+
+This section lists alternatives to the proposed solution and their implications. 
+
+### Referring to other dependencies' versions
+
+Instead of introducing the version field into `package.json`, one could introduce the ability for a dependency's version to be referred to by other dependencies. This could look something like the following:
+
+```json=
+  ...
+  "dependencies": {
+      "@framework/core": "~1.2.3",
+      "@framework/common": "refer:@framework/core",
+      ...
+  }
+  ...
+```
+
+The issue here is twofold: First, it can get messy with many dependencies across multiple dependency types ("dependencies," "devDependencies," "peer dependencies," and so on). Second, it might lead to circular dependencies. Consider the following example:
+
+```json=
+  ...
+  "dependencies": {
+      "A": "refer:B",
+      "B": "refer:A",
+      ...
+  }
+  ...
+```
+
+However, the implementation impact of this solution would be the same as with the proposed solution.
+
+### Overrides
+
+The resolution of packages in "overrides" can be changed to achieve the desired effect. Take, for instance, the following `package.json` example extract:
+
+```json=
+  ...
+  "dependencies": {
+      "@framework/core": "~1.2.3",
+      "@framework/common": "*",
+      ...
+  },
+  "overrides": {
+      "@framework/core": "$@framework/core",
+      "@framework/common": "$@framework/core"
+  }
+  ...
+```
+
+Updating framework packages with overrides would only require updating a single string. However, overriding would affect the entire tree, which may not be desired. Furthermore, as this is not the intended use of the override feature, using overrides for this purpose should not be considered.
+
+### External tooling
+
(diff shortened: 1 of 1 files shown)


ljharb:
Publishing dependent versions like this is something that many consider a bad practice, because it forces releases when none would otherwise be needed, confuses dependency graphs, and causes massive costs in terms of churn and bandwidth to the ecosystem. There’s a reason that projects like Babel moved away from this approach. I don’t think it’d be a good thing for npm to tacitly endorse this approach.

boginw:
Maybe there is something I am missing here, so feel free to correct me, but I cannot see how this would force releases. These "shared versions" are to be used when defining dependencies, not in publishing packages. I would understand if frameworks felt incentivized to publish new versions of packages that did not need updates, but they would never be forced to. Furthermore, the dependency graphs should be strictly equal after implementing this proposal. Wrt. to Babel, I could not find any reference to what you are referring to. Maybe you could point me in the right direction?

It reads to me like there is some misunderstanding in what this proposal is. This might be me, who hasn't given an apt description of what it is. In short, this proposal should be considered a layer of syntactic sugar, which converts the following:

```json=
  ...
  "versions": {
      "framework": "~1.2.3"
  },
  "dependencies": {
      "@framework/core": "versions:framework",
      "@framework/common": "versions:framework",
      ...
  }
  ...
```

To the following:

```json=
  ...
  "dependencies": {
      "@framework/core": "~1.2.3",
      "@framework/common": "~1.2.3",
      ...
  }
  ...
```

ljharb:
ok - so you’re just suggesting version numbers be substitutable from a map of alias names elsewhere in package.json?

This only seems like it’d provide value *when* packages are using dependent versioning. In other words, most of the time you *shouldnt* have more than one package whose version is pegged like this (even DefinitelyTyped packages won’t likely track the patch number).

boginw:
Yes, exactly that :)

Most of the time you would not use this feature, as most packages are not using version numbers that follow each other. In my experience with Gradle, such shared versions are usually only used when I depend on multiple packages from a framework, such as [Spring Boot packages](https://mvnrepository.com/artifact/org.springframework.boot), most of which use the same version as the "framework". 

But this is also useful in NPM. Take for instance [@angular](https://www.npmjs.com/search?q=%40angular) where around 50 packages use the same version (our projects usually use around 15 of them). It saves countless CI minutes and PRs for us, if, instead of updating these packages separately, they would update together all at once.

The versions referred to should be as exact as the referred version says they should be. By that I mean, if the latest version of "@framework/core" is "1.2.3" and the latest version of "@framework/common" is "1.2.4" and if the version alias is "~1.2.3" then the version "1.2.3" and "1.2.4" should be fetched respectively.

Version aliases might also be a better name than shared versions :)

ljharb:
right, but the better thing to do is for angular *not to do that* :-)

thescientist13:
I'm pretty sure the main reason for why Angular (and other tools like Babel) do this, is because they are monorepos that publish for same version for all packages at publish time, as opposed to independent package versioning.  Thus when upgrading, they often recommend to keep all versions the same in _package.json_.

ljharb:
Right - but that's a problematic approach, because it forces a publish of packages even when nothing has changed - and worse, it could force a minor version when nothing was added, or a major version when nothing was broken.

thescientist13:
Fair, but having managed both ways, in very closely knit monorepos, it can be very useful when communicating to consumers just upgrade to `x.y.z` as opposed to having potentially complicated combinations of inter-mixed packages, as opposed to tracking / documenting all the possible permutations of which packages / versions can work with other combinations of packages / versions.

Anyway, with the package manager it is usually a non-issue so personally ambivalent to this RFC but just adding an anecdote to the conversation.

boginw:
> edit: That said, for projects like this, I usually just leverage the package manager, like in Yarn, this would / should update them for me in one go

Doing a scoped upgrade might not always be desired. Organizations might have several frameworks under the same scope, or there might be one package that does not follow the version of the rest, where one might not want to upgrade. Furthermore, not all of these frameworks are scoped, take for instance [react](https://www.npmjs.com/package/react).

> Right - but that's a problematic approach, because it forces a publish of packages even when nothing has changed - and worse, it could force a minor version when nothing was added, or a major version when nothing was broken.

I cannot see how this would force anyone to publish packages. They won't be any more forced than they already are. Sure, the authors might be incentivized to do so, as maybe some of their users will complain, but this wouldn't force anything. And regardless these authors already publish packages under the same version.

ljharb:
@thescientist13 that it’s slightly more work to manage things properly doesn’t mean it’s a good idea to add burden for users in exchange for avoiding the proper work :-)

ljharb:
@boginw because versions can only be kept in sync if a version is published for every package - but not every package will have changes.

boginw:
Well, I don't have insights into how big an issue that would be for the registry. I have my doubts that much would change since as far as I know, most of those who publish packages with dependent versions are already publishing new versions of packages where nothing has changed. You could argue, that these publishers should not do that, but they do.

I would like to add, that this feature would lead to fewer PRs, fewer CI minutes used along with fewer MBs of artifacts and logs, and less work, across all repositories, using this feature in conjunction with tools like Dependabot.

thescientist13:
> that it’s slightly more work to manage things properly doesn’t mean it’s a good idea to add burden for users in exchange for avoiding the proper work

Where is the burden on users in this scenario though?  I think it's a reasonable argument to make that the lowest burden for users _and_ maintainers with an ecosystem like say angular or babel is to keep all versions in lockstep and following server.

```sh
"@angular/cli": "^12.0.1",
"@angular/core": "^12.0.1",
"@angular/http": "^12.0.1",
"@angular/etc": "^12.0.1",
```

vs

```sh
"@angular/cli": "^12.0.1",
"@angular/core": "^7.2.4",
"@angular/http": "^15.0.0",
"@angular/etc": "^1.0.1",
```

With the second one, there now becomes a not direct relationship between ecosystem packages and what they support at the sibling level and that the (relative) effort of tracking and documenting that ever expanding graph of interconnectedness would be just as burdensome to both parties, as opposed to just saying: "everyone upgrade to  `x.y.z`".

----

That aside, your comment seems to imply that maintainer's should not have agency for what they feel is the best way to manage their ecosystem for themselves and / or users?  I get that _you_ personally have a preference, but is said preference a standard documented somewhere that these projects are willfully neglecting?  I am sincere here as I am not aware there is actually a one size fits all solution to versioning so any education on the topic here would be useful in the context of this RFC I think.  I get the point you make about "phantom" releases but does reducing burden on the maintainer not account for anything in today's OSS landscape either?  What is so horrible about that versus having to make user's do all the leg work themselves by having to look up and cross reference packages across a bunch of independent semver iterations?  And that's assuming the docs are not out of date.

Again, I'm not necessarily advocating for this RFC, but maybe I'm just too naive in thinking that (especially mature) projects can figure out how to best manage the distribution and versioning of packages for themselves and their community based on their own goals, needs, and preferences? 🤷‍♂️

ljharb:
@thescientist13 im not saying maintainers should be prevented from doing this; I’m saying npm shouldn’t encourage it by adding a feature like this.

darcyclarke:
I'm in favor of _fixing_ `overrides` (that's right, I think it's a bug this isn't already supported) to support direct dependency swaps/overrides with references (which would give you what you want, as [referenced here](https://github.com/npm/rfcs/blob/55ee796c1d4f0fbfc97b8d83b7496cdf5b39033b/accepted/0000-shared-version-specs.md#overrides)).

We can probably backlog that work without having to accept this RFC (since direct dependency overrides were in scope originally & actually outlined as a [motivation in _that RFC_](https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md#motivation) & shouldn't be throwing an error today - the fact we added references via `$<pkg>` is just the cherry on top which then makes this possible).

boginw:
Due to the issues with this RFC relating to encouragement, that might be the best compromise. I would be very happy with that outcome

PR #547, opened 2022-03-04
add "obey user specifier" RFC

ljharb:
When I type `npm install package@specifier`, a normalized `specifier` is what should be saved to `package.json`, no more, no less.

## Motivation

Currently, if i type `npm install --save-dev eslint@=8.0.0 other`, it installs v8.8.0, but saves `^8.8.0` to `package.json` (with the default `save-prefix` of `^`). (note: i'm intentionally not using `--save-exact` here, because i want `other` to install using the default `save-prefix`).

When no `@version` is provided, or when a dist-tag is provided, it makes *perfect* sense to resolve that to a version, and save it with `save-prefix`. However, when I've provided a version range - `=1.2.3`, `~1.2.3`, `>= 1.2` - I know what I want, I’ve explicitly told npm I want it, and npm should obey my wishes not just in the reification in `node_modules` (which I believe it does), but also in what it saves to `package.json`.

### DIFF
--- a/accepted/0000-obey-user-specifier.md
+++ b/accepted/0000-obey-user-specifier.md
@@ -1,0 +1,33 @@
+# obey the user's specifier
+
+## Summary
+
+When I type `npm install package@specifier`, a normalized `specifier` is what should be saved to `package.json`, no more, no less.
+
+## Motivation
+
+Currently, if i type `npm install --save-dev eslint@=8.0.0 other`, it installs v8.8.0, but saves `^8.8.0` to `package.json` (with the default `save-prefix` of `^`). (note: i'm intentionally not using `--save-exact` here, because i want `other` to install using the default `save-prefix`).
+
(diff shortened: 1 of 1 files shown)


ruyadorno:
agreed 👍 if the user is specifying a valid semver prefix in along with the package spec when running `npm install` we should use **that** prefix instead of respecting the `save-prefix` config option.

killagu:
In some case, save tag is expected. Antd mark the stable version with `conch` dist-tag, when i install antd with `npm i antd@conch`, i also want save it with `conch` tag.

And `latest-x`, `beta`, `next` is also i want to save.

wesleytodd:
> with npm i antd@conch, i also want save it with conch tag.

I think there are two modes of operation here, and I would like to see both continued to be supported. Since resolving a version from a dist-tag has explicit behavior today I would love to see a new flag to achieve this behavior, which might need to be a separate RFC, I don't know if @ljharb (or the npm team) would want to bundle that together into this one since it already has agreement.

ljharb:
I agree they're two different things - one is this RFC, which is that the version specifier range should be obeyed; the other is "save the tag vs resolve the tag", and i agree there should be a flag for that.

PR #550, opened 2022-03-10
RFC: Improve signature verification

feelepxyz:
Add a new CLI command `audit signatures` that verifies the npm signatures in a packages packument. It works on the current install.

Signatures are only useful if people verify them. Signature verification is currently a [manual and complex process](https://docs.npmjs.com/verifying-the-pgp-signature-for-a-package-from-the-npm-public-registry), involving the Keybase CLI to fetch the npm public key.

Our long-term goal for supply chain security is that all software is signed and verified in a transparent and user-friendly way.

Signing and verifying published packages protects users against a malicious mirror or proxy intercepting and tamptering with the package response (MITM attack). Mirrors and proxyies are common in the npm ecosystem, significantly increasing the possible attack surface.

## References
Related to https://github.com/npm/rfcs/pull/76

### DIFF
--- a/accepted/0000-validate-signatures.md
+++ b/accepted/0000-validate-signatures.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Add a new top-level CLI command to verify registry signatures.
+Add a new top-level CLI command to verify registry signatures and update existing signatures from PGP to ESCDA.
 
 ## Motivation
 
@@ -20,17 +20,22 @@ increasing the possible attack surface.
 
 ## Detailed Explanation
 
-Introduce a new CLI command `verify-signatures` that verifies the npm signature
-in a packages packument, using npm's public key, fetched from npmjs.org. It
-works on the current install (`node_modules`), and checks all direct and
-transitive dependencies.
+npm [already signs each published version](https://blog.npmjs.org/post/172999548390/new-pgp-machinery.html) using a private PGP key, with the [public key hosted on Keybase](https://keybase.io/npmregistry).
+
+This RFC proposes improving existing registry signatures:
+- 1. Update the signing key to produce ECDSA signatures which are smaller and can be verified with node's `crypto` library
+- 2. Make signature verification easy using a new CLI command, dropping the requirement for 3rd party tools
+
+To make signature verification easy, we'll introduce a new CLI command `verify-signatures` that verifies the npm signature
+in a packages packument, using npm's public key, fetched from registry.npmjs.org. It
+works on the current install (`node_modules`), and checks all direct and transitive dependencies.
 
 The aim is for this command to plug into users build workflows after `npm ci/npm
 install` and block builds with invalid signatures.
 
 This command is standalone but we could fold this behaviour into `npm install`
 or `audit` in the future once we're confident validation is performant and
-provides a good user experience with little to no noise.
+provides a good user experience.
 
 ### Detailed CLI examples
 
@@ -69,7 +74,7 @@ NOTE: Signatures are missing for all packages hosted outside registry.npmjs.org
 ├─┬ abbrev@3.0.9
 │ ├── INVALID SIGNATURE bar@2.88.0
 
-Someone might have tampered with the package since it was hosted
+Someone might have tampered with the package since it was published
 on npmjs.org (monster-in-the-middle attack)!
 
 Please report this issue: https://github.com/npm/cli/issues/new/choose
@@ -87,7 +92,7 @@ $ npm verify-signatures once
 1 package has a invalid signature (direct):
 ├── INVALID SIGNATURE once@1.4.0
 
-Someone might have tampered with the package since it was hosted
+Someone might have tampered with the package since it was published
 on npmjs.org (monster-in-the-middle attack)!
 
 Please report this issue: https://github.com/npm/cli/issues/new/choose
@@ -108,7 +113,7 @@ project@1.0.0 $HOME/work/project
 ├─┬ lorem@0.4.0
 │ ├── INVALID SIGNATURE ipsum@2.0.0
 
-Someone might have tampered with the package since it was hosted
+Someone might have tampered with the package since it was published
 on npmjs.org (monster-in-the-middle attack)!
 
 Please report this issue: https://github.com/npm/cli/issues/new/choose
@@ -127,7 +132,7 @@ project@1.0.0 $HOME/work/project
 1 package has an invalid signature:
 ├── INVALID SIGNATURE once@1.4.0
 
-Someone might have tampered with the package since it was hosted
+Someone might have tampered with the package since it was published
 on npmjs.org (monster-in-the-middle attack)!
 
 Please report this issue: https://github.com/npm/cli/issues/new/choose
@@ -146,7 +151,7 @@ project@1.0.0 $HOME/work/project
 1 package has an invalid signature (direct):
 ├── INVALID SIGNATURE once@1.4.0
 
-Someone might have tampered with the package since it was hosted
+Someone might have tampered with the package since it was published
 on npmjs.org (man-in-the-middle attack)!
 
 Please report this issue: https://github.com/npm/cli/issues/new/choose
@@ -175,7 +180,7 @@ keys hosted on npmjs.com.
 
 We plan to rotate the signing key as part of this effort. The CLI command will
 support a new (ECDSA) signature key, existing PGP signatures and verification
-workflows will contine work during a deprecation period.
+workflows will contine work during a 6 month deprecation period.
 
 The new ECDSA signatures are smaller in size (64 bytes), compared to the
 existing PGP signature (893 bytes). We can stop writing `npm-signature`s on new
@@ -201,19 +206,13 @@ package releases once we the Keybase key has expired.
   will will only support new (ECDSA) signature keys and come pre-bundled with
   valid `keyid`s, warning users to update the cli if a new key has been found
   from npmjs.org.
-- 6. **Update the Keybase PGP key expiry**: We want to give folks with their own
-  tooling as much notice as possible while deprecating the use of the existing
-  PGP key. We will update the public Keybase key with a new expiration date (the
-  current key expires 2034-03-30). We will communicate this expiration to match
-  the deprecation timeframe (exact timing is TBD).
+- 6. **Update the [Keybase PGP key](https://keybase.io/npmregistry) expiry**: We will update the expiry to be 6 months from the launch of this CLI command and publish of announcement blog post. We want to give folks with their own tooling as much notice as possible while deprecating the use of the existing PGP key.
 - 7. **Stop generating PGP signatures once the key expires**: We will no longer
   populate the packument field `dist.npm-signature` in new releases.
 
 #### Breaking changes
 
-Existing workflows validating PGP signatures will break once the public Keybase
-PGP expires and we stop writing this signature to packages `npm-signature`
-field.
+Existing workflows validating PGP signatures will stop working once the public Keybase PGP expires and we stop writing this signature to packages `npm-signature` field.
 
 ### Adding ECDSA signatures to packuments
 
--- a/accepted/0000-validate-signatures.md
+++ b/accepted/0000-validate-signatures.md
@@ -498,7 +498,17 @@ Are you sure you trust this host? (yes/no)
 Trusted keys are added to the `.npmrc`:
 
 ```
-trusted-signature-keyid[]="{{SHA256_PUBLIC_KEY}}"
+[trusted-signature-keys.sha256:{{SHA256_PUBLIC_KEY}}]
+expires=2023-12-17T23:57:40-05:00
+keytype=ecdsa-sha2-nistp256
+sigtype=ecdsa-sha2-nistp256
+key=123
+
+[trusted-signature-keys.sha256:{{SHA256_PUBLIC_KEY}}]
+expires=2023-12-17T23:57:40-05:00
+keytype=ecdsa-sha2-nistp256
+sigtype=ecdsa-sha2-nistp256
+key=789
 ```
 
 ## Out of scope
--- a/accepted/0000-validate-signatures.md
+++ b/accepted/0000-validate-signatures.md
@@ -7,8 +7,7 @@ Add a new top-level CLI command to verify registry signatures.
 ## Motivation
 
 Signatures are only useful if people verify them. Signature verification is
-currently a [manual and complex process
-(https://docs.npmjs.com/verifying-the-pgp-signature-for-a-package-from-the-npm-public-registry),
+currently a [manual and complex process](https://docs.npmjs.com/verifying-the-pgp-signature-for-a-package-from-the-npm-public-registry),
 involving the Keybase CLI to fetch the npm public key.
 
 Our long-term goal for supply chain security is that all software is signed and
--- a/accepted/0000-validate-signatures.md
+++ b/accepted/0000-validate-signatures.md
@@ -40,7 +40,7 @@ The proposed command is standalone but the behaviour could be folded into `npm
 install` or `audit` in the future, once we're confident validation is performant
 and provides a good user experience.
 
-### Threat Model
+### Threat model
 The threat model assumes the following:
 
 - Attackers cannot compromise npm's signing key stored online
--- a/accepted/0000-validate-signatures.md
+++ b/accepted/0000-validate-signatures.md
@@ -2,7 +2,8 @@
 
 ## Summary
 
-Add a new top-level CLI command to verify registry signatures and update existing signatures from PGP to ESCDA.
+Add a new top-level CLI command to verify registry signatures and update
+existing signatures from PGP to ESCDA.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


ljharb:
What signatures in a packument? If i can use npm to verify them, how can i use npm to generate them?

ljharb:
Separately, I’d love some motivating examples of actual npm incidents where package signing would have prevented or mitigated the incident, to ensure that this doesn’t end up being security theater.

feelepxyz:
> What signatures in a packument? If i can use npm to verify them, how can i use npm to generate them?

@ljharb 👋  The current `dist.npm-signature` field that's [already in the packument](https://registry.npmjs.org/light-cycle/1.4.3) is generated on npm's servers when the package is published using npm's private PGP key. This proposal suggests updating to use a new/more modern key, and bake in validation in the cli. Third-party registries could go also generate signatures on publish and populate the packument in the same way.

Supporting user-generated build time signatures is also something we're actively investigating but not as part of this work.

ljharb:
I’m a bit confused. You’re saying that all public npm registry packages have this signature automatically?

ljharb:
Why then would this require an opt-in, instead of just running as part of the install?

feelepxyz:
> > Separately, I’d love some motivating examples of actual npm incidents where package signing would have prevented or mitigated the incident, to ensure that this doesn’t end up being security theater.

Three are no directly related attacks on npm that I'm aware of but I've found some similar incidents where users downloaded malicious content from a download server/mirror that would have been mitigated by signing and verifying:
- [handbrake](https://github.com/cncf/tag-security/blob/main/supply-chain-security/compromises/2017/handbrake.md)
- [elmedia](https://github.com/cncf/tag-security/blob/main/supply-chain-security/compromises/2017/elmedia.md)
- [keydnap](https://github.com/cncf/tag-security/blob/main/supply-chain-security/compromises/2016/keydnap.md)
- [fosshub](https://github.com/cncf/tag-security/blob/main/supply-chain-security/compromises/2016/fosshub.md)
- [phpMyAdmin corrupted copy on Korean mirror server](https://sourceforge.net/blog/phpmyadmin-back-door/)

Arguably, this proposal is only fixing what's already there (registry signatures), by making signature validation easy, it's not brining something fundamentally new to the table.

There is a whole other host of attack vectors that are not covered with this change that I'd love us to tackle in future. Particularly the ones outlined in [TUF Attacks and Weaknesses](https://theupdateframework.io/security/#attacks-and-weaknesses).

> Why then would this require an opt-in, instead of just running as part of the install?

Yes good point. I think doing validation on install is something we should aim for but adding it as a separate command to begin with will allow us to experiment with the user experience a bit more freely, compared to `npm install`. Once we're happy with the overall ergonomics we could fold it into `install`.

There are a few prickly edge cases with validation that we'd need to consider a bit more carefully for doing it on install:

- While we can enforce validation on all packages fetched from npmjs.org as they will have signatures set, we can't do the same for packages fetched from third-party registries until they add support. Users either need to manually trust keys or be warned some packages are unsigned (these warnings might be unwanted if a user didn't explicitly ask to verify).
- AFAIK there's no way to reliably differentiate between a npmjs.org mirror and a third-party registry where users host their own packages, so we might need to resort to showing a warning when verifying installed packages that have "missing" signatures (the mirror could try circumventing validation by just omitting the `signatures` array from the served packument. 
- While we can use node's `crypto` library to do validation, it still uses OpenSSL installed on the machine, so there's a chance this versions is way out of date and doesn't support the elliptic curves we need (`prime256v1` in OpenSSL). We can collect feedback from users running this validation command, and error with a useful message if OpenSSL needs updating.

MylesBorins:
Re: OpenSSL

nodejs vendors all dependencies including OpenSSL. So while it is possible to use a system version, that is an edge case and not the norm for systems running node.

ljharb:
An extra command for this seems harmless on the surface, but it really seems like it adds minimal value - since it’s confined to public registry packages, pending support in third-party registries), and unless it can actually mitigate a MitM attack, which seems only plausible if someone has hijacked your DNS *and* there’s no cert pinning/HSTS involved.

I’m assuming here that it would literally never fail for any package on the public registry that made it down to the user unmodified?

MylesBorins:
@ljharb I think it makes sense to reframe this as "cleaning up existing debt" rather than thinking of anything as net new.

1. We have signatures already, this improved them
2. We have a [documented way to verify them](https://docs.npmjs.com/verifying-the-pgp-signature-for-a-package-from-the-npm-public-registry), this drops the requirement for 3rd party tools

as @feelepxyz pointed out above this is not by any means the "end" to the improvements or work we want to do around signatures... but the alternative imho is removing signatures altogether, not avoiding this work. Is that what you are advocating for?

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,510 @@
...
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+3 packages have invalid signatures (1 direct, 2 transitive):
+├── INVALID SIGNATURE once@1.4.0
+├─┬ lorem@0.4.0
+│ ├── INVALID SIGNATURE ipsum@2.0.0
+├─┬ abbrev@3.0.9
+│ ├── INVALID SIGNATURE bar@2.88.0
+
+Someone might have tampered with the package since it was hosted
```
trevrosen, line 72: I think we should tweak some of the wording here a bit, as `since it was hosted` is a bit ambiguous to me.

Side question: do we do i18n for the CLI, or is English hard-coded?

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,510 @@
...
+$ npm verify-signatures once
+
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!!  WARNING: SOME PACKAGES DOWNLOADED FROM NPMJS.ORG HAVE INVALID SIGNATURES  !!
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+1 package has a invalid signature (direct):
+├── INVALID SIGNATURE once@1.4.0
+
+Someone might have tampered with the package since it was hosted
```
trevrosen, line 90: I don't want to write the same comment on each instance of `since it was hosted` - how should we be capturing these messages for review? Is this the canonical place where we're deciding that I guess?

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,510 @@
...
+
+Please update the npm CLI to version x.x.x in order to support the new signature
+keys hosted on npmjs.com.
+```
+
+### Moving signatures from PGP to ECDSA
+
+We plan to rotate the signing key as part of this effort. The CLI command will
+support a new (ECDSA) signature key, existing PGP signatures and verification
+workflows will contine work during a deprecation period.
```
trevrosen, line 178: Do we have a proposed length of the deprecation period? Is it six months?

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,510 @@
...
+   }
+```
+
+The `keyid` will map to a public key hosted on `https://registry.npmjs.org/.well-known/npm-signature-keys`.
+
+## Rationale and Alternatives
+
+We propose a pragmatic improvement to the existing signatures and verification
+flow. We're not proposing any support for [signed
+packuments](https://github.com/npm/rfcs/pull/76) or user generated signatures
```
trevrosen, line 249: I feel like we might want to expand this statement to accommodate our discussions around the notion that some form of "signed packument" **is** an appropriate mitigation for the "evil mirror" vector.

trevrosen (changes requested):
Very thorough, glad we're getting granular! Only hard request I have is to make sure that we have the exact deprecation and expiry timeframes called out very explicitly in this document.

ljharb:
@MylesBorins i'm not advocating for anything just yet; i'm trying to understand the use case and implications.

It seems like verifying the signatures automatically on install - as opposed to providing an explicit command for it - and ignoring any packages that lack signatures, would be a no-harm improvement; and then perhaps those who want to fail on packages that lack signatures could use a config option to do so?

By shipping a separate command, it is implied that using it increases security, ie, that it protects against an actual attack. I haven't yet heard any actual argument that it does so, and I think security theater is worse than no security at all (ie, i think removing the signatures would be a more secure outcome than providing this as a separate command _if_ verifying the signatures didn't actually prevent anything).

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,509 @@
...
+  `https://registry.host/.well-known/npm-signature-keys`
+- Configure the endpoint to return the public key and key metadata (see payload
+  example under "Fetching the public key")
+- Start signing package versions on publish, adding the signature (`sig`) and
+  `keyid` to the version packument, nested in the `dist.signatures` array (see
+  example under "Packument example with new `signatures` field")
+
+#### Verifying third-party registry signatures using the npm CLI
+
+There's currently no reliable (or is there?) way to differentiate a mirror/proxy
```
phillmv, line 422: assuming this got ascertained 😉 

```suggestion
There's currently no reliable way to differentiate a mirror/proxy
```

accepted/0000-validate-signatures.md
```diff
@@ -0,0 +1,509 @@
...
+Found untrusted signature keys at https://fast-npm-mirror.tld/.well-known/npm-signature-keys
+
+Are you sure you trust this host? (yes/no)
+> no
+```
+
+Trusted keys are added to the `.npmrc`:
+
+```
+[public-signature-key]
```
phillmv, line 468: can you define more than one trusted key using this notation in the `npmrc`? (I've never written one).

phillmv: (Is that even desirable? Maybe not.)

phillmv (commented):
This is a really good technical document! I really like how you can see all the conversations we've had throughout it, the focus on attack vectors and the weighing of options.

phillmv:
> By shipping a separate command, it is implied that using it increases security, ie, that it protects against an actual attack. I haven't yet heard any specific argument that it does so

Hi @ljharb!

The cli will ship with a copy of the public key, so by default an attacker could no longer MITM packages. @feelepxyz talks about this in the [Motivation section](https://github.com/npm/rfcs/pull/550/files#diff-782c86d46c6887c2cff4151d2f70204ec360a1ae65d09516e01d5c94c27d3e73R16) 🙂.

--

You've mentioned concerns around security theatre, and I think that's appropriate! I think it's correct that this work will not, by itself, add a lot of trust to the registry.

This work will basically only protect you against from someone running a fake mirror that injects its own code into the packages you download.

But it does allow us to start adding new and better trust guarantees over time!

ljharb:
@phillmv thanks for the context. so you're saying that this helps the case where someone MitMs the public registry (which, ftr, would be a drastically reduced attack surface if the npm cli used HSTS/cert pinning), and also tries to modify the contents of public packages?

If so, then I agree it's valuable, and provides a real (if minor) security benefit. However, why require a separate command? Why not just do this by default as part of `npm install`/`npm ci`? Who _wouldn't_ want this? (assuming it doesn't fail on packages with missing keys, which could be an additional config to enable)

MylesBorins:
> Why not just do this by default as part of npm install/npm ci

@ljharb I think this is a step towards that, but the amount of scrutiny that `npm install` gets for performance + stability I think it would be pre-mature to introduce this functionality into the critical path so quickly.

We could introduce a flag for install and setting to support this functionality 🤔 , but with how npm works that would sadly just fail silently on older version that don't support the flag 🙃

ljharb:
Gotcha. ok so, the proposal is for a separate command, that could presumably be configured to fail, or ignore, packages that lack a signature (failing on it would probably break everyone that has an internal registry), with an eventual eye for including it in `npm install`?

If so, thanks for bearing with me, and that sounds great. I'd suggest `npm audit signatures` or something, to pair with `npm audit licenses` :-)

accepted/0000-validate-signatures.md
feelepxyz: Oh this syntax isn't correct, it should be `public-signature-key[]=`, but maybe worth naming it something like `trusted-signature-key[]=` would be clearer what this is 🤔  Also thinking we might not want to include all the attributes here as you might want to be able to update expiry on the key to rotate it etc. Maybe just `keyid` and `registry`. You'll want to be able to set multiple keys, for different registries.

This syntax is similar to how you can configure `ca`'s https://docs.npmjs.com/cli/v8/using-npm/config#ca

feelepxyz:
> where someone MitMs the public registry (which, ftr, would be a drastically reduced attack surface if the npm cli used HSTS/cert pinning), and also tries to modify the contents of public packages?

Disappointingly, this won't protect you against the public registry being MITM'd (yet) as we'd loose guarantees that we're getting the right public keys. It protects you against the case where you are pulling packages from a npm proxy/mirror of the public registry that have been tampered with. Some companies run a mirror that they install packages from. Verifying signatures would verify that the package you got was the same one that was uploaded to npm.

This is arguably an incredibly niche attack vector, but we're planning a bunch of follow up work that will be transparent to users. One particular improvement is actually protecting against the public registry being MITD'd or worse, compromised entirely.

Yarn is an another example where they run their own proxy in front of the npm registry. So if Yarn CLI started verifying npm registry signatures they would give users some more guarantees that the package installed is the same one as the one published to the npm registry.

> Gotcha. ok so, the proposal is for a separate command, that could presumably be configured to fail, or ignore, packages that lack a signature (failing on it would probably break everyone that has an internal registry), with an eventual eye for including it in `npm install`?

Yes exactly 💯 I believe starting with a separate command will give us some wiggle room to figure out what configuration is needed and what a good default will be for most users, that could then be included in `install` by default.

> I'd suggest `npm audit signatures` or something, to pair with `npm audit licenses` :-)

Thanks for pointing this out! I had no idea about `npm audit licenses`, will investigate this.

accepted/0000-validate-signatures.md
feelepxyz: Yup, moved this up under `Detailed explanation > Attack vector`

accepted/0000-validate-signatures.md
darcyclarke: @feelepxyz `ini` does support Objects combining to form Arrays although if we want nested value setting it does look like the syntax has to change a bit.

Here's an example of a valid/supported syntax we could use:

```ini
[public-signature-key.SHA256:{{SHA256_PUBLIC_KEY}}]
expires=2023-12-17T23:57:40-05:00
keytype=ecdsa-sha2-nistp256
sigtype=ecdsa-sha2-nistp256
key={{B64_PUBLIC_KEY}}
```

In this case, you'd have to specify a unique key id (but that's already a requirement of the spec) & then the keys would be grouped into an Array of values/config properly.

Feel free to try this out:
```js
const ini = require("ini")
let text = `
[public-signature-key.SHA256:123]
expires=2023-12-17T23:57:40-05:00
keytype=ecdsa-sha2-nistp256
sigtype=ecdsa-sha2-nistp256
key=123

[public-signature-key.SHA256:456]
expires=2023-12-17T23:57:40-05:00
keytype=ecdsa-sha2-nistp256
sigtype=ecdsa-sha2-nistp256
key=456

[public-signature-key.SHA256:789]
expires=2023-12-17T23:57:40-05:00
keytype=ecdsa-sha2-nistp256
sigtype=ecdsa-sha2-nistp256
key=789
`
console.log(ini.parse(text))
```

feelepxyz: @darcyclarke thanks for pointing this out! I think is what we want 👍  I see that it gets nested in the object as expected when running `npm config list --json` 👌 :

```
"public-signature-key": {
  "SHA256:123": {
    "expires": "2023-12-17T23:57:40-05:00",
    "keytype": "ecdsa-sha2-nistp256",
    "sigtype": "ecdsa-sha2-nistp256",
    "key": "123"
  },
  "SHA256:456": {
    "expires": "2023-12-17T23:57:40-05:00",
    "keytype": "ecdsa-sha2-nistp256",
    "sigtype": "ecdsa-sha2-nistp256",
    "key": "456"
  }
}
```

bnb:
In general, I do like this. The one caveat I do have is that I've often seen users want _author_ signing, as opposed to registry signing. My instinct here is that _most_ people will assume this is, in fact, verifying _author_ signing. I will admit that upon reading the comments before digging into the contents, this was absolutely what I thought this proposal was going for (and I was very confused why @ljharb wasn't sure of an example where author signing would have prevented an attack).

It would be extremely nice if whatever name we land on disambiguates that this is, in fact, signing happening by the npm registry and not authors. 

Alternatively, adding author signing and then having `npm audit signatures` do _both_ registry and author signing would be awesome.

> @darcyclarke is this scheduled for the 6th of April? I won't be able to make it that day, could we move it to the week after, 13th of April?

The meetings are weekly, so we can discuss it in the April 13th meeting :)

feelepxyz:
> It would be extremely nice if whatever name we land on disambiguates that this is, in fact, signing happening by the npm registry and not authors.

Yes I was initially thinking we should call this command `verify-registry-signature`. This is more explicit but also limits the scope of what this command can do in future.

> Alternatively, adding author signing and then having `npm audit signatures` do _both_ registry and author signing would be awesome. 

Agreed! This is in line with what we've been thinking about, hence the more generic command name. Once we release some form of author/build signing this command would also verify these signatures 👌 

Author or build signing would follow much later as it's a big piece of work that requires a lot of design that's still up in the air (we're keen to do better than author PGP signatures and looking at [sigstore/cosign](https://www.sigstore.dev/)).

The current plan is to go with a generic command like `npm audit signatures` or `npm verify-signatures` that only supported registry signatures to begin with, with the plan to also support author/build signatures in future.

@bnb do you have any objections to the above plan if we also make it more obvious what type of signatures are being verified in the command output?

bnb:
> @bnb do you have any objections to the above plan if we also make it more obvious what type of signatures are being verified in the command output?

If this is the plan, I'm chill with it. I think somehow communicating in CLI output that the current implementation is verifying registry-generated signatures _once you've run the command_, that'd be sufficent.

so, like:

```bash
npm audit signatures

> verifying registry-generated package signatures 1/1839
```

and later:

```bash
npm audit signatures

> verifying registry-generated package signatures: 1/1839
> verifying author-generated package signatures: 32/992
```

(Obviously not final UI/UX, but hopefully the point is clear!)

bnb:
this is absolutely outside of the scope of this RFC, but wanted to share because I thought it might be nifty:

I've also been a fan of third-party hooks into the CLI and making it a platform for some time now (nobody else has ever been, though), but it'd be cool if I could install a dependency (`npmDependencies` lol) from GitHub or Snyk or WhiteSource that also let them hook into the above and do their own signature checking:

```bash
npm audit signatures

> verifying registry-generated package signatures: 1/1839
> verifying author-generated package signatures: 32/992
> verifying GitHub Security package signatures: 23/1728
> verifying WhiteSource package signatures: 77/1431
```

feelepxyz:
> I think somehow communicating in CLI output that the current implementation is verifying registry-generated signatures _once you've run the command_, that'd be sufficent.

Yeah exactly 👍  I think we can make it clear in the output once you run it. Will try and incorporate this in the examples.

> I've also been a fan of third-party hooks into the CLI and making it a platform for some time now (nobody else has ever been, though), but it'd be cool if I could install a dependency (`npmDependencies` lol) from GitHub or Snyk or WhiteSource that also let them hook into the above and do their own signature checking:

Interesting! I had thought about supporting third-party npm registries, e.g. GitHub Packages, Artifactory, Verdaccio etc but this seems slightly different. Do you mean the dependency can define a hook/script that is run by the CLI? Do you have an example of how something like this looks?

bnb:
> Interesting! I had thought about supporting third-party npm registries, e.g. GitHub Packages, Artifactory, Verdaccio etc but this seems slightly different. Do you mean the dependency can define a hook/script that is run by the CLI? Do you have an example of how something like this looks?

Yeah, you basically nailed it. Given all this is about packages, leveraging that same package ecosystem to extend functionality helps make the CLI more of a platform. I think the closest example would be along the lines of VS Marketplace and VS Code Extensions, except instead of an Electron app it's just a CLI. There are obvious bounds that are needed (an example in VS Code is that certain rendering bits can't be messed with) but providing that sandbox allows for an ecosystem to grow and hook in the workflows you're already using with their products directly into the CLI.

feelepxyz:
Feedback from the [Open RFC Meeting](https://github.com/npm/rfcs/issues/567):
- We should aim to enable signature verification by default on `install` but seems sensible to launch it as a separate command to test it in the wild first, and do a fast follow to enable by default on `install`
- We should probably pick defaults on `install` that ignore missing signatures and just fail on signatures that explicitly fail verification against a public key
- We could give users options to turn on more strict checks for missing sigs etc when running `install`
- Investigate using `host/.well-known/npm-keys` URL prefix for the public key

ljharb:
In order to enable by default on install, perhaps we should first add a config flag that controls whether it does run on install? that way, the semver-major change could be flipping the default of that config.

owlstronaut (approved):

PR #564, opened 2022-03-30
RFC: Dependency Selector Syntax & `npm query`

darcyclarke:
[Rendered RFC](https://github.com/npm/rfcs/blob/3d5b2130504139bdc8a3b599923aa07d2ff79c96/accepted/0000-dependency-selector-syntax.md)* Updated

### Overview
- introduces new "Dependency Selector Syntax"
- introduces new `.querySelectorAll()` method to `Node`s returned from `Arborist`
- introduces new top-level command `npm query`

### DIFF
--- a/accepted/0000-dependency-selector-syntax.md
+++ b/accepted/0000-dependency-selector-syntax.md
@@ -195,7 +195,7 @@ There are several known attributes that are normalized & queryable living in `No
     "vulnerabilities": [],
     "cwe": []
   },
-  ...
+  …
 ```
 
 #### Usage
--- a/accepted/0000-dependency-selector-syntax.md
+++ b/accepted/0000-dependency-selector-syntax.md
@@ -24,6 +24,7 @@ This RFC's spec & implementation should closely mimic the capabilities of existi
 ### Dependency Selector Syntax
 
 #### Overview:
+
 - there is no "type" or "tag" selectors (ex. `div, h1, a`) as a dependency/target is the only type of `Node` that can be queried
 - the term "dependencies" is in reference to any `Node` found in the `idealTree` returned by `Arborist`
 
--- a/accepted/0000-dependency-selector-syntax.md
+++ b/accepted/0000-dependency-selector-syntax.md
@@ -7,8 +7,8 @@ Introduce a new `npm query` commmand which exposes a new dependency selector syn
 ## Motivation
 
 - Standardize the shape of, & querying of, dependency graphs with a robust object model, metadata & selector syntax
-- Leverage existing, known language syntax & operators from CSS to make disperate package information broadly accessible
-- Unlock the ability to anwser complex, multi-faceted questions about dependencies, their relationships & associative metadata
+- Leverage existing, known language syntax & operators from CSS to make disparate package information broadly accessible
+- Unlock the ability to answer complex, multi-faceted questions about dependencies, their relationships & associative metadata
 - Consolidate redundant logic of similar query commands in `npm` (ex. `npm fund`, `npm ls`, `npm outdated`, `npm audit` ...)
 
 ## Detailed Explanation
@@ -39,15 +39,6 @@ This RFC's spec & implementation should closely mimic the capabilities of existi
 - `>` direct decendent/child selector
 - `~` sibling selector
 
-#### [Attribute Selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)
-- `[]` attribute selector (ie. existence of attribute)
-- `[attribute=value]` attribute value is equivalant...
-- `[attribute~=value]` attribute value contains word...
-- `[attribute*=value]` attribute value contains string...
-- `[attribute|=value]` attribute value is equal to or starts with...
-- `[attribute^=value]` attribute value begins with...
-- `[attribute$=value]` attribute value ends with...
-
 #### Pseudo Selectors
 - [`:not(<selector>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)
 - [`:has(<selector>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)
@@ -60,7 +51,9 @@ This RFC's spec & implementation should closely mimic the capabilities of existi
 - `:link` when a dependency is linked
 - `:deduped` when a dependency has been deduped
 - `:override` when a dependency is an override
-- `:extraneous` when a dependency could be but is not deduped
+- `:extraneous` when a dependency exists but is not defined as a dependency of any node
+- `:invalid` when a dependency version is out of its ancestors specified range
+- `:missing` when a dependency is not found on disk
 - `:outdated` when a dependency is not `latest`
 - `:vulnerable` when a dependency has a `CVE` 
 - `:semver(<spec>)` matching a valid [`node-semver`](https://github.com/npm/node-semver) spec
@@ -68,40 +61,52 @@ This RFC's spec & implementation should closely mimic the capabilities of existi
 - `:realpath(<path>)` [glob](https://www.npmjs.com/package/glob) matching based on dependencies realpath
 - `:type(<type>)` [based on currently recognized types](https://github.com/npm/npm-package-arg#result-object)
 
-#### Generic Pseudo Selectors:
+#### [Attribute Selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)
 
-A standardized pseudo selector pattern will be used for `Object`s, `Array`s or `Arrays` of `Object`s accessible via `Arborist`'s `Node.package` metadata. Pseudo classes generated in this way will allow for iterative attribute selection.
+The attribute selector evaluates the key/value pairs in `package.json` if they are `String`s.
 
-`Array`s specifically use a special `value` keyword in place of a typical attribute name. `Arrays` also support exact `valu`e matching when a `String` is passed to the selector. See examples below:
+- `[]` attribute selector (ie. existence of attribute)
+- `[attribute=value]` attribute value is equivalant...
+- `[attribute~=value]` attribute value contains word...
+- `[attribute*=value]` attribute value contains string...
+- `[attribute|=value]` attribute value is equal to or starts with...
+- `[attribute^=value]` attribute value begins with...
+- `[attribute$=value]` attribute value ends with...
+
+#### `Array` & `Object` Attribute Selectors
+
+The generic `:attr()` pseudo selector standardizes a pattern which can be used for attribute selection of `Object`s, `Array`s or `Arrays` of `Object`s accessible via `Arborist`'s `Node.package` metadata. This allows for iterative attribute selection beyond top-level `String` evaluation.
+
+`Array`s specifically use a special `item` keyword in place of a typical attribute name. `Arrays` also support exact `valu`e matching when a `String` is passed to the selector. See examples below:
 
 #### Example of an `Object`: 
 ```css
 /* return dependencies that have a `scripts.test` containing `"tap"` */
-*:scripts([test~=tap])
+*:attr(:scripts([test~=tap]))
 ```
 
 #### Example of an `Array` Attribute Selection:
 ```css
 /* return dependencies that have a keyword that begins with "react" */
-*:keywords([value^="react"])
+*:attr(:keywords([item^="react"]))
 ```
 
 #### Example of an `Array` matching directly to a value:
 ```css
 /* return dependencies that have the exact keyword "react" */
 /* this is equivalent to `*:keywords([value="react"])` */
-*:keywords("react")
+*:attr(:keywords("react"))
 ```
 
 #### Example of an `Array` of `Object`s: 
 ```css
 /* returns */
-*:contributors([email="ruyadorno@github.com"])`
+*:attr(:contributors([email="ruyadorno@github.com"]))
 ```
 
 ### Classes
 
-Given that `Arborist` will control our understanding of the DOM (Dependency Object Model), claseses are predefined by their relationships to their ancestors & are specific to their dependency type. Dependencies may have, & are allowed to have, multiple classifications (ex. a `workspace` may also be a `dev` dependency).
+Classes are defined by the package relationships to their ancestors (ie. the dependency types that are defined in `package.json`). This approach is user-centric as the ecosystem has been taught to think about dependencies in these classifications first-and-foremost. Dependencies are allowed to have multiple classifications (ex. a `workspace` may also be a `dev` dependency & may also be `bundled` - a selector for that type of dependency would look like: `*.workspace.dev.bundled`).
 
 - `.prod`
 - `.dev`
@@ -109,39 +114,6 @@ Given that `Arborist` will control our understanding of the DOM (Dependency Obje
 - `.peer`
 - `.bundled`
 - `.workspace`
-
-### Attributes
-
-There are several known attributes that are normalized & queryable living in `Node.package` (aka. `package.json`).
-
-- `[name]`
-- `[version]`
-- `[description]`
-- `[homepage]`
-- `[bugs]`
-- `[author]`
-- `[license]`
-- `[funding]`
-- `[files]`
-- `[main]`
-- `[browser]`
-- `[bin]`
-- `[man]`
-- `[directories]`
-- `[repository]`
-- `[scripts]`
-- `[config]`
-- `[workspaces]`
-- `[dependencies]`
-- `[devDependencies]`
-- `[optionalDependencies]`
-- `[bundledDependencies]`
-- `[peerDependencies]`
-- `[peerDependenciesMeta]`
-- `[engines]`
-- `[os]`
-- `[cpu]`
-- `[keywords]`
 
 ### Command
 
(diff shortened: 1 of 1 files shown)


accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+## Summary
+
+Introduce a new `npm query` commmand which exposes a new dependency selector syntax (informed by & respecting many aspects of the [CSS Selectors 4 Spec](https://dev.w3.org/csswg/selectors4/#relational)).
+
+## Motivation
+
+- Standardize the shape of, & querying of, dependency graphs with a robust object model, metadata & selector syntax
+- Leverage existing, known language syntax & operators from CSS to make disperate package information broadly accessible
+- Unlock the ability to anwser complex, multi-faceted questions about dependencies, their relationships & associative metadata
+- Consolidate redundant logic of similar query commands in `npm` (ex. `npm fund`, `npm ls`, `npm outdated`, `npm audit` ...)
```
bnb, line 12: I think this is an extremely good goal to have. One thing I'd likely end up recommending is that the method to do this is exposed via an easy-to-use API to lower the barrier for others _outsite_ of the core npm CLI maintainers to be able to easily add commands via the query language in the future.

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Detailed Explanation
+
+This RFC's spec & implementation should closely mimic the capabilities of existing CSS Selector specifications. Notably, we'll introduce limited net-new classes, states & syntax to ensure the widest adoption & understanding of paradigms. When deviating, we'll explicitely state why & how.
+
+## Implementation
+
+- `Arborist`'s `Node` Class will have a new `.querySelectorAll()` method
+  - this method will return a filtered, flattened dependency Arborist `Node` list based on a valid query selector
+- Introduce a new command, `npm query`, which will take a dependency selector & output a flattened dependency Node list (output is in `json` by default, but configurable)
```
bnb, line 22: Perhaps worth including that the npm CLI should have an interface for commands to effectively be aliases for queries?

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+- the term "dependencies" is in reference to any `Node` found in the `idealTree` returned by `Arborist`
+
+#### Selectors
+    
+- `*` universal selector
+- `#<name>` dependency selector (equivalent to `[name="..."]`)
+- `#<name>@<version>` (equivalent to `[name=<name>]:semver(<version>)`)
+- `,` selector list delimiter
+- `.` class selector
+- `:` pseudo class selector
```
bnb, line 37: What is a class in this context? Might be worth including that in the overview bullets above.

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+## Implementation
+
+- `Arborist`'s `Node` Class will have a new `.querySelectorAll()` method
+  - this method will return a filtered, flattened dependency Arborist `Node` list based on a valid query selector
+- Introduce a new command, `npm query`, which will take a dependency selector & output a flattened dependency Node list (output is in `json` by default, but configurable)
+
+### Dependency Selector Syntax
+
+#### Overview:
+- there is no "type" or "tag" selectors (ex. `div, h1, a`) as a dependency/target is the only type of `Node` that can be queried
```
bnb, line 27:
```suggestion

- there is no "type" or "tag" selectors (ex. `div, h1, a`) as a dependency/target is the only type of `Node` that can be queried
```

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Implementation
+
+- `Arborist`'s `Node` Class will have a new `.querySelectorAll()` method
+  - this method will return a filtered, flattened dependency Arborist `Node` list based on a valid query selector
+- Introduce a new command, `npm query`, which will take a dependency selector & output a flattened dependency Node list (output is in `json` by default, but configurable)
+
+### Dependency Selector Syntax
+
+#### Overview:
```
bnb, line 26:
```suggestion
#### Overview:

```

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+- `.prod`
+- `.dev`
+- `.optional`
+- `.peer`
+- `.bundled`
+- `.workspace`
+
+### Attributes
+
+There are several known attributes that are normalized & queryable living in `Node.package` (aka. `package.json`).
```
bnb, line 114: is there a reason this isn't generic? For example, TypeScript has the [`volta`](https://github.com/microsoft/TypeScript/blob/main/package.json#L120) property in their package.json. It might be nice to be able to dig into other packages that have `volta` as well.

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+    "workspaces": {},
+    "keywords": [],
+    "ancenstry": "",
+    "path": "",
+    "realpath": "",
+    "parent": "",
+    "vulnerabilities": [],
+    "cwe": []
+  },
+  ...
```
bnb, line 197: markdown parsing isn't happy with the `...` here. IMO fine to remove it :P

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+
+// find all packages that have @ruyadorno as a contributor
+*:contributors([email=ruyadorno@github.com])
+```
+
+## Next Steps
+
+### Command Mapping to `query-output`
+
+Previous commands with similar behaivours will now be able to utilize `Aborist` `Node.querySelectorAll()` under-the-hood & will fast-follow the `npm query` implementation.
```
bnb, line 295: This is honestly one of the most exciting parts of this proposal. Hope to see this as soon as is possible :)

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,421 @@
...
+```bash
+npm audit # equivalent to...
+npm query ":root > *:vulnerable" --query-output audit
+
+npm audit --all # equivalent to...
+npm query "*:vulnerable" --query-output audit
+
+npm audit <pkg> # equivalent to...
+npm query "#<pkg>" --query-output audit
+```
```
bnb, line 351: AFAIK this would also allow for easy implementation of https://github.com/npm/rfcs/pull/182

accepted/0000-dependency-selector-syntax.md
darcyclarke: Generics are explained [here[(https://github.com/npm/rfcs/blob/e63094a66d3b690561c129599861f221ffb7305c/accepted/0000-dependency-selector-syntax.md#generic-pseudo-selectors). It may not be apparent (which we should fix) but the intent is that any information in a dependencies `package.json` will be queryable - there's just a subset of those fields that are "known" to us from the outset (aka. https://docs.npmjs.com/cli/v8/configuring-npm/package-json).

Querying for `volta` specific keys/values would look something like:

```bash
npm query "*[volta]"
npm query "*:volta([node=14.15.5])"
```

accepted/0000-dependency-selector-syntax.md
darcyclarke: Parser doesn't like it but it still renders it nicely... I'm okay with removing it though

accepted/0000-dependency-selector-syntax.md
darcyclarke: Happy your excited! I think your right that this outlines/legitimatizes the need for abstracting away the underlying functionality.

accepted/0000-dependency-selector-syntax.md
darcyclarke: 100%

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+
+#### Overview:
+
+- there is no "type" or "tag" selectors (ex. `div, h1, a`) as a dependency/target is the only type of `Node` that can be queried
+- the term "dependencies" is in reference to any `Node` found in the `idealTree` returned by `Arborist`
+
+#### Selectors
+    
+- `*` universal selector
+- `#<name>` dependency selector (equivalent to `[name="..."]`)
```
ljharb, line 34: what about globs? `array.prototype.*` or `@foo/*`? i see below that, matching DOM qSA, `[name^="@foo"]` would work, but it seems like it'd be nice to support shortcuts for common cases :-)

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+#### Pseudo Selectors
+- [`:not(<selector>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)
+- [`:has(<selector>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)
+- [`:where(<selector list>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where)
+- [`:is(<selector list>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is)
+- [`:root`](https://developer.mozilla.org/en-US/docs/Web/CSS/:root) matches the root node/dependency
+- [`:scope`](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope) matches node/dependency it was queried against
+- [`:empty`](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) when a dependency has no dependencies
+- [`:private`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#private) when a dependency is private
+- `:link` when a dependency is linked
```
ljharb, line 60: what about dependency category? prod, dev, bundled, optional, peer?

update: i see "class" below but i think that term is MASSIVELY confusing to apply here (anywhere, really, since it's so horrifically overloaded)

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+#### Generic Pseudo Selectors:
+
+A standardized pseudo selector pattern will be used for `Object`s, `Array`s or `Arrays` of `Object`s accessible via `Arborist`'s `Node.package` metadata. Pseudo classes generated in this way will allow for iterative attribute selection.
+
+`Array`s specifically use a special `value` keyword in place of a typical attribute name. `Arrays` also support exact `valu`e matching when a `String` is passed to the selector. See examples below:
+
+#### Example of an `Object`: 
+```css
+/* return dependencies that have a `scripts.test` containing `"tap"` */
+*:scripts([test~=tap])
```
ljharb, line 80: does this implicitly mean `<name>:<package.json field>(<selector>)`? if so, doesn't this prevent any future additions of unprefixed selectors here?

what about if i have a package.json field named `:realpath` i want to query on?

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+- `:semver(<spec>)` matching a valid [`node-semver`](https://github.com/npm/node-semver) spec
+- `:path(<path>)` [glob](https://www.npmjs.com/package/glob) matching based on dependencies path relative to the project
+- `:realpath(<path>)` [glob](https://www.npmjs.com/package/glob) matching based on dependencies realpath
+- `:type(<type>)` [based on currently recognized types](https://github.com/npm/npm-package-arg#result-object)
+
+#### Generic Pseudo Selectors:
+
+A standardized pseudo selector pattern will be used for `Object`s, `Array`s or `Arrays` of `Object`s accessible via `Arborist`'s `Node.package` metadata. Pseudo classes generated in this way will allow for iterative attribute selection.
+
+`Array`s specifically use a special `value` keyword in place of a typical attribute name. `Arrays` also support exact `valu`e matching when a `String` is passed to the selector. See examples below:
```
ljharb, line 75:
```suggestion
`Array`s specifically use a special `value` keyword in place of a typical attribute name. `Arrays` also support exact `value` matching when a `String` is passed to the selector. See examples below:
```

ljharb: why "value" and not "item"? will there be validation if i pass another attribute name besides "value" for an array?

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+*:keywords("react")
+```
+
+#### Example of an `Array` of `Object`s: 
+```css
+/* returns */
+*:contributors([email="ruyadorno@github.com"])`
+```
+
+### Classes
```
ljharb, line 102: what about selecting things installed via a registry (any, or a specific one) vs installed from git vs installed from a tarball URL?

accepted/0000-dependency-selector-syntax.md
ljharb: npm already conceptually knows about `type`, and `exports`, and `browser`, and webpack's `sideEffects`, and a bunch of others, but these aren't in the list.

accepted/0000-dependency-selector-syntax.md
ljharb:
```suggestion
  …
```

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,422 @@
...
+
+### Command Mapping to `query-output`
+
+Previous commands with similar behaivours will now be able to utilize `Aborist` `Node.querySelectorAll()` under-the-hood & will fast-follow the `npm query` implementation.
+
+#### `npm list`
+
+```bash
+npm list # equivalent to...
+npm query ":root > *"
```
ljharb, line 302: `npm ls` does validation of the graph, which `npm query` wouldn't do - do you mean that *some* of the output is equivalent to that?

accepted/0000-dependency-selector-syntax.md
darcyclarke: For attributes we're not going to support globs. We'll stick with the attribute operators/selectors that exist & use query selectors for any bespoke functionality.

accepted/0000-dependency-selector-syntax.md
darcyclarke: The classes we've defined map to the ecosystems understanding of dependency classification first-and-foremost. We're adding some language in an update to help showcase why we make this distinction.

accepted/0000-dependency-selector-syntax.md
darcyclarke: Reworking attributes here in an update

accepted/0000-dependency-selector-syntax.md
darcyclarke: There will not be any validation. That said, I can change the name to item for you

accepted/0000-dependency-selector-syntax.md
darcyclarke: Yes, we need to resolve this. Liekly, we'll use a scoped pseudo selector to avoid these collisions (ie. `:attr()`)

accepted/0000-dependency-selector-syntax.md
darcyclarke: No. You're right, it's not 100% equivalent given that the above wouldn't throw on error.

accepted/0000-dependency-selector-syntax.md
darcyclarke: This is just the name of this operator in CSS.

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,395 @@
...
+- the term "dependencies" is in reference to any `Node` found in the `idealTree` returned by `Arborist`
+
+#### Selectors
+    
+- `*` universal selector
+- `#<name>` dependency selector (equivalent to `[name="..."]`)
+- `#<name>@<version>` (equivalent to `[name=<name>]:semver(<version>)`)
+- `,` selector list delimiter
+- `.` class selector
+- `:` pseudo class selector
```
ljharb, line 38: i still think these need to be named without the word "class".

accepted/0000-dependency-selector-syntax.md
```diff
@@ -0,0 +1,395 @@
...
+#### `Array` & `Object` Attribute Selectors
+
+The generic `:attr()` pseudo selector standardizes a pattern which can be used for attribute selection of `Object`s, `Array`s or `Arrays` of `Object`s accessible via `Arborist`'s `Node.package` metadata. This allows for iterative attribute selection beyond top-level `String` evaluation.
+
+`Array`s specifically use a special `item` keyword in place of a typical attribute name. `Arrays` also support exact `valu`e matching when a `String` is passed to the selector. See examples below:
+
+#### Example of an `Object`: 
+```css
+/* return dependencies that have a `scripts.test` containing `"tap"` */
+*:attr(:scripts([test~=tap]))
```
ljharb, line 85: would be good to add an example with a quoted key as well

hemanth:
It would be great to have Tab Completion on quires! [Should that also be an RFC in itself?]

nickmccurdy:
`npm query` has already been released.

PR #566, opened 2022-04-06
RFC: Command Specific Configuration

darcyclarke:
[Rendered RFC](https://github.com/npm/rfcs/blob/1d909afaf3b406e1261bffe7dcd1a0fdbaabbefa/accepted/0000-command-specific-config.md)

### DIFF
--- a/accepted/0000-command-specific-config.md
+++ b/accepted/0000-command-specific-config.md
@@ -1,0 +1,59 @@
+# Scoped Command Config
+
+## Summary
+
+Implement a mechanism to scope configuration to commands explicitely.
+
+## Motivation
+
+This change will help resolve many issues user's face when configuring `npm` as the current strategy is to have a single set of global options available to **all** commands. 
+
+## Detailed Explanation
+
+- expand `.npmrc` to support command-specific configuration
+- changes should be as backwards compatible as possible (old `.npmrc` files should still be supported)
+- add clear documentation on configuration precedence (ex. arguments vs. command-specific vs. project, user & global vs. globally set vs. `publishConfig`)
+
+## Rationale and Alternatives
+
+Currently, the only alternative strategy users have is to manual pass arguments to `npm`.
+
+## Implementation
+
+- `npm` should support `npm-` prefixed command configuration in `.npmrc` files
+- prefixed command configuration scopes the nested config
+
+### Example `.npmrc`
+
+```ini
+omit=dev
+
+# equivalent to `npm install --no-audit --no-fund`
+[npm-install]
+audit=false
+fund=false
+
+# equivalent to `npm audit --production`
+[npm-audit]
+production=true
+
+# equivalent to `npm audit fix --force`
+[npm-audit-fix]
+force=true
+
+# equivalent to `npm list --all`
+[npm-list]
+all=true
+
+# equivalent to `npm publish --loglevel=silly`
+[npm-publish]
+loglevel=silly
+```
+
+## Prior Art
+
+- N/A
+
+## Unresolved Questions and Bikeshedding
+
+- N/A


ljharb (approved):
yes please, this should have been in v1 :-D

lukekarrys:
Can there be an optional parent scope for a "config version"? It wouldn't be tied directly to the npm version but it would allow us to making breaking config changes and still honor older versions.

Concerns:
- I'm not sure how to make this scope unambiguous when comparing with a command scope.
- I'm not sure if there is a way to reset the scope in `ini` files

```ini
[config-version=3]

[npm-audit]
fix=true

[config-version=4]

[npm-audit]
fix=false
```

ljharb:
I don't think it'd be possible to have more than one config version at a time? since config can come from env vars, CLI args, and any of 4 possible npmrc files (5 if you include child workspaces), i think it's got to be an all or nothing thing.

dominykas:
😍 

Does that mean we'd be able to do

```
[npm-install]
ignore-scripts=true
```
?

How would that interact with other related commands, e.g. `npm ci`?

Might be out of scope here, but I guess `npm update` is somewhat different, but it does run `install` scripts or does it not?

darcyclarke:
@dominykas yes, your example is correct. 

> How would that interact with other related commands, e.g. npm ci?

This is a good question. IMO, your example shouldn't effect `ci`'s behavior because it's an independent command (ref: https://github.com/npm/cli/tree/latest/lib/commands) but I could be swayed there _maybe_. There's probably other examples/caveats that should be considered & discussed (ex. `npm install` will run `npm pack` for git dependencies, if you have `[npm-pack]` specific config set, potentially that should be respected). Would love to discuss more in a call or async.

> Might be out of scope here, but I guess npm update is somewhat different, but it does run install scripts or does it not?

You're right, there may be a case where you want to do something for `install` & `update`; but I feel like you can just duplicate that config if that's what you want. I guess the question to ask is whether we have commands that invoke other commands (ex.  `it` & `cit`) that we should explicitly note either **do** or **don't** respect the command specific config for the subsequent commands they spawn.

ljharb:
It seems like they probably should - `npm publish` invokes `npm pack`, and i would expect any pack-specific config to be respected during a publish.

dominykas:
> I feel like you can just duplicate that config if that's what you want

Yeah, I agree - for both cases - `ci` and `update` and also for whatever executes other commands. The only concern here is documenting that and making sure that people who advocate a solution advocate the full solution... always hard :D And maybe there are ways to let the users know that e.g. they're running `npm ci` and some settings that apply to `npm install` will not be applied?

darcyclarke:
### Actions
- [ ] needs command & config audit (ie. documenting which commands call other commands & what config should or shouldn't be shared with them: ex. `npm install` calls `npm pack` for git deps)

coolaj86:
Why have a special ini-style syntax for a node program?

JSON.

Human readable.

Machine readable.

Easy to integrate with CI.

Easy to batch edit with `yq` (or a variety of other tools).

No special dependencies or parsers required.

No complexity from YAML.

No confusion with TOML.

You could easily auto-update existing .npmrc to .npmrc.json and never regret it.

I just happened across this due to https://github.com/npm/cli/issues/5279, but it seems an oversight that .npmrc was ever not JSON (especially considering that comments aren't preserved anyway) and no better time to fix it than before it's too late. :)

ljharb:
@coolaj86 npmrc has been ini since time immemorial and that's unlikely to change; it doesn't really matter why it's that way, but it simply is. It is about 12 years "too late" to make such a breaking change.

coolaj86:
>  It is about 12 years "too late"

Nonsense. The whole `--init-author-name` vs `init-author-name` vs `init.author.name` confusion is proving that it's never too late to make a breaking change.

And the whole `npm set` vs `npm config`.

And a dozen of other instances since at least v5 onwards.

npm makes breaking changes all the time.

Making a breaking change that requires _less_ custom code and that's easier to implement, verify, and automate will of course get the same backlash of any of the other breaking changes ([XKCD: Workflow](https://xkcd.com/1172/)), but think of all the CI systems that will get to drop a few dozen dependencies and all the supply chain attacks that will save us from, and all the automation it will make easier.

`init-author-name` isn't helping with any of that and it got in just fine.

nlf:
> think of all the CI systems that will get to drop a few dozen dependencies and all the supply chain attacks that will save us from

just for perspective if we switched to JSON this would only drop one dependency, that being the `ini` package itself. it has no external dependencies, so the impact on the overall dep tree is extremely minimal.

rondales:
> [Rendered RFC](https://github.com/npm/rfcs/blob/1d909afaf3b406e1261bffe7dcd1a0fdbaabbefa/accepted/0000-command-specific-config.md)

PR #591, opened 2022-05-20
RFC: Registry-scoped keyfile / certfile credential options

jenseng:
[Rendered RFC](https://github.com/npm/rfcs/blob/2e0516072ca9ea23fdc341335b68692b58a6338e/accepted/0000-registry-scoped-keyfile-and-certfile.md)

### DIFF
--- a/accepted/0000-registry-scoped-keyfile-and-certfile.md
+++ b/accepted/0000-registry-scoped-keyfile-and-certfile.md
@@ -1,0 +1,84 @@
+# Registry-scoped keyfile / certfile credential options
+
+## Summary
+
+NPM should support registry-scoped config options for specifying the paths to the client certificate and key used in registry fetches.
+
+These options should also be acceptable as authentication credentials by `npm publish` and friends.
+
+## Motivation
+
+Currently you can set a `cert` and `key` config options at the top level, which means they can be used in connections against any registry. Ideally a client certificate would only be presented to the registry that actually needs it, and you should be able to use different ones depending on the registry.
+
+Additionally, the `cert` and `key` options are the actual contents of the cert/key, rather than paths to them. This means you either need to put them in `.npmrc` — possibly keeping it in sync with the actual keys/certs — or you have to specify them as env vars. Both options potentially weaken security (e.g. `.npmrc` could be accidentally checked in or read by other users, `npm_config_key` could be extracted from `/proc/<pid>/environ`, etc.).
+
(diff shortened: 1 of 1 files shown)


accepted/0000-registry-scoped-keyfile-and-certfile.md
```diff
@@ -0,0 +1,84 @@
...
+  - You can't set a scoped registry password/auth/authToken via env vars due to [this bug](https://github.com/npm/config/issues/64).
+
+## Implementation
+
+1. Resolve scoped `certfile`/`keyfile` options and use them, i.e.
+   1. **`npm-registry-fetch`**: Update [`getAuth` + `Auth`](https://github.com/npm/npm-registry-fetch/blob/main/lib/auth.js) to resolve any scoped `certfile` and `keyfile` options into corresponding `cert` and `key` properties, and update [`regFetch`](https://github.com/npm/npm-registry-fetch/blob/main/lib/index.js#L26) to use them in the `fetch` call.
+2. Accept `certfile`/`keyfile` as valid creds, i.e.
+   1. **`@npmcli/config`**: Update [`getCredentialsByURI`](https://github.com/npm/config/blob/1244177d8a68f68f2be46d5b9c21957da7dedfbb/lib/index.js#L759) to process scoped `certfile`/`keyfile` options and then set them in the returned credentials.
+   2. **`npm`**: Update [this `noCreds` check](https://github.com/npm/cli/blob/e992b4a21ecdd96aa33c59682c0ac0cc8a30d776/lib/commands/publish.js#L108) to accept `certfile`/`keyfile` as creds.
```
jenseng, line 77: Digging a little deeper, looks like [cli:lib/utils/get-identity.js](https://github.com/npm/cli/blob/ee3308a7a08799ec7e86237165ebaf278d9a4f9f/lib/utils/get-identity.js) and friends may need to be updated too, as there are some assumptions around it being possible to get the current username, which would not necessarily be the case if authn is strictly done via mTLS 🤔 . 

If that ends up being a whole can of worms, maybe this RFC could be reduced in scope (pun!) so as to cover just registry-scoped certfile/keyfile, leaving out the only-authenticate-via-mTLS feature. Then (I think) the changes can be limited to just npm-registry-fetch.

jenseng: IMO though it is reasonable to require custom registries to implement a `/-/whoami` for compatibility, in which case a meaningful username could be returned by `get-identity` in these cases so that other commands like `unpublish` can work solely via mTLS

wraithgar: The "fail if you're not logged in" checks are a UX thing, and it's probably ok to err on the side of trying the attempt if we have a keyfile.

owlstronaut (approved):

Issue #581, opened 2022-05-03
[RRFC] add message (or opt-out support) for non registry tarball URLs


thescientist13:
## Motivation ("The Why")
This came up during one of the recent RFC meetings a couple weeks ago, and so just wanted to capture it for posterity.  I also think it would be a nice feature to have for the security conscious among us. 

### Example
Whenever a command is run, like `npm i` that otherwise adds / updates installed packages on disk, if any package is _NOT_ coming from a registry (e.g. [a tarball URL](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies)), then a message should be presented to the user.

For example a _package.json_ like this would trigger a message
```json
{
  "dependencies": {
    "@babel/cli": "^7.4.0",
    "eslint": "git+https://github.com/eslint/eslint.git"
  }
}
```

----

I speculate that it might only need to apply to transitive dependencies because presumably as an author, if you are setting it for your direct dependencies in your own top-level _package.json_, then that was a conscious choice.  Where as you have no control over what's at the end of that tarball URL if it's coming in through a transitive dependency.

#### Current Behaviour
There is no messaging about tarball URLs.

#### Desired Behaviour
An npm install going forward would message about tarball URLs.

```
 % npm i     
npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)
npm WARN eslint installed from tarball URL <URL>

added xxx packages, and audited xxx packages in 8s

...
```

Additionally, is there value in having the operation fail if the user wants to opt-out of tarball URLs entirely?  The rationale being that if the package is installed, but only a message is shown, and if that package is malicious, then it is already too late by that point to avoid any harm.  

Some users may want to explicitly opt-out of any anything that does NOT come from a registry, and so the command should fail immediately at that point.
```
$ npm install --no-tarball-urls
```

### References

* n/a

ljharb:
I would still want it on direct dependencies; i might set the flag in npmrc and a coworker might unknowingly install a direct nom-registry dep.

ljharb:
The behavior I’d want is to configure either silent, warn, or fail the install. The non-semver-major default should be “nothing”, and probably in v9 should change to “warn”.

thescientist13:
Some feedback from the call, prior to opening an RFC

- warn by default for direct dependencies
- hide by default for transitive dependencies
[...]

PR #593, opened 2022-05-25
first draft

thescientist13:
When installing dependencies, the npm CLI should have a mechanism for communicating (and optionally failing on) dependencies that do not come from a registry.

## References
related to #581

### DIFF
--- a/accepted/0000-registry-only-tarballs.md
+++ b/accepted/0000-registry-only-tarballs.md
@@ -1,0 +1,107 @@
+### References
+relates to #581
+
+----
+
+# Registry Only Tarballs
+
+## Summary
+
+When installing dependencies, the npm CLI should have a mechanism for communicating (and optionally failing on) dependencies that do not come from a registry, like a [git URL](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies).
+
+## Motivation
+
+In an effort to give users of npm more insight and transparency into the packages they are installing, knowing that a package is _not_ coming from a registry, and thus is susceptible to mutability and related supply chain attacks from that vector, seems like a meaningful heuristic missing from the package manager.
+
+Thus, npm should afford various levels of messaging and erroring through a flag to accommodate users of npm looking to gain this insight from their dependency graph.
+
+## Detailed Explanation
+
+For example a _package.json_ like this
+```json
+{
+  "dependencies": {
+    "@babel/cli": "^7.4.0",
+    "eslint": "git+https://github.com/eslint/eslint.git"
+  }
+}
+```
+
+Would trigger a message  to the console for **eslint** if the `warn` value was passed.
+
+```sh
+$ npm install --only-registry-tarballs=warn
+```
+
+## Rationale and Alternatives
+
+It could be argued that a user could write a script themselves that does all the look-ups and resolution from a given _package.json_ to vet their dependencies that way, potentially reusing arborist in a similar manner to how the CLI uses it.
+
+However, when this issue was raised no other alternatives were suggested at the time, even the one above, so it seems that most feel (implicitly at least) that the package manager is in the best position to vet the location of dependencies before they can be installed, and communicate that information to the user.
+
+## Implementation
+
+As npm is building up the dependency graph, it should check if _any_ dependency in the tree is not using a valid [semver range](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies) and emit a message to the user in the terminal.
+
+The flag would allow three values that the user could be provided based on the outcome of messaging / behavior the user was looking for:
+```sh
+$ npm install --only-registry-tarballs=silent|warn|fail
+```
+
+### Silent (default)
+Same behavior as it is now in the CLI, which is effectively that there is no consideration made for the location of a dependency.  This would be the default value of the flag as far as the npm CLI is concerned, or if the user used the flag but and did not pass any value.
+
+```sh
+# both would behave the same way
+$ npm install --only-registry-tarballs
+$ npm install --only-registry-tarballs=silent
+```
+
+### Warn
+Emits a `console.log` to the terminal indicating that a dependency is installing via a non-registry URL.
+
+```sh
+$ npm install --only-registry-tarballs=warn 
+npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)
+npm WARN eslint NOT installed from a registry, using tarball URL <URL>
+
+added xxx packages, and audited xxx packages in 8s
+
+...
+```
+
+### Fail
+Emits a `console.error` to the terminal indicating that a dependency is installing via a non-registry URL AND then exits the process with an exit code.
+
+```sh
+$ npm install --only-registry-tarballs=error
+npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)
+npm ERROR eslint NOT installed from a registry, using tarball URL <URL>
+
+process exited with code ....
+```
+
+----
+
+Additionally, this behavior should probably captured in relevant area(s) of the documentation, likely in the dependencies section of the docs, where non semver range versions are discussed.
+
+## Prior Art
+
+As far as I know, while other package managers allow installing from URLs and git repositories, none of them call it out as a potential anti-pattern or area of caution.
+- **yarn**: [`add`](https://classic.yarnpkg.com/en/docs/cli/add#toc-adding-dependencies), [`install`](https://yarnpkg.com/cli/install)
+- **pnpm**: [`add`](https://pnpm.io/cli/add#install-from-git-repository), [`install`](https://pnpm.io/cli/install)
+
+This seems like a worthwhile area for npm to be the first mover.
+
+## Unresolved Questions and Bikeshedding
+
+### Default Behavior
+As proposed, the default behavior would be the `silent` option.  As part of a semver-major change, the default behavior could be changed to that of the `warn` option instead.
+
+### Dependency Distinctions
+A point raised was if different dependency types should have different rules applied to them, or if this flag should apply equally across direct and transitive dependencies alike.  ex.
+- warn by default for direct dependencies
+- hide by default for transitive dependencies
+
+### Naming
+Currently the flag is `--only-registry-tarballs` which while explicit, is a bit verbose.  I think the final flag name is less consequential / material to the ultimate objective of this RFC, as long as it gets clearly captured in relevant areas of the documentation.


ljharb:
also, it's not about tarballs, it's about "from a registry" - users don't think about tarballs, and a tarball URL from a registry should be disallowed too.

thescientist13:
Feedback from the meeting for updating this RFC
1. Tie this in with `npm audit` instead of `install`
1. Warn by default
1. Reference `npm deprecate` for output messaging example / conventions
1. Rename to `--only-non-remote-deps`
1. Ensure proper communication / documentation with the rollout

Perhaps in the future as another RFC, we can leverage `npm query` to add more granularity / overrides on this command by being able to "filter" (include / exclude) on `dependencies`, `devDependencies`, etc

cc: @ljharb / @darcyclarke to keep me honest ^ 🙏

ljharb:
re 4, `npm audit dependency-types`, or something similar - and then have `install` auto-run that audit subcommand - and then command-specific config can be used to customize it using an `npm query` selector

dominykas:
Not sure if this needs to be explicitly state, but since I just recently had to do the work internally - the warning should also apply to `overrides` (quite possibly the same logic as transient vs direct deps - although an override could be either).

thescientist13:
Ok, think I covered the feedback here in a follow up revision
1. Linked to `npm audit` and touched upon interplay between `npm install`
1. Default option set to `warn`
1. Leverage `npm why` style output within the _Implementation_ section
1. Captured future examples of future eslint style audits / filters that could be applied (as new RFC) in the _Bikeshedding_ section

For the next meeting, would just like to take another pass at the naming of the flag (` --only-non-remote-deps` ) and overall RFC name to see if another coat of pain on the bike shed is desired.  I know the emphasis here is on registries specifically and not so much tarballs (which I agree with) but wanted to capture a couple of my own personal thoughts on the flag name:
- As I read it, it seems to assert on the negative which is kind of hard to reason about, like an `if(!someNegativeBool)`?  Feel like I have to keep reading it and reversing it to get the right "state" in my mind.
- It doesn't mention anything about registries, but this flag strongly correlates with that as the principal source of truth, and so wondering if discoverability (in docs, google, etc) would be hampered by too abstract of a name?
- To the point above, I believe one call out was that a locally linked specifier would be an exception.  Is that the only case?  Maybe that's just more of an implicit `localhost` style local development convenience that could be documented, rather than pull the name in a more abstract direction?  (e.g. centering the name on principal use case on registry specifiers first and foremost).


Happy to make any updates to the RFC title, filename, etc if it's ok to have another chat on the name and thanks again for all the help and feedback on this so far!  🙌

wesleytodd:
> pass at the naming of the flag (`--only-non-remote-deps`)

Thoughts on `--only-registry-deps`?  I think this more accurate to the intent right? Specifically this is intended to flag "mutable deps", and since we trust the registry to be immutable (mostly) it seems reasonable to me to name it this. And it seems folks above agree with this.

---

FWIW, this conversation is why I think we should consider following a well known path for this problem. We would need to have discussions like this for each of these new additions and they are gated by a central authority. The `eslint` model is federated by design and the most common & agreed upon rules bubble up to the core while letting innovation happen in plugins and opinionated rule-sets. The npm team and npm cli do not need to be the ultimate gate keeper, and shouldn't be for their own time management optimization.

thescientist13:
Had some time to play around with the above and got `npm query` working locally and created a [test bench repo](https://github.com/thescientist13/npm-query-tarball-test) to practice with.

For example, was able to single out the eslint dependency as such
```sh
% npm query ":root > *[resolved^="git"]"
[
  {
    "version": "8.20.0",
    "resolved": "git+ssh://git@github.com/eslint/eslint.git#0bcd2255c40b5c115a95181864776b0dd456c2dc",
    "license": "MIT",
    "dependencies": {
      "@eslint/eslintrc": "^1.3.0",
      "@humanwhocodes/config-array": "^0.9.2",
      "ajv": "^6.10.0",
      "chalk": "^4.0.0",
      "cross-spawn": "^7.0.2",
      "debug": "^4.3.2",
      "doctrine": "^3.0.0",
      "escape-string-regexp": "^4.0.0",
      "eslint-scope": "^7.1.1",
      "eslint-utils": "^3.0.0",
      "eslint-visitor-keys": "^3.3.0",
      "espree": "^9.3.2",
      "esquery": "^1.4.0",
      "esutils": "^2.0.2",
      "fast-deep-equal": "^3.1.3",
      "file-entry-cache": "^6.0.1",
      "functional-red-black-tree": "^1.0.1",
      "glob-parent": "^6.0.1",
      "globals": "^13.15.0",
      "ignore": "^5.2.0",
      "import-fresh": "^3.0.0",
      "imurmurhash": "^0.1.4",
      "is-glob": "^4.0.0",
      "js-yaml": "^4.1.0",
      "json-stable-stringify-without-jsonify": "^1.0.1",
      "levn": "^0.4.1",
      "lodash.merge": "^4.6.2",
      "minimatch": "^3.1.2",
      "natural-compare": "^1.4.0",
      "optionator": "^0.9.1",
      "regexpp": "^3.2.0",
      "strip-ansi": "^6.0.1",
      "strip-json-comments": "^3.1.0",
      "text-table": "^0.2.0",
      "v8-compile-cache": "^2.0.3"
    },
    "bin": {
      "eslint": "bin/eslint.js"
    },
    "engines": {
      "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
    },
    "funding": {
      "url": "https://opencollective.com/eslint"
    },
    "name": "eslint",
    "_id": "eslint@8.20.0",
    "pkgid": "eslint@8.20.0",
    "location": "node_modules/eslint",
    "path": "/Users/owenbuckley/Workspace/github/repos/npm-query-tarball-test/node_modules/eslint",
    "realpath": "/Users/owenbuckley/Workspace/github/repos/npm-query-tarball-test/node_modules/eslint",
    "from": [
      "",
      "node_modules/eslint-utils"
    ],
    "to": [
      "node_modules/@eslint/eslintrc",
      "node_modules/@humanwhocodes/config-array",
      "node_modules/ajv",
      "node_modules/eslint/node_modules/chalk",
      "node_modules/cross-spawn",
      "node_modules/debug",
      "node_modules/doctrine",
      "node_modules/eslint/node_modules/escape-string-regexp",
      "node_modules/eslint-scope",
      "node_modules/eslint-utils",
      "node_modules/eslint-visitor-keys",
      "node_modules/espree",
      "node_modules/esquery",
      "node_modules/esutils",
      "node_modules/fast-deep-equal",
      "node_modules/file-entry-cache",
      "node_modules/functional-red-black-tree",
      "node_modules/eslint/node_modules/glob-parent",
      "node_modules/eslint/node_modules/globals",
      "node_modules/ignore",
      "node_modules/import-fresh",
      "node_modules/imurmurhash",
      "node_modules/is-glob",
      "node_modules/js-yaml",
      "node_modules/json-stable-stringify-without-jsonify",
      "node_modules/levn",
      "node_modules/lodash.merge",
      "node_modules/minimatch",
      "node_modules/natural-compare",
      "node_modules/optionator",
      "node_modules/regexpp",
      "node_modules/strip-ansi",
      "node_modules/strip-json-comments",
      "node_modules/text-table",
      "node_modules/v8-compile-cache"
    ],
    "dev": false,
    "inBundle": false
  }
]
```

So will just keep tinkering around with the command and crafting the API to update this RFC with.  Will look forward to a few minutes to discuss progress and next steps on the next RFC call.

thescientist13:
From the call, got tipped off to [these `type` hints](https://github.com/npm/npm-package-arg#result-object) that I can use with query to get specific information about the type of package, e.g. `remote`, `version`, `git`, etc.

The takeaway was that the only types that should be initially blocked are `git` and `remote`.

_example:_
```sh
npm query ":root > *:not(:type(git,remote))"
```

thescientist13:
Ok, new round of feedback applied!  🙌 

1. Aligned this RFC with the landing of `npm query`
1. Updated _Implementation_ section to detail delegation of implementation to suggested query
    ```sh
    $ npm query ":root > *:is(:type(git,remote))"
    ```
1. RFC name change to **Only Registry Dependencies** (was _Only Registry Tarballs_)
1. Flag name change from `--only-non-remote-deps` to `--only-registry-deps`

wraithgar:
`:is(:type(git,remote))` would give you everything in your tree that is a git or remote reference.

thescientist13:
OK, query updated!  Think this might be good to go!  🙇

accepted/0000-registry-only-dependencies.md
```diff
@@ -0,0 +1,167 @@
+### References
+relates to #581
```
ruyadorno, line 2: I believe this line can be removed prior to ratifying / accepting the RFC
```suggestion
```

accepted/0000-registry-only-dependencies.md
```diff
@@ -0,0 +1,167 @@
...
+
+----
+
+# Registry Only Dependencies
+
+## Summary
+
+When auditing dependencies with `npm audit`, the npm CLI should have a mechanism for communicating (and optionally failing on) dependencies that _do not_ come from a registry, like a [git URL](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies).
+
+> _**Note**: this RFC has a hard dependency on [`npm query`](https://github.com/npm/cli/pull/5000) landing to support its implementation._
```
ruyadorno, line 12: Given that `npm query` has already landed I believe this note can be removed from the RFC text

```suggestion
```

accepted/0000-registry-only-dependencies.md
```diff
@@ -0,0 +1,167 @@
...
+
+As far as I know, while other package managers allow installing from URLs and git repositories, none of them call it out as a potential anti-pattern or area of caution.
+- **yarn**: [`add`](https://classic.yarnpkg.com/en/docs/cli/add#toc-adding-dependencies), [`install`](https://yarnpkg.com/cli/install)
+- **pnpm**: [`add`](https://pnpm.io/cli/add#install-from-git-repository), [`install`](https://pnpm.io/cli/install)
+
+This seems like a worthwhile area for npm to be the first mover.
+
+To demonstrate, if you see [this demo repo](https://github.com/thescientist13/npm-query-registry-only-deps-rfc-demo) and follow the steps to `npm link` with a version that has `npm query`, you will see output for **eslint** but not **babel**, which is the desired outcome in this situation given that _package.json_ has eslint as a `git` dependency.
+
+## Unresolved Questions and Bikeshedding
```
ruyadorno, line 143: This section is usually removed prior to ratifying / accepting, are all these questions answered?

ruyadorno (changes requested):

accepted/0000-registry-only-dependencies.md
thescientist13: will do!

thescientist13:
Ok, all notes and bikeshedding cleaned up!

ljharb:
Here's the overall plan of record as I understand it:

`npm audit` will run multiple "audit types". One of these, of course, is "vulnerabilities", which is the current thing it checks. This RFC is for "dependencies" (name to be bikeshedded). Another one is "licenses" (#182). Many more can be envisioned, including a repeatable type that lets users provide a custom `npm query` selector.

Once [command-specific config](https://github.com/npm/rfcs/pull/566) lands, each audit type will be configurable:
 1. does it run in `npm install`, or `npm audit`, or both, or neither?
 1. When it runs (in each of install or audit), does it fail the overarching action? or just warn?
 
 This will allow users to set whatever level of strictness they want, separately for both audit and install, while simultaneously allowing npm maximal freedom to change the defaults in semver-majors however they like - because users who dislike the defaults can just configure it in advance.

wraithgar:
Now that npm query can be told to exit uncleanly if any results come back with [`--no-expect-results`](https://docs.npmjs.com/cli/v10/using-npm/config#expect-result-count) I think npm has all the functionality needed for users to do this on their own.

ljharb:
An example in the docs of how to do just that, especially in npmrc, would be amazing - and it’d still be nice and ergonomic (and help message that npm is encouraging this restriction) to have a single cli flag that wraps all the pieces.

PR #595, opened 2022-05-30
Propose backwards-compatible improvements to compression

EvanHahn:
This proposes a method to shrink the size of npm packages by about 5% while maintaining backwards compatibility.

### DIFF
--- a/accepted/0000-backwards-compatible-tarball-compression.md
+++ b/accepted/0000-backwards-compatible-tarball-compression.md
@@ -1,0 +1,98 @@
+# Improving tarball compression while maintaining backwards compatibility
+
+## Summary
+
+At publish time, improve compression of tarballs. Do this in a backward-compatible way, using [Zopfli][zopfli].
+
+## Motivation
+
+The goal of this proposal is to improve compression of npm packages while limiting usability disruption. I believe this can be implemented in a way that is transparent to package maintainers, users, and registries.
+
+For developers, this could speed up installations, lower bandwidth usage, improve reliability on spotty connections, and leave more room on the hard drive. Registries and mirrors would also benefit; reduced resources likely means lower costs.
+
+I compressed the latest version of several popular packages using my proposed method. Here are the size savings:
+
+| Package | Original size   | Compressed size | Savings             |
+| ------- | --------------- | --------------- | ------------------- |
+| lodash  | 318,961 bytes   | 300,307 bytes   | 18,654 bytes (~6%)  |
+| express | 55,756 bytes    | 53,157 bytes    | 2599 bytes (~5%)    |
+| react   | 81,166 bytes    | 77,147 bytes    | 4019 bytes (~5%)    |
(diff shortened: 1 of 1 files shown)


EvanHahn:
I missed [the meeting where this was discussed](https://github.com/npm/rfcs/issues/596), but I took the following notes from the recording:

- A proof of concept would be useful.
- Concerns about how to run Zopfli came to the fore because the project is written in C, not JavaScript. There were three options proposed: port it to JavaScript, use it in WebAssembly, or request its addition to Node core.
- The RFC can be accepted without a full solution. In other words, it can be an intent to implement, not a full implementation.
- Overall, this proposal is worth evaluation.

I built a proof of concept that uses WebAssembly to repack an npm tarball and report any size savings. You can [run it in your browser](https://evanhahn.github.io/npm-repack-with-zopfli-proof-of-concept) and [see the source code](https://github.com/EvanHahn/npm-repack-with-zopfli-proof-of-concept).

I plan to attend the meeting on June 15. Please let me know if you have any questions before then!

See also:

- [Universal Zopfli](https://github.com/gfx/universal-zopfli-js), a WebAssembly-powered port of Zopfli, used in my demo
- [wasm-zopfli](https://github.com/dfrankland/wasm-zopfli), another WebAssembly implementation
- [zopfli.js](https://github.com/imaya/zopfli.js), a 9-year-old port of Zopfli that uses Emscripten

EvanHahn:
Finally got around to thinking more about this, and decide it's not worth adding to the npm CLI.

The pros:

- Newly-published packages would be smaller.

  It's hard to give exact figures here, but I tried recompressing the latest version of [the top 250 npm packages](https://socket.dev/npm/category/popular) with Zopfli. All packages got smaller.

  - Median bytes removed: 273.5
  - Median percentage removed: ~4.3%
  - Mean bytes removed: 4148.668 (~4 KiB)
  - Mean percentage removed: ~4.7%

  This was only after 1 Zopfli iteration. Increasing the number (for example, to 100) would further shrink packages.

- This would be backwards compatible.

  Zopfli creates valid gzip data. Unlike introducing new compression schemes, like Brotli, this proposal would be entirely backwards compatible.

The cons:

- It doesn't retroactively apply to old packages.

- It's more complex for the CLI.

  Zopfli is written in C, which presents challenges. Unless it was added to Node core, the CLI would need to (1) rewrite Zopfli in JS, possibly impacting performance (2) rely on a native module, impacting reliability (3) rely on a WebAssembly module. All of these options add complexity.

- Publishing would be slower. Sometimes, much slower.

  For example, I tried recompressing the latest version of the `typescript` package. GNU `tar` was able to completely compress the archive in about 1.2 seconds on my machine. Zopfli, with just 1 iteration, took 2.5 _minutes_.

I still think this is worth doing, but probably "in userland". More valuable, I think, would be finding a way to introduce better compression algorithms like Brotli.

I'm going to close this issue after doing this research. Thanks for your time!

PR #618, opened 2022-07-29
RFC: `npm debug` command

about-code:
This RFC proposes an `npm debug` command to simplify debugging npm packages.

Prior discussions at https://github.com/npm/feedback/discussions/585


accepted/0000-debug-command.md
```diff
@@ -0,0 +1,121 @@
...
+npm debug
+~~~
+
+1. GIVEN a `script.debug` property is present in `package.json`
+   - THEN run the run-script similar to how `npm start` would run `script.start`. END.
+   - ELSE continue
+1. GIVEN a `bin` property is present in `package.json`
+   - THEN launch a debug session for the script referred to by the `bin` property using `node --inspect-brk <bin-script> <script-argv>`. END.
+   - ELSE continue
+1. GIVEN a `main` property is present in `package.json`
```
ljharb, line 64: what about packages with both a bin and a main that want to debug both?

(i realize it's a bad practice now to have both in the same package, but there's lots of older packages that do)

ljharb:
I believe that setting `NODE_OPTIONS='--inspect-brk'` would make debugging Just Work with any node-based command you wanted to run. Does that work for you?

about-code:
> I believe that setting `NODE_OPTIONS='--inspect-brk'` would make debugging Just Work with any node-based command you wanted to run. Does that work for you?

From a purely technical point of view, yes, *it works* (similar to how `node --inspect-brk <some-path>` would do it, as well). However, this RFC is less about making debugging *just work* but should be perceived as being about *developer experience* and how to make debugging *particular packages* more *approachable*. 

I see the following use cases:

### 1: Authoring an npm package with a CLI

*Instead of crafting a command-line or `run-script`* with

~~~
NODE_OPTIONS='--inspect-brk' node ./bin/foo.js
~~~

this RFC proposes to provide something by npm out of the box which allows typing

~~~
npm debug
~~~

### 2: Monorepos where one or more workspace packages are packages with a CLI 

...of course, people could craft a command line and make a `run-script` for it. But from a usability perspective I consider it uncomfortable for users for two reasons:
   - they'd need a separate `run-script` for every such package in the workspace
   - they'd need to set up such `run-scripts` for every new project they start, even though *debugging* is a recurrent development task quite similar to *init*, *start*, *test* or *publish* a package.

Instead of crafting a command-line or `run-script`
~~~
NODE_OPTIONS='--inspect-brk' node ./packages/foo-package/bin/foo.js
~~~

this RFC proposes to provide something by npm out of the box which allows typing

~~~
npm debug foo
~~~

### 3: Debugging tools and packages in a `node_modules` folder

Consider a project with a package like [eslint](https://npmjs.com/packages/eslint) installed within its `node_modules` folder. So the package has a CLI. Except, the tool not being eslint but a less prominent one `foo` with less maintainers and less quality. 

- *As a user of such package* you've tried everything to make it work, but it doesn't. You want to confirm by debugging the tool that the bug is not in the way you use it but in the tool itself. 
- *As a maintainer* of such package you request a sample repository for debugging your tool's usage within that sample repository.

Instead of crafting a command-line or `run-script`
~~~
NODE_OPTIONS='--inspect-brk' node ./node_modules/.bin/foo.js
~~~

this RFC proposes to provide something by npm out of the box which allows typing 

~~~
npm debug foo
~~~

accepted/0000-debug-command.md
about-code: I see the following options:

1. Keep implict priority and order and execute `bin`. Require a `--main` switch to allow for debugging `main`.
2. Keep implicit priority but change order and execute `main`. Require a `--bin` switch to allow for debugging `bin`.
3. Consider it to be a conflict and exit with an NPM ERR. Require a `--main` or `--bin` argument in `npm debug`s `argv`

Consequences
1. would require us to have a `--main` and `--bin` argument, since a `--bin` argument seems to be necessary for the case of *multiple `bin` executables*, anyways.
2. would get us around a `--main` argument and just require a `--bin` CLI argument. The latter seems to be necessary for the case of *multiple executables* (update on this, follows) 
3. would ask for an explicit decision from the user.

about-code: Update https://github.com/npm/rfcs/pull/618/commits/70c0a015f5ad8589c34fd609765b022ae620a995 decides for 1 and would allow users to be explicit about whether to debug `main` or `bin`.

~~~
npm debug --main
npm debug --bin
~~~

Given they provided both `--bin` *and* `--main` or none at all then `--bin` would take precedence:

~~~
npm debug
npm debug --main --bin
~~~

... executes a packages `bin` script if a `package.json` has a `bin` and a `main` script.

about-code:
Proposal has not got much support. Closing this for the sake of clean up. Reopen if you do not agree.

PR #636, opened 2022-09-20
RFC: Add `--query` to `npm audit`

darcyclarke:
[Rendered RFC](https://github.com/npm/rfcs/blob/bf58d9b1be42eb478ffc33b3cd4a079448a8cc51/accepted/0000-npm-audit-queries.md)

### DIFF
--- a/accepted/0000-npm-audit-queries.md
+++ b/accepted/0000-npm-audit-queries.md
@@ -1,4 +1,4 @@
-# Add `--query` flag to `npm audit`
+# Add `--audit-query` flag to `npm audit`
 
 ### Motivation
 
@@ -8,11 +8,11 @@ The current configuration options are based on `Arborist`'s underlying  support 
 
 ### Solution
 
-Add support for a new `--query` flag to `npm audit` which takes a **Dependency Selector** as it's value. 
+Add support for a new `--audit-query` flag to `npm audit` which takes a **Dependency Selector** as it's value. 
 
 ### Implementation
 
-When the `--query` flag is defined it will set `--omit` & `--include` values to empty & configure `Arborist` to run the `tree.querySelectorAll("<query-selector-string-value>")`.
+When the `--audit-query` flag is defined it will set `--omit` & `--include` values to empty & configure `Arborist` to run the `tree.querySelectorAll("<query-selector-string-value>")`.
 
 ### Prior Art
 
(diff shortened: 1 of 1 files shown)


accepted/0000-npm-audit-queries.md
```diff
@@ -0,0 +1,19 @@
+# Add `--audit-query` flag to `npm audit`
+
+### Motivation
+
+Today `npm audit` has a limited set of options to filter the packages that are included or excluded from it's scope. There has also been [very public, legitimate critcism about the signal to noise ratio](https://overreacted.io/npm-audit-broken-by-design/) of `npm audit`'s ouput. Unfortunately, this has lead many developers to turn off `npm audit` checks completely during installation (ex. `--no-audit`) or switching to alternative package managers who do not run audits by default. 
+
+The current configuration options are based on `Arborist`'s underlying  support for, & limited to, package types (ie. `--omit` & `--include` can be used today to filter by only `prod`, `dev`, `optional` & `peer` dependency types today). Expanding `npm audit`'s filter capabilities & leveraging the new, rebust [Dependency Selector Syntax](https://docs.npmjs.com/cli/v8/using-npm/dependency-selectors) - at the command-line/project-level - will help end-users define complex groups of dependencies to be included.
```
ljharb, line 7:
```suggestion
The current configuration options are based on `Arborist`'s underlying  support for, & limited to, package types (ie. `--omit` & `--include` can be used today to filter by only `prod`, `dev`, `optional`, & `peer` dependency types today). Expanding `npm audit`'s filter capabilities & leveraging the new, robust [Dependency Selector Syntax](https://docs.npmjs.com/cli/v8/using-npm/dependency-selectors) - at the command-line/project-level - will help end-users define complex groups of dependencies to be included.
```

ljharb (commented):
So this is just a filter on whatever `npm audit` is doing? If so, sounds great - it would presumably be settable in `.npmrc`, so the default audit done on install could be changed?

ruyadorno:
@darcyclarke the PR title mentions the name of `--audit-query` (with a prefix) for the new config option while in the RFC itself it's noted multiple times as `--query` (with no cmd name prefix). I believe one of them was a mistake 😅 can you please confirm what is the intended config name?

darcyclarke:
@ruyadorno updated the title. I'm easy with the flag name tbh but am leaning toward the un-prefixed `--query` so that it can be easily mapped to other commands easily (ex. `install` / `update` etc.). Prefixing of flags ideally won't be as much of an issue when/if we ever land Command-specific Config

darcyclarke:
@ljharb essentially, yes. In previous RFC discussions I've used the term "Audit Filters" a lot but it's really just ensuring we can passing around a query selector to define the set of dependencies we want to actually execute the action against.

ljharb:
As a generic means to filter whatever audit action happens to be being run, this makes perfect sense.

PR #637, opened 2022-09-20
RFC: Add Audit Policies Support

darcyclarke:
[Rendered RFC](https://github.com/npm/rfcs/blob/13d046eaa171f10bb7e8dce4ae0f656adcd33894/accepted/0000-npm-audit-policies.md)

### DIFF
--- a/accepted/0000-npm-audit-policies.md
+++ b/accepted/0000-npm-audit-policies.md
@@ -1,0 +1,89 @@
+# Audit Policies
+
+### Motivation
+
+Today there are a limited set of conditions in place that prevent the installation of a package (ex. integrity mismatches & engines conflicts); audits also happen post-installation meaning they are only advisory in practice.
+
+### Solution
+
+Introduce easily configurable audit definitions that can gate the installation of packages. This new feature should leverage existing functionality/commands (ex. `install`, `update` & `audit`), syntax (ex. Dependency Selectors) & metadata without expanding the scope to unbounded, arbitrary code execution (unlike `preinstall` scripts or lifecycle hooks).
+
+### Known Caveats
+- Adding extra validation during installation will slow down execution
+  - this will be up to end-users to control & determine what validations are necessary to meet their own requirements
+- Not all usecases will be met
+  - we will be limited by the existing commands, syntax & metadata supported
+  - we aim to meet 80% (or the majority) of usecases with this feature
+  - end-users with broader security needs can & still should look at locking down developer environments & enforce policies at the system/network level (something that is outside the scope of the `npm` CLI today)
+
+### Implementation
+
+```json
+{
+    "audit": {
+        "policies": [
+            {
+                "name": "Vulnerable",
+                "type": "error",
+                "query": ":vulnerable"
+            },
+            {
+                "name": "Peer Conflicts",
+                "type": "error",
+                "query": ".peer:not(:deduped)"
+            },
+            {
+                "name": "Deprecated",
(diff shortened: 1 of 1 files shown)


accepted/0000-npm-audit-policies.md
```diff
@@ -0,0 +1,89 @@
...
+### Implementation
+
+```json
+{
+    "audit": {
+        "policies": [
+            {
+                "name": "Vulnerable",
+                "type": "error",
+                "query": ":vulnerable"
```
ljharb, line 28: the query specifies which packages are selected, but what defines how the results are presented?

Vulnerability output is perhaps a special case, but is the output just "error, nonzero packages showed up in the query!"?

darcyclarke: Great question - I believe the idea would be to use something like the above but I can put some potential examples of output in the RFC itself

ljharb: I think that it won't be nearly as useful if all i can do is control the exit code with a query that produces nonzero results. I suspect people will want to provide their own *package* that can be a function that receives the query results as an argument, and can do whatever it likes, as long as it returns an exit code, or something.

naugtur:
Hello!

Where does the JSON go? If it's in a policies file, it feels like I could use it for npm-audit-resolver to specify what to ignore too. 
Can we consider an ignore condition here? Or setting the policy to error on all vulnerable and then specify exceptions with queries for each?

Do at me ;)

darcyclarke:
@naugtur the original idea was to have these policies live in `package.json` & get picked up by `npm audit` (they would then also apply during `install` or `update` if `--audit=true`); I can/should expand the RFC to note this & other specifics left out today (including what inclusion/exclusion might look like - although I think it would be fairly straightforward ie. DSS combinators ex. `:vulnerable:not(.dev)...`, but maybe you're hoping for more flexibility in composition of _which_ policies to run)

ljharb:
I think putting them in `package.json`, thereby inflating the size of the packument and published packages, is probably not the best idea - a separate JSON file seems ideal.

ljharb:
@bnb "too many files" is a subjective opinion - separate files allow you to have granular CODEOWNERS, as well, beyond being data that doesn't belong in a published package. The burden exists regardless because you still have to google for how to configure the package.json field.

bnb:
noting a combination of suggestions from the office hours: if there were a top-level package.json property like `npmConfig`, having the ability to have audit policies _and_ the ability to point to a file (or a module!) rather than having a pre-determined filename, both options could be supported. If policies as a devDep is an option, there's also _so much_ flexibility for both large-scale open source projects and companies alike.

ljharb:
i'd be totally fine with a field that optionally points to a file so that the config is discoverable AND the contents don't have to live in package.json :-)

rondales (approved):

rondales:
> [Rendered RFC](https://github.com/npm/rfcs/blob/13d046eaa171f10bb7e8dce4ae0f656adcd33894/accepted/0000-npm-audit-policies.md)

rondales:
> > [Rendered RFC](https://github.com/npm/rfcs/blob/13d046eaa171f10bb7e8dce4ae0f656adcd33894/accepted/0000-npm-audit-policies.md)

rondales (commented):
accepted/0000-npm-audit-policies.md

rondales (commented):
accepted/0000-npm-audit-policies.md

PR #640, opened 2022-10-04
RFC: dist-tag resolution correction

gemwuu:


gemwuu:
@wraithgar Hope this RFC could be discussed.

PR #648, opened 2022-10-26
Pull request: [improvement] reenable npm cli api access

ganeshkbhat:
## reenable npm cli api access

reenable npm cli api access - https://github.com/npm/libnpm to work with npm, npm commands, and package.json / npm scripts


wraithgar:
The functionality for these was moved into the workspaces in npm that are all published as separate packages on the npm registry.

The biggest barrier right now for folks using them is config parsing.  We do have plans on making `@npmcli/config` able to do everything for you which will improve things.

PR #653, opened 2022-11-11
add safety checks and UX considerations when running as root

isaacs:
[rendered RFC](https://github.com/isaacs/rfcs/blob/isaacs/root-restrictions/accepted/0050-root-restrictions.md)


ljharb (approved):
These all sound great to me; the only time someone should be using sudo with npm is `sudo npm install -g` when they've installed node in a global place.

nlf:
this is great, thanks for writing this up!

PR #626, opened 2022-08-08
RFC for linking packages to their source and build

feelepxyz:
[Rendered version](https://github.com/npm/rfcs/blob/link-packages-to-source-and-build/accepted/0000-link-packages-to-source-and-build.md)

PR #666, opened 2023-01-04
RFC for linking packages to their source and build without using Sigstore Fulcio

ycchang:
[Rendered RFC](https://github.com/ycchang/rfcs/blob/main/accepted/0000-remove-dependency-on-fulcio.md)

### DIFF
--- a/accepted/0000-remove-dependency-on-fulcio.md
+++ b/accepted/0000-remove-dependency-on-fulcio.md
@@ -1,0 +1,69 @@
+# RFC for linking packages to their source and build without using Sigstore Fulcio
+
+## Summary
+
+As it is showed in a recent paper that Sigstore Fulcio is an unnecessary trusted third party for linking npm packages to their source and build, we suggest the npm team consider the new approach presented in that paper which can be found at:
+
+https://eprint.iacr.org/2023/003
+
+The new approach not only eliminates the unnecessary dependency on an external black-box CA but also accelerates the signing process.
+
+## Motivation
+
+Last year an RFC for linking npm packages to their source and build using Sigstore Fulcio was proposed and accepted:
+
+https://github.com/npm/rfcs/pull/626
+
+However, the paper mentioned above shows the same result can be achieved without using Sigstore Fulcio at all. Since relying on an unnecessary black-box CA poses a serious threat to the npm community, we draft this RFC.
+
+## Detailed Explanation
+
+The key observation made by the paper mentioned above is the following:
+
+- Instead of setting the Audience claim of a requested OIDC token to be `sigstore` (as proposed by the previous RFC), a CI/CD workflow can in fact set the Audience claim to be something like `sigstore?pk=PubKey`, where `PubKey` is (the Base64 encoding of) the public key used by the CI/CD workflow for signing the software artifact.
+
+In other words, such an OIDC token can be understood as a witness signed by the CI/CD service, with the Audience claim encoding the workflow's intention to use the public key to represent it. Thus, instead of exchanging an OIDC token (and a public key) for a Fulcio certificate, we can directly use the OIDC token in place of the certificate (which is, after all, transcribed from the OIDC token).
+
+As for verification, instead of hardcoding the root certificate of Fulcio, we can just hardcode the public key of the corresponding CI/CD service (aka OIDC provider), which is part of the root of trust of Sigstore Fulcio (so that no new trust is introduced).
+
+## Rationale and Alternatives
+
+- The previous RFC is an alternative that depends on an external black-box CA
+- This RFC requires no external black-box CA
+- The paper mentioned above suggests Sigstore Rekor, also proposed in the previous RFC, can be directly replaced by standard RFC 3161 Time Stamping Authority (TSA) servers hosted by established CAs such as DigiCert and Sectigo for business grade SLAs
+
+## Implementation
+
+### Signing
+
+The signing procedure (within a CI/CD workflow) proposed by the previous RFC is roughly the following:
+
+1) Generate a disposable key pair
+2) Obtain a customizable OIDC token with Audience claim set to be "sigstore"
+3) Request a public-key certificate from Fulcio using the key pair and the token
+4) Digitally sign (the hash of) the software artifact using the private key
+5) Timestamp the signature using Rekor
+6) Output the certificate, the digital signature from step 4, and the timestamp data from step 5
+
+Below we adapt the above procedure by using strikethrough lines to denote deletions and using underlines to denote additions:
+
+1) Generate a disposable key pair 
+2) Obtain a customizable OIDC token Audience claim set to be "sigstore<ins>?pk=PubKey</ins>"
+3) ~~Request a public-key certificate from Fulcio using the key pair and the token~~
+4) Digitally sign (the hash of) the software artifact using the private key 
+5) Timestamp the signature using Rekor 
+6) Output the ~~certificate~~ <ins>OIDC token</ins>, the digital signature from step 4, and the timestamp data from step 5
+
+### Verification
+
+On the other hand, the offline verification proofs required by the previous RFC are the following:
+
+- The output from step 6
+- The root certificate of Fulcio
+- The public key that Rekor used to sign the timestamp data
+
+And we just need to adapt the second item:
+
+- The output from step 6 
+- ~~The root certificate of Fulcio~~ The public key that the OIDC identity provider used to sign the OIDC token
+- The public key that Rekor used to sign the timestamp data


accepted/0000-remove-dependency-on-fulcio.md
```diff
@@ -0,0 +1,69 @@
...
+
+As it is showed in a recent paper that Sigstore Fulcio is an unnecessary trusted third party for linking npm packages to their source and build, we suggest the npm team consider the new approach presented in that paper which can be found at:
+
+https://eprint.iacr.org/2023/003
+
+The new approach not only eliminates the unnecessary dependency on an external black-box CA but also accelerates the signing process.
+
+## Motivation
+
+Last year an RFC for linking npm packages to their source and build using Sigstore Fulcio was proposed and accepted:
```
ljharb, line 13: to be clear, that RFC was merged, but never was accepted in an RFC call per the npm team's process.

ljharb (commented):
is this just a modification of RFC #626? iow, it doesn't change the publishing workflow for authors or the consuming workflow for users?

accepted/0000-remove-dependency-on-fulcio.md
ycchang: Yes, it is just a modification of the mentioned RFC and does not change the workflows except for removing the unnecessary dependency on Sigstore Fulcio.

Sorry I did not realize the mentioned RFC, though merged to the accepted folder, was never accepted. Some related code found in the cli/provenance branch also made me feel so. Thank you for letting me know.

saquibkhan: Hi @ljharb @ycchang I am new Engg Mgr for npm cli and my team also works on part of nm registry,

I want to reconfirm that the RFC was accepted and merged after thourough review and sicussions with npm team.

We do agree and respect open source community feedbacks and there can be disagreements, sometime we would like to move fast and course correct progressively based on all inputs (ship to learn), 

we would thouroughly reiview and discuss this RFC, consider the inputs and feedback and if needed would course correct on previous rfcs

codysoyland:
It's an interesting and clever solution to embed a public key into the audience claim of the ID token, but I think this proposal ignores how vital Fulcio is for establishing and maintaining a root of trust. Fulcio uses [TUF](https://theupdateframework.io/) to distribute its root and intermediate certificates, which clients can use to verify signing certificates across all IDPs. Without Fulcio, there is not a consistent way to verify the ID tokens over the long term (as IDPs rotate their keys). This proposal shifts the responsibility of key management to client code that must directly integrate with each IDP (none of which use TUF AFAIK). This would present a big challenge to the NPM CLI.

accepted/0000-remove-dependency-on-fulcio.md
ljharb: @saquibkhan that RFC was the first time that an RFC was accepted outside of an RFC call, and this one in particular was *not* thoroughly discussed, because there were no RFC calls (due to the npm team being heads down on v9). It sends a very negative message that erodes the trust built up with the community over years to accept an RFC in this way.

accepted/0000-remove-dependency-on-fulcio.md
```diff
@@ -0,0 +1,69 @@
...
+
+In other words, such an OIDC token can be understood as a witness signed by the CI/CD service, with the Audience claim encoding the workflow's intention to use the public key to represent it. Thus, instead of exchanging an OIDC token (and a public key) for a Fulcio certificate, we can directly use the OIDC token in place of the certificate (which is, after all, transcribed from the OIDC token).
+
+As for verification, instead of hardcoding the root certificate of Fulcio, we can just hardcode the public key of the corresponding CI/CD service (aka OIDC provider), which is part of the root of trust of Sigstore Fulcio (so that no new trust is introduced).
+
+## Rationale and Alternatives
+
+- The previous RFC is an alternative that depends on an external black-box CA
+- This RFC requires no external black-box CA
+- The paper mentioned above suggests Sigstore Rekor, also proposed in the previous RFC, can be directly replaced by standard RFC 3161 Time Stamping Authority (TSA) servers hosted by established CAs such as DigiCert and Sectigo for business grade SLAs
```
haydentherapper, line 33: Rekor, a transparency log, is not equivalent to a timestamping authority. Rekor currently provides two things for each entry:
* The time the entry was logged in Rekor
* A way to cryptographically prove inclusion

The benefit of Rekor is that all entries are logged publicly in an append-only log. The second point here lets you prove that an entry is in the log, which means that the artifact and signature is publicly auditable. As a verifier, you shouldn't trust an entry that isn't publicly logged, which forces malicious actors into the public and can therefore be detected.

The first point is just timestamping, which isn't new to code signing. See Authenticode, Microsoft's approach to code signing that's been around for decades. You could use a timestamping authority to provide a timestamp instead of Rekor - We actually just added support for this in Cosign, Sigstore's CLI.

Using a TSA in addition to Rekor gives you both a third-party to assert time, and auditability. Using a TSA without Rekor is just timestamping, and is no different than previous approaches to code signing that lack transparency.

accepted/0000-remove-dependency-on-fulcio.md
```diff
@@ -0,0 +1,69 @@
...
+
+https://github.com/npm/rfcs/pull/626
+
+However, the paper mentioned above shows the same result can be achieved without using Sigstore Fulcio at all. Since relying on an unnecessary black-box CA poses a serious threat to the npm community, we draft this RFC.
+
+## Detailed Explanation
+
+The key observation made by the paper mentioned above is the following:
+
+- Instead of setting the Audience claim of a requested OIDC token to be `sigstore` (as proposed by the previous RFC), a CI/CD workflow can in fact set the Audience claim to be something like `sigstore?pk=PubKey`, where `PubKey` is (the Base64 encoding of) the public key used by the CI/CD workflow for signing the software artifact.
```
haydentherapper, line 23: OIDC providers are pretty strict on what can go into an audience. I think this is an interesting approach, but in practice, I don't see dynamic audiences being supported.

accepted/0000-remove-dependency-on-fulcio.md
```diff
@@ -0,0 +1,69 @@
+# RFC for linking packages to their source and build without using Sigstore Fulcio
+
+## Summary
+
+As it is showed in a recent paper that Sigstore Fulcio is an unnecessary trusted third party for linking npm packages to their source and build, we suggest the npm team consider the new approach presented in that paper which can be found at:
+
+https://eprint.iacr.org/2023/003
+
+The new approach not only eliminates the unnecessary dependency on an external black-box CA but also accelerates the signing process.
```
haydentherapper, line 9: What is meant by "external black-box CA"?

Fulcio uses certificate transparency, like other webPKI CAs, to allow users to monitor certificate issuances. Verifiers should only trust certificates that have been publicly logged, much like verifiers should only trust artifacts logged in Rekor. This makes misconfigured or malicious CA actions publicly auditable.

accepted/0000-remove-dependency-on-fulcio.md
```diff
@@ -0,0 +1,69 @@
...
+
+However, the paper mentioned above shows the same result can be achieved without using Sigstore Fulcio at all. Since relying on an unnecessary black-box CA poses a serious threat to the npm community, we draft this RFC.
+
+## Detailed Explanation
+
+The key observation made by the paper mentioned above is the following:
+
+- Instead of setting the Audience claim of a requested OIDC token to be `sigstore` (as proposed by the previous RFC), a CI/CD workflow can in fact set the Audience claim to be something like `sigstore?pk=PubKey`, where `PubKey` is (the Base64 encoding of) the public key used by the CI/CD workflow for signing the software artifact.
+
+In other words, such an OIDC token can be understood as a witness signed by the CI/CD service, with the Audience claim encoding the workflow's intention to use the public key to represent it. Thus, instead of exchanging an OIDC token (and a public key) for a Fulcio certificate, we can directly use the OIDC token in place of the certificate (which is, after all, transcribed from the OIDC token).
```
woodruffw, line 25: To be clear: this entails publishing the OIDC token as proof, right?

I understand the witness proof provided by doing so, but it's worth noting that this is potentially dangerous to do: OIDC tokens are just JWTs under the hood, and audience confusion/logic errors are a well-known class of bugs with JWTs. Even if the token is only short-lived, this could potentially result in lateral movement/unintended access to systems that accept OIDC tokens but fail to adequately filter based on audience.

woodruffw: To use an example:

1. Alice uses GitHub Actions to interact with both Amazon Web Services and Sigstore. Both use OIDC to perform authentication; Amazon Web Services uses the `sub` claim and Sigstore uses the `aud` claim.
2. Alice performs a Sigstore signing operation, and publishes her OIDC token per the amended scheme here.
3. Mallory takes the OIDC token made public by Alice and uses it to impersonate Alice on AWS, gaining access to whatever resources the token's `sub` claims allow.

woodruffw: (TL;DR this requires every OIDC RP to handle the `aud` claim, when many will just check `sub` or some other claim(s) relevant to their purposes.)

ycchang: Mallory cannot impersonate Alice on AWS if Alice configures AWS correctly by requiring a check of the audience claim, as showed in the official GitHub Actions document [here](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services). But I understand your point is not every Alice is knowledgeable enough to do this right. Thank you for your insight.

However, the Alice in the scope of our discussion is a repository owner/collaborator, and I tend to believe she should be knowledgeable enough. After all, if she cannot do this right, there are more things she could do wrong with her resources. And the bottom line is the security of her resources should not be based on hiding the OIDC tokens.

ycchang:
@codysoyland If the mentioned client integration is indeed difficult, to reduce the trust base and to improve the audibility, we just need a trusted third party to certify the IDPs’ current and past keys, instead of a trusted third party to certify all the ephemeral keys which significantly outnumber IDPs’ keys.

accepted/0000-remove-dependency-on-fulcio.md
ycchang: From the perspective of a theoretical cryptographer, a CA is a black box with or without certificate transparency. Sorry if the term means different to you.

accepted/0000-remove-dependency-on-fulcio.md
ycchang: Unfortunately, to verify if a signature is on Rekor, you need to either check it online (which renders a signature less meaningful as one can just check the hash listed in a GitHub release note online) or trust a Rekor-signed proof offline (which raises many questions such as why a proof that never expires should be trusted, why the proof-signing key can never be compromised, what if no one will do online check when everyone is used to the offline experience, and et al). However, since this RFC is about Fulcio, I am not against using Rekor for the time being.

accepted/0000-remove-dependency-on-fulcio.md
haydentherapper: Any server component that cannot be publicly audited will be a black box. How don’t you know that the identity provider isn’t issuing tokens with your identity to malicious actors? Without a transparency mechanism, this behavior can’t be audited.

accepted/0000-remove-dependency-on-fulcio.md
haydentherapper: > …you need to either check it online (which renders a signature less meaningful as one can just check the hash listed in a GitHub release note online)

A GitHub release note isn’t immutable. An attacker could compromise a repository and change the artifact hash to point to one they control

> or trust a Rekor-signed proof offline (which raises many questions such as why a proof that never expires should be trusted, 

Why should a proof expire? By asserting that a code signing certificate was valid at a time when either the entry was added to a log or a signed time stamp was generated, a verifier can know that a signer was in possession of the artifact signing key. A signature doesn’t need to expire, and if it did, it would be a significant burden on maintainers and verifiers to have to re-sign or fetch updated signatures. 

> why the proof-signing key can never be compromised

It could be compromised, and this is the risk of offline proofs not being easy to verify. Note that in certificate transparency, a verifier should verify an offline proof in an online manner at some point, but this mechanism hasn’t been implemented at scale.
Using TUF, we can respond quickly to compromises by shipping new key material with the TUF root. 

> what if no one will do online check when everyone is used to the offline experience, and et al)

that’s up to a verifier. Online verification will always provide stronger assurances.

dlorenc:
Zack tried to write up a lot of this in a single blog post here, the comments are hard to follow: https://blog.sigstore.dev/why-you-cant-use-sigstore-without-sigstore-de1ed745f6fc

nlf:
First off, thank you @ycchang for writing this RFC. We really appreciate your enthusiasm and the efforts you took to share your thoughts and how we might change this feature in the future. There's been a lot of really great discussions coming out of this, and that's amazing!

The blog post linked above ([here again for posterity](https://blog.sigstore.dev/why-you-cant-use-sigstore-without-sigstore-de1ed745f6fc)) explains in great detail some of the technical reasons that the approach you're proposing here is not something we are planning to adopt at this time. In the future as new products and technologies emerge we are always happy to revisit these decisions, though!

**Thank you** again for all of the time and effort you put into this, npm :heart:s you

nlf:
Actually, I realized that closing this early was a mistake as I didn't provide an opportunity for a real time discussion around this. That's my fault.

I'm reopening this for now so that we have a chance to discuss this in our open office hours calls. They are scheduled for every Wednesday at 11am pacific time. @ycchang if you're able to attend, we'd love to hear from you. @znewman01 I believe said he should be available for discussion.

ycchang:
Hi, sorry for the late reply as I am in a different time zone. 

I would like to thank you all for your time spent on this proposal. And I would like to especially thank [@znewman01](https://github.com/znewman01) for making the blog post and [@dlorenc](https://github.com/dlorenc) for liking the IACR tweet:P

I have no further feedback on this proposal at this moment and feel totally fine if you just close this RFC. Have a nice weekend:)

saquibkhan:
> I have no further feedback on this proposal at this moment and feel totally fine if you just close this RFC.

thanks eveyone, based on the discussions, blog post and author's final comments closing this RFC

PR #687, opened 2023-04-05
RFC: npm publish, unpublish, and [republish] functionality change

ganeshkbhat:
npm publish, unpublish, and [republish] functionality and data/repository management and log management policy (security policy) changes.

This RFC is a proposal where I recommend allowing republishing the same npm package version v1.0.0 with a different codebase B after unpublishing a version v1.0.0 with codebase A; with a possibility to view the publish, unpublish, republish logs/ codebase, etc. This recommended change improves the npm package publish-unpublish process, (historical) publish-unpublish data management policy, and (historical) publish-unpublish log management policy (and security management policy).

## References
Detailed in RFC

### DIFF
--- a/accepted/0000-template-republish-npm-package-after-unpublish.md
+++ b/accepted/0000-template-republish-npm-package-after-unpublish.md
@@ -1,0 +1,89 @@
+# RFC: npm publish, unpublish, and deprecate functionality and data/repository management and log management policy (security policy) changes
+
+## Summary
+
+When I publish a package v1.0.0 and unpublish it, it is unpublished correctly. However, I am not able to re-publish any other codebase B/ C/ D into v1.0.0. I will not be able to re-publish the same version v1.0.0 with a new or same codebase.
+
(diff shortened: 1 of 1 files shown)


ljharb:
A version number, once used, can never be used for anything else - otherwise it would be a massive security hole. I already answered this on your cli issue, and this makes the third place you’ve posted about this.

accepted/0000-template-republish-npm-package-after-unpublish.md
```diff
@@ -0,0 +1,89 @@
+# RFC: npm publish, unpublish, and deprecate functionality and data/repository management and log management policy (security policy) changes
+
+## Summary
+
+When I publish a package v1.0.0 and unpublish it, it is unpublished correctly. However, I am not able to re-publish any other codebase B/ C/ D into v1.0.0. I will not be able to re-publish the same version v1.0.0 with a new or same codebase.
```
bnb, line 5: This is good and intended. While it may suck as a publisher (I recently had to help internally with an ancient mispublish that bumped a semver minor, meaning we had to go straight to 4.8.0 rather than starting at 4.0.0!), the existing approach is very rarely an issue, is a simpler solution, and is less potentially harmful than what's proposed here IMO.

ljharb:
You can already do this with `npm deprecate`. Having it be programmatic and implicit is also a security issue.

PR #696, opened 2023-06-04
[RFC] Transactional publish in `npm-publish workspace` command

vighnesh153:
When running `npm publish --workspace packages --transactional` in a monorepo, it should run in a transaction so that if publishing of any (or multiple) package fails, it automatically rollback the other published packages.

## References
None

### DIFF
--- a/accepted/0000-transactional-publish.md
+++ b/accepted/0000-transactional-publish.md
@@ -1,0 +1,48 @@
+# Transactional Publish
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


ljharb:
This seems like a good enhancement on top of #92, which remains very critically needed.

wraithgar:
This has been brought up many times in the past and is essentially a registry feature not a cli feature.  There is currently no way to "roll back" an already published package.  That version is out and can not be used again.

ruyadorno:
well, I'd argue that the cli component proposed here is blocked by registry improvements such as #92 pointed by @ljharb above, that could potentially unlock a transactional-like publishing experience for workspaces once all packages are in a staging area

sin-strokes (approved):

PR #714, opened 2023-08-07
RFC: SBOM generation

bdehamer:
[Rendered RFC](https://github.com/bdehamer/rfcs/blob/bdehamer/sbom/accepted/0000-sbom-command.md#sbom-generation-for-npm-projects)


ljharb:
Long before generation, I think we'd need to ensure that the following things are addressed:
 - validation tools, that can confirm that a generated SBOM is valid
 - since there's 6 (or 7?) kinds of SBOMs, and npm packages can be used for multiples of them, how would this command handle generating those? production deps are different than dev deps, and the build environment is important too and may have nothing to do with npm
 - a validation tool that package authors can run to determine if end users will be _able_ to generate a valid SBOM (and/or, help an author make needed changes to allow for a more complete SBOM to be created)

bdehamer:
> since there's 6 (or 7?) kinds of SBOMs...

Given that one of the primary functions of the npm CLI is to manage a project's dependencies, I see this new command being useful primarily for the creation of [Source- or Build-type SBOMs](https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf) where it is important to capture the dependencies for an artifact. 

Similar to the work done to capture package provenance, it's not hard to imagine the Build SBOM expanding in the future to also record information about the build process at the time of package publication. However, in the spirit of iterative improvement, generating a basic SBOM enumerating a project's dependencies seems like a good starting point.

> production deps are different than dev deps

I think I addressed this in the RFC already (see the `--omit` flag and the format-specific annotations for dev dependencies) but lemme know if you think I should further clarify.

> validation tools, that can confirm that a generated SBOM is valid

Not sure I understand this point -- the goal would be to have the CLI generate a valid SBOM in one of the two supported formats. I can imagine having a suite of integration tests to ensure that the SBOMs generated from the CLI are compatible with tools that consume SBOMs (things like [osv](https://github.com/google/osv-scanner), [GH Dependency submission API](https://docs.github.com/en/enterprise-server@3.8/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api), or [snyk](https://snyk.io/code-checker/sbom-security/)).

> ...determine if end users will be able to generate a valid SBOM

The CLI is already pretty good at detecting dependency issues. I would imagine that issues like missing or extraneous dependencies would be reported as errors when executing the command. Perhaps I should add some text to the RFC to make this explicit.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Unresolved Questions and Bikeshedding
+
+* There’s a case to be made that SBOM generation should just be a flag on the existing `npm-ls` command – an SBOM is really just a different format of the output already generated by `npm-ls` (in fact, the CycloneDX tool described above parses the output from `npm-ls` in order to generate its SBOM). However, many of the arguments supported by `npm-ls` don’t really make sense in the context of SBOM generation (package spec filtering, depth, etc…) so it feels like a bit of an awkward fit. \
+ \
+Making it a distinct command allows us to add SBOM-specific features in the future like a `--sign` flag which could be used to generate a signed SBOM. \
+ \
+_Recommendation: Add a distinct command for generating an SBOM._
+
+* SPDX doesn’t provide a natural way to differentiate between the various npm dependency types like “dev”, “peer”, and “optional”. We might consider using something like the [package comment field](https://spdx.github.io/spdx-spec/v2.3/package-information/#720-package-comment-field) to capture this information. \
```
puerco, line 146: This is not correct. The elements in the SBOM have to be related to a top element (the software being described by the SBOM). 

Each dependency package needs to be related to the top-level element. My suggestion is to use  `DEV_DEPENDENCY_OF` for dev dependencies, `OPTIONAL_DEPENDENCY_OF` for optional and for peer deps i think I would relate them with `PREREQUISITE_FOR`.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,427 @@
...
+  "dataLicense": "CC0-1.0",
+  "SPXID": "SPDXRef-DOCUMENT",
+  "name": "hello-world@1.0.0",
+  "documentNamespace": "http://spdx.org/spdxdocs/hello-world-1.0.0-<uuid>",
+  "creationInfo": {
+    "created": "2023-08-04T21:41:02.071Z",
+    "creators": [
+      "Tool: pkg:npm/cli@9.8.1"
+    ]
+  },
```
puerco, line 311: The document needs to define its top-level element, this is the entry point to the SBOM graph:
```suggestion
  },
  "documentDescribes": [
    "SPDXRef-Package-hello-world-1.0.0"
  ],

```

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,417 @@
...
+        }
+      ],
+      "checksums": [
+        {
+          "algorithm": "SHA512",
+          "checksumValue": "b0690fc7e56332d9..."
+        }
+      ]
+    }
+  ]
```
puerco, line 407: The document needs to have its relationships, it should look something similar to the following:

```suggestion
  ],
  "relationships": [
    {
      "spdxElementId": "SPDXRef-DOCUMENT",
      "relatedSpdxElement": "SPDXRef-Package-hello-world-1.0.0",
      "relationshipType": "DESCRIBES"
    },   
    {
      "spdxElementId": "SPDXRef-Package-hello-world-1.0.0",
      "relationshipType": "DEPENDS_ON",
      "relatedSpdxElement": "SPDXRef-Package-tsconfig.node14-1.0.3"
    },
    {
      "spdxElementId": "SPDXRef-Package-hello-world-1.0.0",
      "relationshipType": "DEPENDS_ON",
      "relatedSpdxElement": "SPDXRef-Package-debug-4.3.4"
    },
    {
      "spdxElementId": "SPDXRef-Package-debug-4.3.4",
      "relationshipType": "DEV_TOOL_OF",
      "relatedSpdxElement": "SPDXRef-Package-ms-2.1.2"
    }
  ]
```

Note the first relationship. It is the same data as the `documentDescribes` field above.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,417 @@
...
+    {
+      "name": "hello-world",
+      "SPDXID": "SPDXRef-Package-hello-world-1.0.0",
+      "versionInfo": "1.0.0",
+      "downloadLocation": "NOASSERTION",
+      "filesAnalyzed": false,
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE-MANAGER",
+          "referernceType": "npm",
```
puerco, line 322:
```suggestion
          "referenceType": "npm",
```

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,417 @@
...
+{
+  "spdxVersion": "SPDX-2.3",
+  "dataLicense": "CC0-1.0",
+  "SPXID": "SPDXRef-DOCUMENT",
+  "name": "hello-world@1.0.0",
+  "documentNamespace": "http://spdx.org/spdxdocs/hello-world-1.0.0-<uuid>",
+  "creationInfo": {
+    "created": "2023-08-04T21:41:02.071Z",
+    "creators": [
+      "Tool: pkg:npm/cli@9.8.1"
```
puerco, line 309: I think the tool version has to be appended with a dash:
```suggestion
      "Tool: pkg:npm/cli-9.8.1"
```

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,417 @@
...
+
+### SPDX
+
+The proposed SPDX SBOM generated for the project above would look like the following:
+
+```json
+{
+  "spdxVersion": "SPDX-2.3",
+  "dataLicense": "CC0-1.0",
+  "SPXID": "SPDXRef-DOCUMENT",
```
puerco, line 303:
```suggestion
  "SPDXID": "SPDXRef-DOCUMENT",
```

puerco (approved):
This is an awesome proposal @bdehamer, I am one of the contributors of the @SPDX project and I have some suggestions to improve the proposed output. 

With the changes I added, the SBOM ends up mirroring the structure of the sample project in the RFC, looking like this:

image

I'm attaching an SPDX document with the corrections I'm suggesting below, I hope it's useful!

[npm.spdx.json.gz](https://github.com/npm/rfcs/files/12296869/npm.spdx.json.gz)

goneall:
Super excited to see this RFC - Thanks @bdehamer for proposing this!

> validation tools, that can confirm that a generated SBOM is valid

Just for reference, we do have an [online validation tool](https://tools.spdx.org/app/validate/) for SPDX files which may be helpful during development of this feature to check if it is producing valid SPDX.  Since it is online, you wouldn't want to use it as part of the actual CI/CD or part of the NPM code itself.

We also have a command line implementation of SPDX validation in Java and one in Python.  I know the Java command line utility is used in some CI/CD environments to validate the produced SPDX file.  

Unfortunately, we don't have one (yet) in JavaScript.  We are looking for volunteers to implement JavaScript validation if anyone is interested in contributing.

We also have a [JSON Schema](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json) file you can use to validate the syntax.

accepted/0000-sbom-command.md
bdehamer: @puerco thanks for all of the great feedback!

I was a little unsure about the `relationships` collection. I had that in an early draft of the RFC but then removed it after noticing that a lot of the tools I looked at weren't including it (and none of the SPDX-consuming tools I played with seemed to care if it was present). If you think it would be useful, I'm happy to put it back in (I know CycloneDX has a similar structure for capturing relationships as well).

ljharb:
@bdehamer what i mean is, how can someone independently verify the validity of an SBOM? it's very easy to determine the validity of JSON or XML or anything with a schema - where's the open source validation tool that tells me that npm did the right thing?

feelepxyz:
> @bdehamer what i mean is, how can someone independently verify the validity of an SBOM?

@ljharb are you thinking of ways to validate that a SBOM can be reproduced given the same inputs later? One way could be to re-run `npm ci` given the original package json and lockfile and verify that the packages and versions you get in `node_modules` is exactly what's in the SBOM.

My understanding is that npm doesn't always provide reproducible installs of `node_modules`. But it would probably be enough for verifiable SBOMs if we can ensure we always get the same packages and versions, not that every single installed file is 100% reproducible.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+* <code>[spdx-sbom-generator](https://github.com/opensbom-generator/spdx-sbom-generator)</code> - A CLI for generating SPDX-style SBOMs for a number of different package managers (including npm). Currently, this tool only works with npm projects using <em>lockfileVersion</em> 1 so it’s not viable for a large number of projects (current <em>lockfileVersion</em> is 3)
+
+While you can effectively generate the same output we’re proposing with this combination of tools, there is value in having this capability supported directly in npm. Beyond the obvious developer-experience benefit of having SBOM generation baked-in to the CLI, it gives us a future path to do things like automatic-signing of SBOMs or integration of SBOMs into the package publishing process.
+
+
+## Implementation
+
+The `npm-sbom` command is similar in function to `npm-ls` command and will likely utilize a similar implementation. We’ll use <code>[arborist](https://github.com/npm/cli/tree/latest/workspaces/arborist)</code> to construct the dependency tree and the <code>[treeverse](https://github.com/isaacs/treeverse)</code> library to traverse the tree and assemble the SBOM.
+
```
jkowalleck, line 54: you might use [@cyclonedx/cyclonedx-library](https://www.npmjs.com/package/@cyclonedx/cyclonedx-library) to generate the CycloneDX SBOM. written in Typescript, already used by mentioned [@cyclonedex/cyclonedx-npm](https://www.npmjs.com/package/@cyclonedx/cyclonedx-npm)

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+
+### Local (Linked) Dependencies
+
+The generated SBOM will exclude any dependencies which are referenced via links on the local file system. A warning will be displayed for any linked dependency informing the user that it was omitted from the SBOM.
```
jkowalleck, line 58: For which reasons shall these dependencies be omitted?
Hiding these would drop important information.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+
+The <code>[externalRefs](https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field)</code> field will contain two <code>[PACKAGE-MANAGER](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f3-package-manager)</code> references, one using the <code>[npm](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f32-npm)</code> reference type and the other using the <code>[purl](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f35-purl)</code> reference type.
+
+
+## Unresolved Questions and Bikeshedding
+
+* There’s a case to be made that SBOM generation should just be a flag on the existing `npm-ls` command – an SBOM is really just a different format of the output already generated by `npm-ls` (in fact, the CycloneDX tool described above parses the output from `npm-ls` in order to generate its SBOM). However, many of the arguments supported by `npm-ls` don’t really make sense in the context of SBOM generation (package spec filtering, depth, etc…) so it feels like a bit of an awkward fit. \
```
jkowalleck, line 140: > in fact, the CycloneDX tool described above parses the output from `npm-ls` in order to generate its SBOM

`npm-ls` output is not sufficient for gathering all information a SBOM needs.
see https://github.com/CycloneDX/cyclonedx-node-npm/blob/main/docs/how.md

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+ \
+Making it a distinct command allows us to add SBOM-specific features in the future like a `--sign` flag which could be used to generate a signed SBOM. \
```
jkowalleck, line 142: > [...] like a `--sign` flag which could be used to generate a signed SBOM

disagree. see TODO-add-link

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+ \
+_Recommendation: Add a distinct command for generating an SBOM._
```
jkowalleck, line 144: agree

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+ 
+* Does `npm-sbom` command have a notion of a “default” SBOM format? Do we give preference to one of CycloneDX/SPDX or do we remain totally neutral (possibly at the expense of DX)? \
+ \
+_Recommendation: Default to SPDX if no format is specified._
```
jkowalleck, line 148: disagree. better be neutral.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+
+* Both CycloneDX and SPDX support multiple document formats (JSON, XML, Protocol Buffers, etc). Should we support output of multiple formats, or do we stick w/ JSON? \
+ \
+_Recommendation: Stick with JSON-only for the first version of this feature._
```
jkowalleck, line 152: agree.
NPM is a package manager for the JavaScript-centered `nodeJS`. 
JavaScript's native data format (for serialization and information-interchange) is JSON.

ljharb:
While that’s an important thing to have, i just mean, how do we know npm has implemented things correctly?

Is there anything package authors will be asked to “fix” to help produce better SBOMs? If so, how can package authors find this out in advance, before a slew of issues is filed?

jkowalleck (changes requested):
I see some points that need to be discussed and then reworked in the RFC text.
So I will take the chance to comment this Request-for-Comments ;-D


----

Full disclosure: I am the author of [@cyclonedex/cyclonedx-npm](https://www.npmjs.com/package/@cyclonedx/cyclonedx-npm) and [@cyclonedx/cyclonedx-library](https://www.npmjs.com/package/@cyclonedx/cyclonedx-library), and a member of the CycloneDX Core Working group.

PS: Speaking for the CycloneDX Core team, we welcome ecosystems to implementing CycloneDX SBOM themselves, We are always there to support, give guidance and answer questions. Feel free to reach out to us, preferably on our [official Slack server (invite link)](https://cyclonedx.org/slack/invite).
This being said, we would welcome the NPM community to sustain a CycloneDX generator themselves. We are looking forward to adding your organization and tools to our lists of list of CycloneDX-related [tools](https://cyclonedx.org/tool-center/) and [supporters (companies, vendors, projects)](https://cyclonedx.org/about/supporters/).

jkowalleck:
re https://github.com/npm/rfcs/pull/714#issuecomment-1672859758
> Is there anything package authors will be asked to “fix” to help produce better SBOMs? If so, how can package authors find this out in advance, before a slew of issues is filed?

the same it happens with `npm-ls` :-)

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+```json
+{
+  "$schema": "https://cyclonedx.org/schema/bom-1.4.schema.json",
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.4",
+  "serialNumber": "urn:uuid:f2fa9eae-72f1-430c-a9b3-986ffe05bc6e",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2023-08-10T00:19:08.697Z",
+    "tools": [
```
jkowalleck, line 204: tools are missing important information. 
- hashes
- externalReferences
  - website

see https://cyclonedx.org/docs/1.4/json/#metadata_tools

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+      {
+        "vendor": "npm",
+        "name": "cli",
+        "version": "9.8.1"
+      }
+    ],
+    "component": {
```
jkowalleck, line 211: `metadata.component` is missing the following important information: 
- licenses
- author
- description
- scope
- hashes
- externalReferences
  - vcs
  - issue-tracker
  - website
  - documentation

see https://cyclonedx.org/docs/1.4/json/#metadata_component

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+      "purl": "pkg:npm/hello-world@1.0.0",
+      "properties": [
+        {
+          "name": "cdx:npm:package:path",
+          "value": ""
+        }
+      ]
+    }
+  },
+  "components": [
```
jkowalleck, line 225: `components.component` are missing the following important information: 
- licenses
- author
- description
- scope
- hashes
- externalReferences
  - distribution -- most importent, to know where a package was sourced from 
  - vcs
  - issue-tracker
  - website
  - documentation

see https://cyclonedx.org/docs/1.4/json/#components

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+Adding SBOM generation to the tool which many developers are already using to manage their project dependencies eliminates any friction which may come from having to adopt/learn a separate tool. 
+
+
+## Detailed Explanation
+
+A new `sbom` command will be added to the npm CLI which will generate an SBOM for the current project. The SBOM will use the current project as the root and enumerate all of its dependencies (similar to the way `npm-ls` does) in one of the two supported SBOM formats. See the 
+[Example SBOMs](#example-sboms) section for sample CycloneDX and SPDX SBOM documents.
+
+Supported command options: 
+
```
jkowalleck, line 25: missing option:

SBOM may contain whether the system under analysis is an application or of other type.
see the examples in the RFC.
maybe the project under analysis is not an application, but an encapsulated library with bundled dependencies?
or it is a firmware for an embedded JavaScript (like [elk](https://github.com/cesanta/elk))
this is why [@cyclonedx/cyclonedx-npm](https://www.npmjs.com/package/%40cyclonedx/cyclonedx-npm) has an option for that:
```text
--mc-type <type>          Type of the main component.
                            (choices: "application", "firmware", "library", default: "application")
```

accepted/0000-sbom-command.md
stevespringett: Also disagree. If you look at the download stats from the various CycloneDX Javascript implementations, it's clear that CycloneDX is what most people are using today. But, I'd rather see npm be neutral and explicitly require the user to specify a format.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+`--sbom-format` - SBOM format to use for output. Valid values are “spdx” or “cyclonedx”. In the future, the set of valid values can be expanded to select differents versions of the supported output formats (e.g. "cyclonedx1.4" vs "cyclonedx1.5")
+
+`--omit` - Dependency types to omit from generated SBOM. Valid values are “dev”, “optional”, and “peer” (can be set multiple times). By default, all development, optional, and peer dependencies will be included in the generated SBOM unless explicitly excluded.
+
+`--package-lock-only` - Constructs the SBOM based on the tree described by the _package-lock.json_, rather than the contents of _node_modules_. Defaults to _false_. If the _node_modules_ folder is not present, this flag will be required in order to generate an SBOM.
+
```
nscuro, line 31: I know the CycloneDX module supports this option, but given this is about integration with NPM itself, I can't help but feel like this "shortcut" should not be supported. IMO, if it's a feature native to the package manager, it should have one, and only one, "correct" option.

The only reason for supporting the `--package-lock-only` flag (in my mind), is that you want to generate a BOM in workflows where you can't or don't want to build the project. Say, a "generate BOMs for all projects in my org, without developers having to change their pipelines" kind of thing. That is a different lifecycle stage (pre-build or earlier) than analyzing the packages you have installed, which will be post-build.

If this option *must* be supported, then there should be ways for consumers to differentiate between how the BOM was generated. CycloneDX v1.5 comes with support for [lifecycle phases](https://cyclonedx.org/docs/1.5/json/#metadata_lifecycles_items_oneOf_i0_phase), which could be of use. There are also ways to communicate *how* components where identified via [evidence](https://cyclonedx.org/docs/1.5/json/#components_items_evidence_identity), which even includes [confidence](https://cyclonedx.org/docs/1.5/json/#components_items_evidence_identity_confidence).

For example, if components were identified solely via `package-lock.json`, the confidence should be lower than if they were identified on disk, from `node_modules`. The evidence should then also define the [method](https://cyclonedx.org/docs/1.5/json/#components_items_evidence_identity_methods_items_technique) as `manifest-analysis`.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+`--sbom-format` - SBOM format to use for output. Valid values are “spdx” or “cyclonedx”. In the future, the set of valid values can be expanded to select differents versions of the supported output formats (e.g. "cyclonedx1.4" vs "cyclonedx1.5")
+
+`--omit` - Dependency types to omit from generated SBOM. Valid values are “dev”, “optional”, and “peer” (can be set multiple times). By default, all development, optional, and peer dependencies will be included in the generated SBOM unless explicitly excluded.
+
+`--package-lock-only` - Constructs the SBOM based on the tree described by the _package-lock.json_, rather than the contents of _node_modules_. Defaults to _false_. If the _node_modules_ folder is not present, this flag will be required in order to generate an SBOM.
+ 
+`--workspace` - Runs the command only in the context of the specified workspace (can be set multiple times).
+
+`--workspaces` - Runs the command in the context of all configured workspaces.
```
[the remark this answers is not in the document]
bdehamer: I'll update this section to be more clear. Both of these options only make sense in the context of a project utilizing workspaces. When generating an sbom in the context of a workspace-enabled project, the SBOM root is the base directory of the project (a single SBOM which aggregates the base project and the child projects in the workspace). The `--workspace` / `--workspaces` command just act as filters to control which projects in the workspace will be included in the output.

accepted/0000-sbom-command.md
bdehamer: I had considered proposing this, but I know that the npm team is quite intentional about the 3rd-party dependencies that get bundled with the CLI (carefully considering things like package size and long-term support). Given the deep dependency tree that `@cyclonedx/cyclonedx-library` brings with it, it seems unlikely that it would get accepted as a CLI dependency.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,461 @@
...
+  "hashes": [
+    {
+      "alg": "SHA-512",
+      "content": "3d15851ee494dde0ed4093ef9cd63b..."
+    }
+  ]
+}
+```
+
+The <code>[properties](https://cyclonedx.org/docs/1.4/json/#components_items_properties)</code> collection also provides for a standard property under the [npm taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/npm.md) for annotating development dependencies. For any package which was determined to be a development dependency of the root project, we would add the following to the <code>properties</code> collection:
```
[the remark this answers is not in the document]
bdehamer: Good call-out. Will add.

accepted/0000-sbom-command.md
bdehamer: Updated RFC to include `scope` and `externalReferences.distribution`. Other fields are mentioned as being populated as-available.

accepted/0000-sbom-command.md
bdehamer: Updated RFC to include `scope` and `externalReferences.distribution`. Other fields are mentioned as being populated as-available.

accepted/0000-sbom-command.md
bdehamer: I personally don't feel strongly one way or the other about this particular option -- and can't really say whether the I-need-to-generated-an-SBOM-without-also-building-my-project use case is even a thing. I added it mainly because other SBOM tools support it (and it's pretty easy to implement within the npm CLI).

If the consensus is that this is not useful, I'm happy to remove it (and maybe reconsider in the future if real use cases arise). Explicitly identifying package-lock-only SBOMs also seems reasonable.

accepted/0000-sbom-command.md
bdehamer: `hashes` is an interesting one as it relates to the tools -- this should obviously reference the npm CLI . . . but I'm not sure that npm knows what it's own hash is at runtime 😆 

Will have to dig into that one.

accepted/0000-sbom-command.md
ljharb: of course it's a thing; any task should be able to be run in isolation. Including this option makes perfect sense.

stevespringett: I would suggest adding support for CycloneDX lifecycle phases and including evidence and confidence as @nscuro suggested earlier. With an `--package-lock-only` option, lifecycles and evidence become much more important.

stevespringett: For additional information about lifecycles and evidence, refer to:

- https://cyclonedx.org/guides/sbom/lifecycle_phases/
- https://cyclonedx.org/guides/sbom/evidence/

This all ties into SBOM quality which OWASP CycloneDX defines here: 
- https://cyclonedx.org/guides/sbom/bom/#sbom-quality

accepted/0000-sbom-command.md
jkowalleck: I would just mention it as an argument how to simplify the implementation. Knowledge that there is a lib that does most things for you is a huge benefit when making decisions whether to accept a feature, don't you think?

It shall be to the developers to find reasons not to use an existing library. 

> Given the deep dependency tree that `@cyclonedx/cyclonedx-library` brings with it [...]
You are wrong. Only one new dependency. The `spdx-*` dependencies are already bundled with NPM.  The optional dependencies are not needed for the features described in your RFC. see 
```shellSession
$ npm i -S --omit optional @cyclonedx/cyclonedx-library
added 5 packages, and audited 6 packages in 10s

1 package is looking for funding
  run `npm fund` for details

found 0 vulnerabilities


$ npm ls --all
foo@1.0.0 /tmp/foo
└─┬ @cyclonedx/cyclonedx-library@4.0.0
  ├── UNMET OPTIONAL DEPENDENCY ajv-formats-draft2019@^1.6.1
  ├── UNMET OPTIONAL DEPENDENCY ajv-formats@^2.1.1
  ├── UNMET OPTIONAL DEPENDENCY ajv@^8.12.0
  ├── UNMET OPTIONAL DEPENDENCY libxmljs2@^0.31.0 || ^0.32.0
  ├── packageurl-js@1.0.2
  ├─┬ spdx-expression-parse@3.0.1
  │ ├── spdx-exceptions@2.3.0
  │ └── spdx-license-ids@3.0.13
  └── UNMET OPTIONAL DEPENDENCY xmlbuilder2@^3.0.2
```

bdehamer: That's great info! I'll add a note to the RFC and the npm team can decide.

accepted/0000-sbom-command.md
bdehamer: Was trying to err on the side of developer experience (i.e. fewer required flags) but also understand the case for format neutrality. 

Would appreciate someone from the npm team to weigh-in on the most idiomatic way to represent this -- I don't believe npm has any notion of "required" flags, so it would just need to be an argument to command:

```
npm sbom spdx
```

That doesn't feel quite right 🤔

Also would be nice if developers could set this value once (either via env var, .npmrc config, or package.json entry) to alleviate the burden of having to always specify this value.

stevespringett: Many orgs will be producing both formats. Having something set in .npmrc also feels wrong as the same build system may likely build different projects requiring different formats. One option is to automatically produce both formats with `npm sbom`.

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,493 @@
...
+
+All packages will specify a `false` value for the <code>[filesAnaylzed](https://spdx.github.io/spdx-spec/v2.3/package-information/#78-files-analyzed-field)</code> field to signal that no assertions are being made about the files contained within a package.
+
+The <code>[externalRefs](https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field)</code> field will contain two <code>[PACKAGE-MANAGER](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f3-package-manager)</code> references, one using the <code>[npm](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f32-npm)</code> reference type and the other using the <code>[purl](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f35-purl)</code> reference type.
+
+
```
jkowalleck, line 148: Could you add the missing the section "Prior Art"
see https://github.com/npm/rfcs/blob/main/accepted/0000-template.md#prior-art

> {{This section is optional if there are no actual prior examples in other tools}}

Actually, almost all features, switches, functionality, and implications described in the RFC are already implemented in the already mentioned tool: [@cyclonedex/cyclonedx-npm](https://www.npmjs.com/package/%40cyclonedx/cyclonedx-npm)
(which had a predecessor: [cyclonedx-node-module v3](https://github.com/CycloneDX/cyclonedx-node-module/tree/3.x))

jkowalleck (changes requested):

accepted/0000-sbom-command.md
bdehamer: My thinking was that an SBOM which references linked packages ends-up being very specific to the environment in which it is generated -- it would be difficult for anyone else to reason about or reproduce the SBOM contents.

That said, there may well be use cases where the user simply wants a strict representation of the project's dependencies even if some of those are links. Perhaps a flag that would include links? Or maybe the links are included by default but you still get a warning message in the output?

jkowalleck: before I go into details, let me ask the following: 
does `npm ls` list information about locally linked packages?

jkowalleck: for the same reasons `npm ls` lists these information, I'd argue they must be part of an SBOM.

> an SBOM which references linked packages ends-up being very specific to the environment in which it is generated

One would want to have that specific SBOM for the specific setup, all dependencies listed in my SBOM, regardless whether they are linked or not, so one can do proper risk assessment. 

> [...] it would be difficult for anyone else to reason about [...]

why would that be? Could you explain further? 

> [...] or reproduce the SBOM contents.

As long as I can reproduce the SBOM in the same specific environment, all is good.

here are example setups, so we have a common ground to discuss the meta:
- https://github.com/CycloneDX/cyclonedx-node-npm/tree/main/demo/local-workspaces
- https://github.com/CycloneDX/cyclonedx-node-npm/tree/main/demo/local-dependencies

accepted/0000-sbom-command.md
bdehamer: Looks like this has good overlap with the SPDX [primaryPackagePurpose](https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field) field

goneall: This is indeed the "purpose" of the primaryPackagePurpose.

accepted/0000-sbom-command.md
dlorenc: Right now everyone working on SBOMs knows and cares about the formats, that's not going to be the case for long. Making people pick and look this up is going to hurt long-term adoption of SBOMs. I'm in favor of just picking a default to avoid requiring choices.

If you care which SBOM you use, it's the same amount of work. If you don't, it's less mental energy, and the people that care need the least friction.

ljharb: That seems the better choice than trying to pick a “winner”.

jkowalleck: CycloneDX has some [Recognized file patterns](https://cyclonedx.org/specification/overview/#recognized-file-patterns), like `*.cdx.json`

I know it is common for SPDX files to match `*.spdx.json`, but i did not find the docs for it.

So npm's SBOM command could result in creating two files at once: `sbom.cdx.json` and `sbom.spdx.json`

bdehamer: I think it's important to be able to write the generated SBOM to stdout (npm users expect to be able to pipe/redirect output from the various commands). This would seem to eliminate the "output both formats" option (either writing directly to a file or emitting both SBOMs to stdout).

After talking w/ the npm dev team, it seems like creating a mandatory `--sbom-format` flag is do'able. This has the added benefit of automatically enabling the ability to set this value via env var or npmrc config for use cases where developers know they always want a specific format.

I'm going to update this bullet item with this new recommendation.

accepted/0000-sbom-command.md
```diff
@@ -162,7 +162,7 @@ _Recommendation: Add a distinct command for generating an SBOM._
 
 * Does `npm-sbom` command have a notion of a “default” SBOM format? Do we give preference to one of CycloneDX/SPDX or do we remain totally neutral (possibly at the expense of DX)? \
  \
-_Recommendation: Default to SPDX if no format is specified._
+_Recommendation: Remain neutral with regard to SPDX vs CycloneDX. Make the `--sbom-format` flag mandatory.
```
ljharb, line 165: npm should still be listing the options, at which point, why not default to all of them? being neutral would encourage users to support everything until the ecosystem (maybe) picks a winner.

accepted/0000-sbom-command.md
bdehamer: No one argued that the `npm-ls` command should be overloaded w/ SBOM support, so I've removed this question from the RFC.

accepted/0000-sbom-command.md
bdehamer: Have updated the recommendation to be neutral and use a mandatory option to force the user to make an explicit format selection.

accepted/0000-sbom-command.md
bdehamer: Added an option to capture the "type" which will be used to populate either the CDX `type` field or the SPDX `primaryPackagePurpose` field.

accepted/0000-sbom-command.md
```diff
@@ -207,51 +207,57 @@ The proposed CycloneDX SBOM generated for the project above would look like the
 
 ```json
 {
-  "$schema": "https://cyclonedx.org/schema/bom-1.5.schema.json",
+  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
```
ljharb, line 210: why would a schema for a security-based format be served over http? O.o

bdehamer: I believe this merely functions as an [identifier](https://github.com/CycloneDX/specification/blob/fd4d383658196992364e5d62568a48c431ace515/schema/bom-1.5.schema.json#L3) and is not actually resolvable at this URL.

jkowalleck: > https://github.com/npm/rfcs/pull/714#discussion_r1323681082
exactly

this string is an identifier, like a name. 

it is not expected or treated as a URL. it just is common practice to
- use a URL-like prefix to indicate ownership
- use a URL that is actually resolvable, so people have an easy way to fetch the related schema document.

ljharb: thanks, that makes sense, but it's very strange to me to use the http form as an identifier rather than the https form

mrutkows:
Feel free to try this util. for validation of either SDPX or CycloneDX formats: https://github.com/CycloneDX/sbom-utility#sbom-utility

goneall:
For SPDX validation, I would recommend either the [online tools validate function](https://tools.spdx.org/app/validate/) or the [tools-java command line utility Verify command](https://github.com/spdx/tools-java#verifier).

In addition to the schema validation, it validates some of the parameter string formats and relationship restrictions that can't be easily validated in the JSON schema (e.g. validating a license expression parses correctly).

accepted/0000-sbom-command.md
bdehamer: No one argued that the npm-ls command should be overloaded w/ SBOM support, so I've removed this question from the RFC.

wraithgar:
There is now an [implementation PR](https://github.com/npm/cli/pull/6801) up for this.

bdehamer:
Thanks to everyone for the great feedback/discussion! There seems to be general consensus that this feature is worth adding to the CLI and that the proposed approach is correct. 

I've got a :white_check_mark: from @puerco from the SPDX camp and would love to get the blessing from someone on the CycloneDX side (perhaps @mrutkows or @stevespringett). Once ratified we can move on to the next step . . .

> There is now an [implementation PR](https://github.com/npm/cli/pull/6801) up for this.

The code in ☝️ PR was used to generate the samples that appear at the bottom of the RFC.

riteshnoronha:
As sbom generator tools are updated on a regular basis, it would be good idea to monitor the quality of the sbom. https://github.com/interlynk-io/sbomqs helps by generating a quality score for the sbom, which can be used in the pipeline to accept or reject it. 
image

accepted/0000-sbom-command.md
```diff
@@ -0,0 +1,566 @@
...
+      "url": "https://registry.npmjs.org/debug/-/debug-4.3.4.tgz"
+    }
+  ],
+  "properties": [
+    {
+      "name": "cdx:npm:package:path",
+      "value": "node_modules/debug"
+    }
+  ],
+  "hashes": [
```
jkowalleck, line 91: these hashes probably belong to the `externalReferences` of type `distribution`,
not to the component.

the component is a folder in `node_modules` dir , ...

Whitecockl (approved):

wesleytodd:
My understanding of this is that the cli PR landed and this the `sbom` command shipped sometime last year. Should this PR be closed out/merged?

ljharb:
Closed, since the RFC was never approved or fully evaluated.

ljharb:
I think that without the public RFC calls, or a replacement RFC process, this entire repo should probably be archived.

FELIXrobust (approved):

owlstronaut:
Closing - the sbom command shipped in https://github.com/npm/cli/issues/6801 and has continued to evolve through additional features/refinements. The behavior in production has moved beyond what the original RFC text describes, so merging this document as a spec wouldn't accurately reflect the current command.

PR #850, opened 2025-12-17
Add RFC: Credential Provider Plugin Protocol for Secure NPM Authentication

pwoosam:
This RFC proposes the implementation of a credential provider plugin protocol for the NPM CLI to enable secure, dynamic retrieval of authentication tokens. The protocol will allow external credential providers to supply short-lived tokens at runtime, eliminating the need to store secrets in plaintext within `.npmrc` files or environment variables.

### DIFF
--- a/accepted/0000-credential-provider-plugin.md
+++ b/accepted/0000-credential-provider-plugin.md
@@ -11,16 +11,42 @@ Currently, NPM requires authentication tokens to be stored in plaintext within `
 
 This practice violates enterprise security policies and is a known vector for exploitation.
 
-The expected outcome is a secure, extensible mechanism for token retrieval that supports enterprise-grade authentication workflows and aligns with modern security standards.
+NPM also lacks a native hook to automatically acquire/refresh short-lived credentials. This discourages short token lifetimes (e.g., ≤ 60 minutes) because developers are forced into manual copy/paste rotation.
+
+A standardized credential provider protocol fixes this at the tooling layer and enables modern enterprise authentication workflows (brokered auth, device-bound tokens, MFA, etc.) without compromising developer experience.
+
+### Goals
+
+- Enable runtime token acquisition via a standardized provider interface.
+- Support short-lived tokens without manual rotation.
+- Support scoped registries and multiple registry configurations.
+- Support 3rd party registry authentication using npm cli and 3rd party credential providers.
+- Graceful fallback behavior if no provider is available.
+
+### Non-Goals
+- Mandate a single identity provider or authentication mechanism (protocol is IdP-agnostic).
+- Define a universal keychain-based storage requirement for Node tooling (providers may use OS keychains or brokers internally).
+- Persist returned tokens to disk (explicitly avoided).
+
 
 ## Detailed Explanation
 
-The proposed plugin protocol will define a standard interface for credential providers that can be invoked by the NPM CLI during authentication workflows. When a token is required (e.g., for publishing or installing from private registries), the CLI will call the configured credential provider, which will return a valid token. This avoids the need to persist tokens in `.npmrc` or environment variables.
+The proposed plugin protocol defines a standard interface for external credential providers that can be invoked by the NPM CLI during authentication workflows. When NPM needs credentials for a registry request (e.g., installing from or publishing to a private registry), the CLI will invoke a configured credential provider at runtime and receive a token in response. This design avoids persisting tokens in `.npmrc` or relying on environment variables as a long-term secret store.
 
-Key components:
-- CLI logic to detect when a credential is needed and invoke the provider.
-- Support for scoped registries and multiple providers.
-- Graceful fallback behavior if no provider is available.
+### How it works (high level)
+
+1. **Registry request requires auth**: NPM determines that a request to a registry requires authentication (install, publish, etc.).
+2. **Select provider**: NPM checks for a provider configured globally or for the specific registry host.
+3. **Invoke provider**: NPM executes the provider as a child process and receives a structured response over `stdout`.
+4. **Use token in-memory**: NPM uses the returned token for the outgoing HTTP request without writing it to disk.
+5. **In-process caching**: Respecting `expiresAt`, NPM will cache the token **in-memory for the lifetime of the process** to avoid repeated invocations during a single command (no persistence).
+6. **Fallback / failure handling**: If no provider is configured, NPM continues with existing credential behavior. If a provider is configured but fails, NPM surfaces a errors written to `stderr` and NPM will fallback to legacy sources.
+
+### Key components
+- **CLI auth flow integration**: Detect when credentials are required and invoke the configured provider.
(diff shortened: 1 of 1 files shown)


accepted/0000-credential-provider-plugin.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **NuGet Credential Providers:** NuGet supports external credential providers for secure token retrieval, widely adopted in enterprise environments.
+- **pip Keyring Integration:** Python’s pip can retrieve credentials from OS keychains via plugins.
+- **Git Credential Helpers:** Git supports credential helpers for dynamic authentication, including integration with OS keychains and cloud identity providers.
+
+These examples demonstrate the viability and benefits of plugin-based credential retrieval in developer tooling.
+
+## Unresolved Questions and Bikeshedding
+
+- How will credential providers be registered? (e.g. bundled with npm-cli, well-known globally installed node_module, well-known and dynamically installed by npm-cli, etc.).
```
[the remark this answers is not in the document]
pwoosam: I'll update this to allow project-level and user-level configuration. Noted that credential providers should be able to be provided without any action taken by NPM.

accepted/0000-credential-provider-plugin.md
```diff
@@ -0,0 +1,72 @@
...
+- Can custom credential providers be registered via a new setting in `.npmrc`?
```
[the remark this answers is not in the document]
pwoosam: `.npmrc` will have a configuration to register credential providers then. This is something that pnpm has implemented: https://pnpm.io/settings#urltokenhelper

However, pnpm's implementation has limitations that are not desirable:

> The configuration for the path to the helper must be an absolute path, with no arguments. In order to be secure, it is only permitted to set this value in the user .npmrc. Otherwise a project could place a value in a project's local .npmrc and run arbitrary executables.
>
> Setting a token helper for the default registry:
>
> > tokenHelper=/home/ivan/token-generator
>
> Setting a token helper for the specified registry:
>
> > //registry.corp.com:tokenHelper=/home/ivan/token-generator

The registry context isn't passed to the Credential Provider in pnpm's case, which is a challenge for generating credentials scoped to the registry. Ideally, the registry uri would be sent to the provider for context. Otherwise, a wrapper script would need to be created to supply the context, which is not a great solution.

accepted/0000-credential-provider-plugin.md
```diff
@@ -0,0 +1,157 @@
...
+#### Request (NPM → Provider)
+- Provider in invoked with exactly as specified by `<command>`
+- No **stdin**
+
+#### Response (Provider → NPM)
+- Provider writes error logs to **stderr**
+- Provider writes a single UTF-8 JSON object to **stdout**
+- Example **stdout**:
+    ```json
+    {
```
embetten: Cargo has a version field in their protocol api worth considering for future proofing. See:
https://doc.rust-lang.org/nightly/cargo/reference/credential-provider-protocol.html. 

There are instances in the past where this would have been useful in the nuget [aritfacts-credprovider ](https://github.com/microsoft/artifacts-credprovider) integration.

accepted/0000-credential-provider-plugin.md
```diff
@@ -0,0 +1,157 @@
...
+        "_auth": "",      // basic auth token
+        "expiresAt": 0    // unix timestamp
+    }
+    ```
+    ```json
+    {
+        "username": "",   // username
+        "_password": "",  // password
+        "expiresAt": 0    // unix timestamp
+    }
```
jmyersmsft: Looks like this block is missing its end-fence
```suggestion
    }
    ```
```

accepted/0000-credential-provider-plugin.md
```diff
@@ -0,0 +1,157 @@
...
+    - https://www.npmjs.com/package/@azure/identity
+    - This method would be more challenging because NPM cli would need to be updated to support new `CredentialProvider` interfaces and this would require more maintenance.
+- **@aws-sdk/credential-providers**: Similar to `@azure/identity`, this credential provider interface can be implemented in code to support customer authentication chains. Some of these implementations include: From SSO, cognito identity, container metadata service, etc.
+    - https://www.npmjs.com/package/@aws-sdk/credential-providers#createcredentialchain
+    - Similar challenges compared to `@azure/identity`
+- **NuGet Credential Providers:** NuGet supports external credential providers for secure token retrieval, widely adopted in enterprise environments.
+- **pip Keyring Integration:** Python’s pip can retrieve credentials from OS keychains via plugins.
+- **Git Credential Helpers:** Git supports credential helpers for dynamic authentication, including integration with OS keychains and cloud identity providers.
```
jmyersmsft: More prior art: [Cargo (Rust) credential providers](https://doc.rust-lang.org/cargo/reference/credential-provider-protocol.html)

Filyus:
Hi! I drafted an alternative protocol spec with LLM — bidirectional JSON via stdin/stdout (closer to Cargo/Git credential helpers), where provider receives full context (registry + scope + package + operation), with version negotiation, refresh flow, batch mode, and login/logout for OAuth/SSO. Attached.
[rfc-credential-provider-v2.md](https://github.com/user-attachments/files/26599471/rfc-credential-provider-v2.md)

embetten:
I'm concerned about essentially introducing a new install-time hook to execute arbitrary binaries before/during npm install. The ecosystem is actively moving toward disabling install scripts (#488 #861 ) pnpm v10 already blocks lifecycle scripts by default). Adding a new pre-install code execution vector seems contradictory to that trajectory — we'd be closing one door while opening another.

@Filyus's v2 RFC defines trust boundaries based on PATH resolution and node_modules, and restricts project .npmrc to provider names only (no absolute paths). A few questions on whether this is sufficient:

- PATH poisoning: If providers are resolved via PATH, an attacker who can modify PATH (e.g. via a malicious .env, shell profile, or CI misconfiguration) can redirect to a malicious binary. Is PATH-based resolution the right trust anchor, or should we require something stronger (e.g. a hash/checksum of the provider binary, or only allowing providers installed via a known package manager)?
- node_modules typosquatting: If providers can be resolved from node_modules, the npm-credential-provider-* naming convention creates a typosquatting surface. A malicious credprovider in node_modules could intercept credentials. The v2 RFC's "suggest only, never auto-execute" discovery mitigates this, but what about resolution order when a name matches both a global install and a local node_modules entry?

More broadly — I'd like to understand how npm thinks about trust boundaries before settling on a resolution/discovery mechanism. Are the global install folder and node_modules inherently trusted? Does npm see a need for integrity checking here (like a provider lockfile or hash), or is explicit user configuration considered sufficient trust?

Filyus:
Thanks, I think this is valid feedback. I revised and expanded the draft here: https://github.com/Filyus/unofficial-npm-credential-provider-rfc/blob/main/rfc-credential-provider.md

Changes include a stricter trust model, no arbitrary PATH resolution, no project-local `node_modules/.bin`, no project/workspace `.npmrc` provider config, suggest-only discovery, fail-closed ambiguity, optional integrity/allowlist hooks, retry context, provider capabilities, JSON Schema, tests, and Rust/Python PoCs.

PR #851, opened 2025-12-29
RFC: propose RFC for optional icon field in package.json

adithya1770:
This PR proposes to add an icon field in the package.json file, which shall enable image availability on the npm website. On the npm website, most packages do not have icons, which makes it difficult for users to navigate or find packages easily. Adding this field to package.json will expose it through the npm API and help in displaying images.

## References
Related to npm/cli#8878

### DIFF
--- a/icon.md
+++ b/icon.md
@@ -11,7 +11,7 @@ especially in large plugin or library ecosystems.
 
 ## Proposed Solution
 
-This RFC proposes adding an optional `icon` field to `package.json`.
+This RFC proposes adding an optional `icon` field to `package.json`
 This field allows package authors to explicitly specify a URL to an image
 that represents the package.
 


adithya1770:
> This would not just be for the website. Making it available in search API results means that clients could also display the icon.
> 
> **Edit:** Adding it to the search API results was the original purpose for this feature suggestion, but expanding it to the npm website makes sense.
> 
> Also, I originally envisioned this as allowing for sharing an icon primarily from a github repo without having to package it into the tarball. So maybe it could be placed in the `links` object:
> 
> ```
>         "links": {
>           "homepage": "https://github.com/ebaauw/homebridge-lib#readme",
>           "repository": "git+https://github.com/ebaauw/homebridge-lib.git",
>           "bugs": "https://github.com/ebaauw/homebridge-lib/issues",
>           "icon": "https://github.com/ebaauw/homebridge-lib/path/to/icon_file.png",
>           "npm": "https://www.npmjs.com/package/homebridge-lib"
>         }
> ```

Understood, that makes sense!

wraithgar:
This is a much bigger undertaking than just adding a link to the package.json.  There are very serious registry/hosting concerns here especially surrounding security.  Where will these images be hosted?  Is the npm registry supposed to host these like readme images?  What happens if these images point to hostnames that expire and are taken over by someone else?

Then the problem of "what should the value in the package.json be?" comes up.  If we rehost this like readme images, the package.json that is published will no longer match the package.json in the tarball for that version.  This is already an issue and we would not want to add new places where this difference can happen.

This represents quite a lot of registry work to make this work, it's more than just adding an attribute to the package.json.

justjam2013:
The main issue is that while I can add custom fields to the `package.json` file, these fields are not getting passed on in the search results. I wrote and maintain a plugin for [Homebridge](https://homebridge.io/). When searching for Homebridge plugins in npm, the results returned from the search API are all text based. In this age of graphic UIs this feels regressive and barbaric

So maybe a custom object could be added to the `package.json` file called `custom` (or whatever), whose contents would be custom fields that are passed through as-is to the results of the search API:
```
{
  "name": "homebridge-virtual-accessories",
  "version": "3.14.0",
  "description": "Virtual HomeKit accessories for Homebridge.",
  "author": "justjam2013",
  ...
  "custom": {
    "icon": "https://github.com/ebaauw/homebridge-lib/path/to/icon_file.png",
  }
```

The outcome is the same, albeit without the `icon` field being a field officially supported by npm.

The downsides to this approach though are that npm would not be able to use that icon url in its own UI, remaining a 1980's text-only UI. And with a custom object, npm is unable to control what gets passed and sooner of later, someone is going to add a field with a value containing a base64 encoded 4k image.

Additionally, this could be an iterative feature. If npm wanted to cache the icon for its own use, this could be something that is done in a future iteration. You don't have to swallow the entire elephant up front.

ljharb:
The readme of a package can have any image you like - if you're using tools that fail to render the readme, i'd invite you to consider that those may be the source of the barbarism :-)

ljharb:
I'm not saying there's no utility in an icon. I'm saying that adding an icon is a massive increase of complexity and failure modes and attack vectors, and "plain text is barbaric" isn't a universally held belief nor does it outweigh any of those concerns, let alone all of them, imo.

justjam2013:
That's a totally fair comment and I was out of bounds.

So let's strip away everything else and boil this down to the core ask: I would like to be able to add a custom field to the `package.json` file and have this data passed through as-is to the search API results.

In my particular case, I would like the pass through to be metadata about the package, which in my particular case are plugins in the Homebridge ecosystem. Specifically, that metadata would be a link to an icon, so that when Homebridge receives the search results from the search API, Homebridge can use this link to retrieve and display the plugin icon. This piece of metadata would have nothing to do with npm, although if npm wanted to, it could use it.

The options I see are:

1. As this is just metadata for Homebridge to use, simply require that the icon file be named a something specific, like `icon.png` and then retrieve it using the existing `repository` link. This requires no changes to the `package.json` file structure and it remains something used exclusively by Homebridge. It relies on process.
2. Add an `icon` field to the `package.json` file definition, which could be either an absolute URL or could be relative to the `repository` link. This requires a change to the `package.json` file definition. This means that the package author can place the icon file wherever they want and name it whatever they want. The advantage is also that if this field is part of the `package.json` file definition, then other packages and clients could make use of it.
3. Add a `custom` object to the `package.json` file definition, where a package can specify any custom metadata that has meaning within its ecosystem. This requires a change to the `package.json` file definition and would be extremely flexible. However this would leave npm without any form of control over this custom object and its contents and would be wide open for abuse (a base64 encoded 4k image could impact npm storage and search performance).

My ask is for option 2. You agree that it would/could be a useful addition, but have security concerns and that is reasonable. So my point is, instead of throwing out the baby with the bathwater, can the security concerns be addressed?

1. The file URL length is capped to 256 characters (or whatever makes sense)
2. The file URL is a path relative to the `repository` link. If the icon file is now a security risk, then the entire repo is a security risk
3. The file size is restricted and an http/s HEAD request could check it by checking the `content-length` header
4. The file format is restricted to web friendly image formats, like `png` and `jpeg`, and maybe `webp`. A `png` file will have "‰PNG" on the first line, a `jpeg/jpg` file will have "JFIF" somewhere near the start.

The first two items can be checked at publish time and will not change, as they are part of the package's `package.json` file. The last two items can change, but they can be checked in flight.

As far as using the icons in the npm UI, that is a separate concern. This field can exist without npm having to use it. But if npm wanted to use the icons, the paginated UI by default shows 20 entries per page. The file sizes and formats for 20 images can be checked in flight without adding significant or noticeable overhead.

ljharb:
I mean, you could already just start doing this - make an "icon" field with whatever you like, and have homebridge read that field from the packument to get the icon. The only reason to integrate it into npm and change the official definition is if npmjs.com would show it, and/or to give it legitimacy.

I agree that with your modifications, most of the concerns go away except for:
1. it's still registry-side work that github would have to roadmap
1. showing an image in search results escalates a trust & safety burden, since images need to be monitored for various problems on a much larger scale

justjam2013:
The issue is that if you get a list of 100 results from the search API, you cannot retrieve the the packument for all 100 results, process them then retrieve the icons. I mean we could, but it's a long way around. At that point it would be much faster to simply use the `repository` link to retrieve a known-named file from the repo.
Having the link in the search results provides the flexibility to name it anything you want without incurring the cost of the double call (packument + image).

> The only reason to integrate it into npm and change the official definition is if npmjs.com would show it, and/or to give it legitimacy.

The idea was for any client to use it, not specifically including npmjs.com, but not excluding it either. And I wouldn't say "give it legitimacy", but rather to make it standard, as opposed to process based.

wraithgar:
> So while I fully understand the concerns and they certainly are valid concerns, npm is not vetting or sanitizing image links it is currently serving through its pages

I'm not sure how you came to this conclusion.  All relative images in a readme are linked via `raw.githubusercontent.com`.  All absolute images in a readme are rehosted via `camo.githubusercontent.com`.  This is the same process used in GitHub itself and you can read more at https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls

This does represent quite a large undertaking from a trust and safety perspective.  I would not recommend that folks who proxy search results for their own sites start picking up un-sanitized entries from package.json files and linking them on their own sites.

As it stands now, none of this would have anything to do with the npm cli itself, which is what this RFC repo is for.  This is fully a feature that would need to be implemented in the registry.  Discussions for things like that should take place at https://github.com/orgs/community/discussions/categories/npm.  There isn't really anything for the cli to do here, as even if this field were added it wouldn't need to be included in the normal search output since the cli is text based.

PR #857, opened 2026-03-16
Exclude node_modules from Time Machine backups

### LINKED ISSUE #471: [RRFC] Exclude `node_modules` from Time Machine backups (macOS)
## Motivation
`node_modules` often contains a lot of files that change frequently, which can significantly slow down backup solutions like Time Machine, so it would make sense to exclude them from backups.

### Current Behaviour
`node_modules` folders are included in Time Machine backups

### Desired Behaviour
Exclude `node_modules` from Time Machine backups using a "sticky exclusion". This can be done by adding the xattr `com.apple.metadata:com_apple_backup_excludeItem` to the `node_modules` folders with the value `com.apple.backupd` encoded as a binary plist:
```node
00000000  62 70 6C 69 73 74 30 30 5F 10 11 63 6F 6D 2E 61  |bplist00_..com.a|
00000010  70 70 6C 65 2E 62 61 63 6B 75 70 64 08 00 00 00  |pple.backupd....|
00000020  00 00 00 01 01 00 00 00 00 00 00 00 01 00 00 00  |................|
00000030  00 00 00 00 00 00 00 00 00 00 00 00 1C           |.............|
0000003d
```

## References
Rust already does this for their `target` folders: https://github.com/rust-lang/cargo/pull/4386
My PR for implementing this for the `poetry` Python package manager: https://github.com/python-poetry/poetry/pull/4599

guillaumekh:
## What this PR does

This PR adds an opt-in `time-machine-exclude` boolean flag to npm configuration.

When enabled, on macOS machines, it sets a file system extended attribute on the `node_modules` folder that excludes it from Time Machines backups.

## Motivation

`node_modules` directories can take a lot of space, and contain a lot of files, resulting in slower Time Machine backups and wasted backup storage space.

## References

- Fixes https://github.com/npm/rfcs/issues/471
- [2021-10-13 meeting notes](https://github.com/npm/rfcs/blob/816944d6fe9c85130a5a7a92883c29c69e5777e2/meetings/2021-10-13.md?plain=1#L94)

## Prior art

- Rust cargo: https://github.com/rust-lang/cargo/pull/4386
- python-poetry by @probablykasper: https://github.com/python-poetry/poetry/pull/4599
- Myriad of other [projects, dotfiles and scripts](https://github.com/search?q=%22com.apple.metadata%3Acom_apple_backup_excludeItem%22&type=code)

## Disclaimer

The PR was drafted with Claude, but tested and edited by me. If something stupid slipped in, it's on me.

### DIFF
--- a/accepted/0050-exclude-node-modules-from-time-machine.md
+++ b/accepted/0050-exclude-node-modules-from-time-machine.md
@@ -15,16 +15,13 @@ and therefore carry no unique, irreplaceable data.
 
 On macOS, Apple's Time Machine backup tool backs up every file that it can see,
 including the often tens-of-thousands of small files inside `node_modules`. This
-has several painful side-effects:
+has several regrettable side-effects:
 
-* **Slow incremental backups.** Each `npm install` or `npm update` touches many
+- **Slower backups.** Each `npm install` or `npm update` touches many
   files, forcing Time Machine to scan and snapshot all of them on the next backup
   run.
-* **Rapid backup-disk consumption.** Accumulated snapshots of large
-  `node_modules` trees eat disk space quickly and can cause backups to fail
-  entirely when the backup volume fills up.
-* **Slower system performance during backups.** The high file count in
-  `node_modules` can cause noticeable I/O spikes while Time Machine is indexing.
+- **Higher backup-disk consumption.** Accumulated snapshots of large
+  `node_modules` trees eat disk space quickly.
 
 The xattr-based "sticky exclusion" mechanism provided by macOS lets any tool
 flag a directory so that it is permanently skipped by Time Machine — even when
@@ -43,7 +40,7 @@ same mechanism used by the Xcode build toolchain for its `DerivedData`
 directories, and is equivalent to running:
 
 ```sh
-tmutil addexclusion -p /path/to/node_modules
+tmutil addexclusion /path/to/node_modules
 ```
 
 ### When the attribute is applied
@@ -52,6 +49,8 @@ When the opt-in flag is set, npm should apply this attribute on macOS
 (i.e. `process.platform === 'darwin'`) as soon as a `node_modules` directory is
 created on disk — that is, at directory-creation time inside
 `@npmcli/arborist`, before any package files are written into it.
+
+According to the [2021-10-13 meeting notes](https://github.com/npm/rfcs/blob/816944d6fe9c85130a5a7a92883c29c69e5777e2/meetings/2021-10-13.md?plain=1#L94), it was agreed the setting should be first introduced as opt-in.
 
 ### Setting the attribute
 
@@ -75,9 +74,9 @@ or when the `xattr` binary is unavailable.
 
 A new boolean flag `time-machine-exclude` is added to npm's configuration:
 
-| Key                    | Default  | Description |
-|------------------------|----------|-------------|
-| `time-machine-exclude` | `false`  | When `true`, marks `node_modules` with the macOS Time Machine exclusion xattr after every install on macOS. |
+| Key                    | Default | Description                                                                                                 |
+| ---------------------- | ------- | ----------------------------------------------------------------------------------------------------------- |
+| `time-machine-exclude` | `false` | When `true`, marks `node_modules` with the macOS Time Machine exclusion xattr after every install on macOS. |
 
 Users who want `node_modules` excluded from their backups set
 `time-machine-exclude=true` in their `.npmrc` (or via `npm config set`).
@@ -89,11 +88,12 @@ Users who want `node_modules` excluded from their backups set
 npm could apply the xattr on every macOS install and provide a flag to opt back
 in to backups. This would protect the majority of users automatically.
 
-**Drawback:** Silently preventing something from being backed up — without the
-user asking for it — is surprising behaviour and has caused concern among npm
-collaborators. A user who relies on their backup for disaster recovery may not
-realise `node_modules` is excluded until they attempt a restore. The opt-in
-approach ensures the user has made a deliberate choice.
+**Drawback:** Silently preventing something from being backed up without the
+user asking for it is likely to surprise some users and has some maintainers have
+[raised concerns](https://github.com/npm/rfcs/issues/471#issuecomment-943215770).
+A user who relies on their backup for disaster recovery may not realise `node_modules`
+is excluded until they attempt a restore. The opt-in approach ensures the user
+has made a deliberate choice.
 
 ### Alternative 2 – Write the value as a binary plist
 
@@ -103,8 +103,9 @@ approach ensures the user has made a deliberate choice.
 **Drawback:** Generating a binary plist in Node.js without a native library
 requires manual `Buffer` construction (or shelling out to an additional tool).
 In practice, Time Machine respects the attribute whether the value is a raw
-string or a binary plist — many projects write it as a plain string with the
-same effect. Plain strings are simpler and carry no additional risk.
+string or a binary plist — [many projects](https://github.com/search?q=%22com.apple.metadata%3Acom_apple_backup_excludeItem%22&type=code)
+write it as a plain string with the same effect. Plain strings are simpler
+and carry no additional risk.
 
 ### Alternative 3 – Use `tmutil addexclusion`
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):
this would be great

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
+# Exclude `node_modules` from macOS Time Machine backups
+
+## Summary
+
+On macOS, npm should provide an opt-in flag that marks a project's `node_modules`
```
lgarron, line 5: I feel like "should" isn't doing enough heavy lifting here, as it implies non-committal beahviour by `npm` is adopted (at least using the interpretation of "should" most widely used by specs and RFCs: https://www.ietf.org/rfc/rfc2119.txt).

I think it would be a lot more clear to state what *will* happen if this is adopted.
(Same throughout the doc.)

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+directory with the extended attribute that tells Time Machine to skip it. This
+prevents `node_modules` — which can be fully reconstructed from `package-lock.json`
```
lgarron, line 7: This is empirically inaccurate. Packages can and do place various state inside `node_modules` in all sorts of ways, either during post-install or other at other times.

For example:

- Downloading platform-specific binaries, into:
  - Other package folders.
  - Their own package folder
  - `node_modules/bin`
- Downloading data
- Caches

These can be created or downloaded (from arbitrary URLs) using arbitrary code, and there is no guarantee that running `npm install` on a repo in the future will produce a compatible result (much less an identical result).

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+— from filling up backup disks and slowing down backup runs.
+
+## Motivation
+
+`node_modules` directories are, in effect, a local cache derived entirely from
+`package-lock.json`. They can be recreated at any time with a single `npm install`
+and therefore carry no unique, irreplaceable data.
```
lgarron, line 14: Again, I think this is an overly bold claim. I think a thorough survey of the ecosystem is warranted if this may ever become opt-out.

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+According to the [2021-10-13 meeting notes](https://github.com/npm/rfcs/blob/816944d6fe9c85130a5a7a92883c29c69e5777e2/meetings/2021-10-13.md?plain=1#L94), it was agreed the setting should be first introduced as opt-in.
+
+### Setting the attribute
+
+Because Node.js does not provide a built-in API for extended attributes, npm
+should shell out to the macOS `xattr` command-line utility, which ships with
+every macOS installation:
+
+```sh
+xattr -w com.apple.metadata:com_apple_backup_excludeItem \
```
lgarron, line 62: I would specify `/usr/bin/xattr`.

Otherwise, another `xattr` binary on the `$PATH` could take precedence, and it is notoriously difficult to use a consistent `$PATH` across all programs on macOS. So it could produce inconsistent behaviour on a single system.

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+  com.apple.backupd \
+  /path/to/node_modules
+```
+
+Writing the value as a plain UTF-8 string is sufficient — Time Machine respects
+the attribute regardless of whether the value is a raw string or a binary plist.
+The call is made with `child_process.execFile` (non-blocking, errors suppressed)
+so that it does not affect npm's exit code or performance on non-macOS systems
+or when the `xattr` binary is unavailable.
```
lgarron, line 71: What happens if the failure is a flake? Is this "best-effort", e.g. is it expected behaviour in that case that the flag is sometimes not successfully set?

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+directories, and is equivalent to running:
+
+```sh
+tmutil addexclusion /path/to/node_modules
+```
+
+### When the attribute is applied
+
+When the opt-in flag is set, npm should apply this attribute on macOS
+(i.e. `process.platform === 'darwin'`) as soon as a `node_modules` directory is
```
lgarron, line 49: 1. What if the folder already exists?
2. What if there are multiple `node_modules` inside ancestor dirs?

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+and carry no additional risk.
+
+### Alternative 3 – Use `tmutil addexclusion`
+
+`tmutil addexclusion /path/to/node_modules` is the official command-line
+interface for adding a "sticky" backup exclusion.
+
+**Drawback:** `tmutil` is significantly slower than writing the xattr directly,
+which would noticeably increase install time. I'm not sure what extra work tmutil
+does which explain the longer execution time.
```
lgarron, line 117: Is there any documentation that does explain it? If there is a difference, it seems good to know before-hand.

```suggestion
which would noticeably increase install time. I'm not sure what extra work `tmutil`
does which explain the longer execution time.
```

accepted/0050-exclude-node-modules-from-time-machine.md
```diff
@@ -0,0 +1,213 @@
...
+const { execFile } = require("child_process");
+
+const ATTR_NAME = "com.apple.metadata:com_apple_backup_excludeItem";
+const ATTR_VALUE = "com.apple.backupd";
+
+function excludeFromTimeMachine(nodeModulesPath) {
+  if (process.platform !== "darwin") {
+    return;
+  }
+  execFile("xattr", ["-w", ATTR_NAME, ATTR_VALUE, nodeModulesPath], () => {
```
lgarron, line 177: As mentioned above, `/usr/bin/xattr` might be warranted.

lgarron:
I came across [a link to this on Mastodon](https://hachyderm.io/@meego/116330472788177321) inviting comments. Just stopping by to weigh in, as someone who spends a non-trivial time managing the interaction between `node_modules` and backup systems (because there is value in backing them up but also the sheer number and size of files across a handful of projects can greatly slow down backups).

I do think this is a useful feature, but I think it is important to keep this opt-in, and for any opt-out mechanism to have a long and clear announcement period (perhaps with messages in the CLI when the flag is set).

In particular:

- Anyone who's been backing up their projects so far has been able to rely on the fact that a restore generally brings the `node_modules` for a project into full working order. It would be a breaking change to change this, and the time when you restore a backup is the absolute worst time to find out something like this has changed.
- My understanding is that it's much more common for `node_modules` to include state other than just installed package files — in particular, a lot of packages perform post-install work like downloading binaries and setting up caches inside `node_modules` using arbitrary logic. So the claims in the summary and motivation aren't quite accurate. (I would love it if `node_modules` was used in a less stateful way, like how Rust handles its deps. But in practice it can a bigger issue to lose local state in `node_modules`.)
- If this lands, I think the intended semantics should be *very thoroughly* described, and any deviations in functionality from this should be considered important bugs to fix. For example:
  - If `node_modules` already exists, when is the flag set?
  - Are other package managers expected to set this flag? Do they need to do anything carefully to avoid breaking it?
  - Are there any caveats when either `node_modules` itself or packages inside are symlinked?
  - If a package has `node_modules` inside nested folders, under what circumstances are they affected? Which program/command is responsible for this?

accepted/0050-exclude-node-modules-from-time-machine.md
lgarron, line 177: Also, depending on what value is passed as `nodeModulesPath`, it can be interpreted as a flag. (For example, `--my-css-custom-property/node_modules` can be a valid path if someone decided to name their project folder `--my-css-custom-property`.)

Unfortunately, it does not look like `xattr` supports `--` to delineate the file arg. The most robust thing would be to use lower-level APIs, but resolving `nodeModulesPath` to an absolute path might be a safe middle ground.

owlstronaut (changes requested):
The concept seems good to me. The prior art from Cargo/Poetry is reassuring. I think this needs anothe rpass before it's ready. @lgarron comments cover most of what I'd flag, especially that `node_modules` isn't really just a derivable cache and the `/usr/bin/xattr` + absolute path points. When does the xattr get applied? After every install, or at directory creation? We also need info about what happens for workspaces, pre-existing node_modules, or global installs.

PR #861, opened 2026-04-08
Add option to require install script approval

everett1992:
This is an attempt to revive https://github.com/npm/rfcs/pull/488 and control which packages can run install lifecycle scripts. Unlike [Make npm install scripts opt-in](https://github.com/npm/rfcs/pull/488) this RFC doesn't change any default behavior. It adds a new flag that when enabled errors or skips unapproved postinstall scripts. 

[Rendered here](https://github.com/everett1992/rfcs/blob/skip-scripts/accepted/0000-install-script-allowlist.md)

I'm able to implement this.

## References
Related to #488
https://github.com/npm/npm/issues/15144
https://github.com/npm/cli/issues/619
https://github.com/npm/cli/issues/538
https://github.com/npm/rfcs/discussions/80

### DIFF
--- a/accepted/0000-install-script-allowlist.md
+++ b/accepted/0000-install-script-allowlist.md
@@ -1,0 +1,274 @@
+# Install Script Allowlist
+
+## Summary
+
+Add an opt-in mode where npm requires explicit approval before running install
+lifecycle scripts (`preinstall`, `install`, `postinstall`, and auto-detected gyp
+files). The default behavior is unchanged — scripts run as they do today. Users
+who enable the feature choose between two enforcement modes: `ignore` (silently
+skip unapproved scripts) and `error` (fail the install). Approved packages are
+tracked in a JSON allowlist that can live in `package.json`, in one or more
+standalone JSON files referenced by `.npmrc`, or both — all sources are merged
+together.
+
+## Motivation
+
+This proposal builds on [npm/rfcs#488](https://github.com/npm/rfcs/pull/488) by
+@tolmasky and the extensive community discussion there. The core security
+argument from that RFC still holds and has only grown stronger:
+
+- Install lifecycle scripts execute arbitrary code at install time, before the
+  consuming project ever `require`s or `import`s the package. Only ~0.6% of
+  packages on npm use install lifecycle scripts (12,035 out of ~2 million
+  packages as of November 2021, per
+  [tolmasky's analysis in RFC #488](https://github.com/npm/rfcs/pull/488)), yet
+  they represent the easiest vector for supply-chain attacks.
+- Real-world exploits continue to demonstrate the risk. The 2021 `coa`/`rc`
+  compromises, the 2025–2026 "Shai Hulud" worm, and numerous typosquatting
+  attacks all leveraged postinstall scripts to propagate.
+- Alternatives to install lifecycle scripts have matured. N-API,
+  prebuild-install, and platform-specific optional dependencies
+  (`os`/`cpu`/`libc` fields) cover the many legitimate use cases that
+  historically required compilation at install time.
+
+The original RFC proposed changing the default so that scripts are off unless
+explicitly allowed. That was rejected as too disruptive. This proposal takes a
+different approach: **the default does not change**. Instead, npm gains a new
+opt-in enforcement mode with an allowlist, giving security-conscious users and
+organizations a first-class way to control which packages may run scripts —
+without breaking anyone who doesn't opt in.
+
+For organizations managing large codebases, shared management of the allowlist
+is essential. Organizations need to populate a shared allowlist with every
+package whose scripts are known to be in use, then enable `error` mode globally
+so that any _new_ unapproved script is caught immediately. The allowlist must be
+shareable across repositories without requiring changes to each project's
+`package.json`. See [Allowlist sources](#allowlist-sources) and
+[bikeshedding](#unresolved-questions-and-bikeshedding) for how this could work.
+
+## Detailed Explanation
+
+### Enforcement mode
+
+A new `.npmrc` config value controls the feature:
+
+```ini
+install-script-policy=off|ignore|error
+```
+
+- `off` (default): Current behavior. All install lifecycle scripts run
+  unconditionally.
+- `ignore`: Unapproved scripts are silently skipped. At the end of
+  `node_modules` construction, npm prints a summary of every package whose
+  scripts were skipped.
+- `error`: Unapproved scripts are skipped. At the end of `node_modules`
+  construction, npm exits non-zero and prints the list of unapproved packages
+  that had scripts.
+
+In both `ignore` and `error` modes, npm collects the full list of skipped
+packages and reports them together at the end of the install, rather than
+failing on the first one. This makes it practical to build an allowlist in one
+shot.
+
+### Allowlist format
+
+The allowlist is a JSON object with a single top-level field:
+
+```json
+{
+  "approvedInstallScripts": {
+    "node-gyp-build": "allowed",
+    "esbuild": "allowed",
+    "esbuild@>=0.25.0": "allowed",
+    "sqlite3@4.x": "allowed",
+    "husky": "ignored",
+    "sponsorware-nag": "ignored"
+  }
+}
+```
+
+Keys are `<package-name>` or `<package-name>@<semver-range>`. A bare package
+name (no `@range`) is equivalent to `@*` and matches all versions. Semver ranges
+in this list match pre-release versions.
+
+Values:
+
+- `"allowed"` — the package's install lifecycle scripts will run.
+- `"ignored"` — the package's install lifecycle scripts will not run, and the
+  package will not be listed as an error in `error` mode. This is useful for
+  packages whose scripts are non-essential (telemetry, sponsorship messages,
+  optional native compilation with a JS fallback, etc.).
+
+### Allowlist sources
+
+We need allowlist entries to come from at least two places because npmrc's
+config model flattens values — a project-level `.npmrc` replaces a global-level
+value rather than merging with it. If the allowlist lived only in `.npmrc`, a
+project that sets `install-script-allowlist=project-scripts.json` would silently
+discard the global `install-script-allowlist=company-scripts.json`. That defeats
+the central-management use case entirely.
+
+The proposed solution uses two complementary sources:
+
+1. **External JSON files referenced by `.npmrc`** — for shared and global
+   config. An organization sets this once in a global or user `.npmrc` and every
+   project inherits it.
+
+   ```ini
+   install-script-allowlist[]=/etc/npm/approved-scripts.json
+   ```
+
+2. **`package.json`** — for per-project entries that are tracked in source
+   control alongside the code.
+
+   ```json
+   {
+     "name": "my-app",
+     "approvedInstallScripts": {
+       "esbuild": "allowed"
+     }
+   }
+   ```
+
+All sources are merged (unioned). If the same key appears in multiple sources,
+the most permissive value wins (`"allowed"` > `"ignored"`). The goal is to
+reduce toil — an organization can maintain one shared list and have every
+project inherit it automatically, while individual projects can still add their
+own entries. Whether a project should also be able to _override_ a shared entry
+(e.g. downgrade `"allowed"` to `"ignored"`) is an open question, but not a hard
+requirement either way.
+
+This two-source approach is a concrete proposal, but we're open to better ideas
+for achieving the same goal: a project-level list that coexists with a
+centrally-managed list. If there's a cleaner way to get mergeable lists out of
+npmrc, or a different file format that works better, we'd welcome that.
+
+### Interaction with `--ignore-scripts`
+
+`--ignore-scripts` continues to work as today — it disables all scripts
+unconditionally, regardless of the allowlist. The allowlist only applies when
+`install-script-policy` is `ignore` or `error` and `--ignore-scripts` is not
+set.
+
+### Scope
+
+This feature applies to the install lifecycle scripts that run during
+`npm install` for dependencies:
+
+- `preinstall`
+- `install`
+- `postinstall`
+- Implicit `node-gyp rebuild` (when a `binding.gyp` is detected and no `install`
+  script is defined)
+- `prepare` scripts of git dependencies (which run during install to build the
+  package from source)
+
+It does not affect:
+
+- Scripts in the root project and workspaces' `package.json` (the package being
+  developed). It's specifically for installed dependencies.
+- `npm run` / `npm test` / `npm start` etc.
+
+## Rationale and Alternatives
+
+### Alternative 1: Change the default (tolmasky's original RFC #488)
+
+Making scripts opt-in by default is the ideal end state, but the npm team and
+community rejected it as too disruptive in 2021–2022. This proposal provides the
+same security benefit for users who opt in, without breaking anyone. It could
+serve as a stepping stone toward eventually changing the default in a future
+major version, once the ecosystem has had time to adapt.
+
+### Alternative 2: Use `@lavamoat/allow-scripts` or similar userland tools
+
+Tools like `@lavamoat/allow-scripts` and `can-i-ignore-scripts` exist and work
+today. However, they require adding a dependency and a workflow change to every
+project. A first-party npm feature would be more discoverable and require no
+extra dependencies.
+
+### Alternative 3: Migrate to `pnpm`
+
+Package owners that would opt-in to install script allowlists could migrate to
+pnpm instead. I know this would be a difficult migration for the codebase I
+manage.
+
+## Implementation
+
+The implementation touches a few areas of the npm CLI:
+
+1. **Config**: Add `install-script-policy` (enum: `off`, `ignore`, `error`) and
+   `install-script-allowlist` (array of file paths) to the config schema.
+
+2. **Allowlist loading**: At the start of `npm install` / `npm ci`, read and
+   merge the allowlist from all configured sources (package.json + external
+   files). Build a lookup structure mapping `(package-name, version)` →
+   `allowed | ignored | unapproved`.
+
+3. **Script gating**: In the lifecycle script runner (likely in
+   `@npmcli/run-script` or the reification step in `@npmcli/arborist`), before
+   executing an install lifecycle script for a dependency, check the merged
+   allowlist. If the package+version is not approved, skip the script and record
+   it.
+
+4. **Reporting**: After reification completes, if any scripts were skipped,
+   print a summary. In `error` mode, exit non-zero if any skipped packages were
+   not marked `"ignored"`.
+
+## Prior Art
+
+- [npm/rfcs#488](https://github.com/npm/rfcs/pull/488) — @tolmasky's "Make npm
+  install scripts opt-in" RFC (2021). Extensive discussion, 369 👍, not merged
+  due to breaking-change concerns.
+- [pnpm `onlyBuiltDependencies` / `allowBuilds`](https://pnpm.io/settings#allowbuilds)
+  — pnpm 10 blocks lifecycle scripts by default with an allowlist in
+  `package.json`.
+- [`@lavamoat/allow-scripts`](https://www.npmjs.com/package/@lavamoat/allow-scripts)
+  — Userland tool that disables scripts via `--ignore-scripts` and selectively
+  re-runs them from an allowlist in `package.json`.
+- [`can-i-ignore-scripts`](https://www.npmjs.com/package/can-i-ignore-scripts) —
+  Helps audit which dependencies actually need install lifecycle scripts.
+
+## Unresolved Questions and Bikeshedding
+
+### Config key names
+
+`install-script-policy`, `install-script-allowlist`, and the JSON field
+`approvedInstallScripts` are suggestions. The names should be bikeshedded. What
+matters is the semantics: a mode toggle, a list of external config files, and a
+mergeable allowlist structure.
+
+I proposed using both package.json fields and npmrc because that allows package
+and global config. Another option could be to use npmrc, but merge the arrays
+across rc files instead of replace them.
+
+```
+# ~/.npmrc
+install-script-allowlist[]=/etc/npm/shared-approved-scripts.json
+
+# /project/.npmrc
+install-script-allowlist[]=./approved-scripts.json
+
+# runtime sees ['/etc/npm/shared-approved-scripts.json', './approved-scripts.json']
+```
+
+I didn't propose this because it doesn't behave like other npm config arrays.
+
+### Should `npm install <pkg>` in interactive mode prompt the user?
+
+When adding a new dependency that has install lifecycle scripts, npm could
+interactively ask whether to approve it, similar to how `apt` confirms disk
+usage or how Composer prompts for plugin permissions.
+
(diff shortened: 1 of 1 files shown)


wraithgar:
This problem space overlaps in a lot of ways with the new `allow-git` flag (which is soon to expand to other non-registry dependency types).

It is important that as we think about how to solve/configure for these use cases that it is done in a way that is applicable to these other approaches.

Script execution in npm has a fundamental problem: there is no distinction between your package.json scripts and the scripts in your dependencies.  Even the solution for `allow-git` which says "root" "all" or "none" doesn't help because there is no concept of "git actions in the package.json itself" baked in.

This has always been the first problem to solve, how do we say "I want my own scripts to run but not my dependencies'" and now "I want my top level dependencies' scripts to run".  The use case of allowing an individual package (at any depth or version) is also not addressed with this.

The "min-release-age" flag is now also coming up against this problem.  How do we override this in a secure way (i.e. does not invite misconfigurations that reduce security), and is not a wholly unique approach to all other exceptions we might write?

Not to overload this rfc with scope creep, but I do want to have these things in mind as we architect here.  Is there a paradigm or a pattern that will allow us to grow into these solutions without having to invent wholly new configs for each iteration?

To summarize, the end goal is to have a config approach for:

 - non-registry dependencies
 - min-release-age
 - script lifecycle events
 - others?
 
That allows us to
 - distinguish between package.json things, top level dependencies, and subdependencies
 - override in a consistent way that does not compromise security
 - iterate with our solutions (we don't have to ship the every iteration just to have the feature)

wraithgar:
`--allow-git` already exists, and the underlying packages already support `--allow-remote` `--allow-registry` `--allow-file` and `--allow-directory`, we just need to add those to npm itself.  We can consider those "prior art" that already exists with the pattern of those allowed values being `root` `all` or `none`.  Adding `package` into the mix for lifecycle scripts may be a logical place to start?

everett1992:
IMO `--allow-* root|all|none` is not granular enough, I want explicit approval for each exception. But I also see the value in shorthands like 'all my direct dependencies'. All direct seems okay for `allow-git` but not for `allow-install-scripts` because your direct git dependencies are explicitly spelled out in source, while your direct dependencies can unexpectedly gain install scripts when upgrading versions.

Adding `package:...` or `spec:...` options to match nodes at any depth of the tree seems like a good addition for --allow-git et al. and can work for a new `--allow-install-scripts`. 

The ability to share configurations across packages is important to me, because I plan to opt 200,000 packages into using an allowlist of install scripts. Without updating each package, or breaking their builds, I'll populate a list with known packages. Central configuration will reduce toil if a popular package adds legitimate postinstall scripts. I don't know how we would merge option into `--allow-*` arguments.

---

What if we reuse npm query, or a subset, and have policies that allow the matched dependency to be resolved or installed?

```jsonc
{
   // No git dependencies can be resolved or installed
  ":type(git)": { "allowed": false },
  
  // better-sqlite3 11.x may run install scripts
  "#better-sqlite3@^11.0.0": { "install-scripts": false },
  
  // Use latest release of @scope packages.
  "[name^='@scope/']": { "min-release-age": 0 },
  
  // Package is allowed to declare dev git dependencies
  ":root > .dev:not(.prod):type(git)": { "allowed": false },
}
```

There's a lot I like about this, but it probably fails to "not invite misconfigurations that reduce security"

wraithgar:
> What if we reuse npm query, or a subset, and have policies that allow the matched dependency to be resolved or installed?

We have discussed this a lot in the past but never implemented it anywhere.  It does seem like the logical solution here.

wraithgar:
Where would this policy be configured?  The cli itself is often where it is frustrating to construct larger query strings.

wraithgar:
Is there any value in distinguishing WHAT lifecycle events we want to allow? This seems like a bad idea since it may give a false sense of security (any of those scripts can have security issues) and is another layer of code we don't need to write if we don't want it.

everett1992:
> Is there any value in distinguishing WHAT lifecycle events we want to allow? This seems like a bad idea...

I agree, I don't think there's a difference between prepare or postinstall. Maybe distinguish between postinstall and implicit node-gyp?

I also don't  think there's any point to check the scripts _contents_, like approve esbuild to run `./install.sh`. There's too much possible indirection to say if its 'the same' script.

everett1992:
>Where would this policy be configured?

New file(s) configured by a cli flag. Multiple files allows a per-project config and shared global config.

wraithgar:
npm already has a config file, adding a second file for npm to look at seems like a bad pattern.

everett1992:
Whether it's a new file or not isn't important to me, but the ability to have a shared config that is extended by a package is. I didn't see precedence for that in .npmrc, it looked like project npmrc always overwrites global.

wraithgar:
Yes, there is no way to "merge" configs across sources.  That may be something we have to give up here.

wraithgar:
Does the query language approach allow us to distinguish between the scripts in my own package.json and those of my dependencies?  This is one of the core problems to solve to even be able to block install scripts in the first place without blocking the root scripts.

"root" in the context of "allow-git" means "git references I have in my package.json" so "root" would have to mean "my own local lifecycle scripts" for an "allow-scripts" flag.  Unfortunately that leaves out "lifecycle scripts for top level dependencies" which is closer to what the analog means in "allow-git"

Trying to have something like "root" "all" "none" or "dependency selector" doesn't work because the dependency selector can't include "root".  We need something that says "root and these limited dependencies" OR we need an extra config that works in addition that lets an 'allow-scripts' flag to limit to root/all/none and then also open up extra permissions through a dependency selector.

`--allow-scripts=root`
`--allow-scripts-selector='.prod:root > :not(:outdated)'`

The potential for confusion here is that the selector adds to the list.  If I did `--allow-scripts=all` then the selector would be meaningless.

everett1992:
I wrote in the RFC that root and workspace's scripts are _always_ allowed. I can't imagine a use case for skipping the repo's own scripts, but running other package's scripts. `--ignore-scripts` would still exist as an escape hatch to install dependencies without running any scripts.

>  the dependency selector can't include "root"

Is that right? `:root` returns only the root package.

wraithgar:
> I wrote in the RFC that root and workspace's scripts are always allowed. I can't imagine a use case for skipping the repo's own scripts

The way the feature currently works is a very very helpful ux affordance.  If I run `npm test --ignore-scripts` it means I want to run the tests without the `posttest` linting script that happens normally.  I'd hate to break this functionality.

wraithgar:
> Is that right? :root returns only the root package.


Yes. Perhaps this is just me not fully understanding css selectors.  How would I query "the root package itself AND its direct dependencies"?


```sh
~/D/s/bundled-test $ npm pkg get dependencies
{
  "semver": "7.5.0"
}
$ npm query ':root'|npx json -a name
gar-test
$ npm query ':root .prod'|npx json -a name
lru-cache
semver
yallist
$ npm query ':root > .prod'|npx json -a name
semver
```

wraithgar:
Aaand it was just me being rusty

```sh
$ npm query ':root, :root > .prod'|npx json -a name
gar-test
semver
```

Selectors totally solve the entire problem space!

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+[bikeshedding](#unresolved-questions-and-bikeshedding) for how this could work.
+
+## Detailed Explanation
+
+### Enforcement mode
+
+A new `.npmrc` config value controls the feature:
+
+```ini
+install-script-policy=off|ignore|error
```
wraithgar: There are MANY more lifecycle scripts than just install.  While we have decided not to filter based on script type, it's probably worth naming the config something less specific to just the install scripts.  `--ignore-scripts` is the currentl config that turns these totally on or off, and "lifecycle script" is how they're referred to in the documentation.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+  package will not be listed as an error in `error` mode. This is useful for
+  packages whose scripts are non-essential (telemetry, sponsorship messages,
+  optional native compilation with a JS fallback, etc.).
+
+### Allowlist sources
+
+We need allowlist entries to come from at least two places because npmrc's
+config model flattens values — a project-level `.npmrc` replaces a global-level
+value rather than merging with it. If the allowlist lived only in `.npmrc`, a
+project that sets `install-script-allowlist=project-scripts.json` would silently
```
wraithgar: Does this approach still make sense if we are using a query selector?   We only have a few config items that point to files and we usually don't let those paths traverse outside of the project itself.  Having a "user level" config with a user-defined path like this seems unwise from a security standpoint.  I think we may just want to have these as bare config strings and folks can solve the "shared config" problem in the ways they already are for other config concerns.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+This two-source approach is a concrete proposal, but we're open to better ideas
+for achieving the same goal: a project-level list that coexists with a
+centrally-managed list. If there's a cleaner way to get mergeable lists out of
+npmrc, or a different file format that works better, we'd welcome that.
+
+### Interaction with `--ignore-scripts`
+
+`--ignore-scripts` continues to work as today — it disables all scripts
+unconditionally, regardless of the allowlist. The allowlist only applies when
+`install-script-policy` is `ignore` or `error` and `--ignore-scripts` is not
```
wraithgar: It's good to call this out and keep it.  It largely solves the UX issue I was bringing up about "running tests without posttest" etc.  This catchall will be what folks can use in those non-install cases.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+set.
+
+### Scope
+
+This feature applies to the install lifecycle scripts that run during
+`npm install` for dependencies:
```
wraithgar: This should apply to all lifecycle scripts.  Let's not limit here.  Exceptions usually have more surprises than rules that apply to everything.  It also future-proofs us against new scripts that may happen to come up.

wraithgar:
I left some more specific commentary inline.  I hope you understand where I'm coming from with my suggestion that this isn't limited to just install scripts.  This is admittedly coming from my own approach to having thought about this in the past, and having always included ALL lifecycle scripts in this problem space.

This is by no means a line in the sand.  If the community says "no this should only be for lifecycle scripts in my dependencies, my own top level scripts are outside the scope of this" I'd be ok with that.  My only pushback is that there is no way to disable my own package's install/postinstall but allow my dependencies' to still run.  The use cases of disabling project level scripts and dependency scripts are very separate, and the current `--ignore-scripts` does not have any granularity.  I think that granularity is important here.

accepted/0000-install-script-allowlist.md
everett1992: I wrote the RFC with 'scripts defined by dependencies' in mind, my goal is to limit the exposure to running arbitrary code when installing dependencies. That's why I didn't account for scripts in the root, or lifecycle scripts run on test/start ect non-install commands. 

It makes sense to pick a more generic name if the setting can apply to the root package and it's other hooks. Could we repurpose the `ignore-scripts` option to accept query filters and apply the policies? Or `allow-scripts` to be consistent with allow-git and ignore-scripts?

ljharb:
> no this should only be for lifecycle scripts in my dependencies, my own top level scripts are outside the scope of this

I think that's the entire point. If you don't want to run lifecycle scripts in your own project you can just delete them.

accepted/0000-install-script-allowlist.md
everett1992: Do you have examples of how people solve shared configs problems in npm?
pnpm has [config dependencies](https://pnpm.io/config-dependencies) as a general way to share setting values.

I'll be able to populate a default allowlist in a global .npmrc, but package's won't be able to extend that, any local .npmrc would overwrite.

ljharb: It hasn't been much of a problem, so it's usually just symlinked npmrc files, or a separate script to set config values.

accepted/0000-install-script-allowlist.md
wraithgar: Mixing types in command line flags is a HUGE source of bugs and tech debt. I'd avoid this altogether if possible.  If you want a fun exercise go look at `--browser`

wraithgar:
> If you don't want to run lifecycle scripts in your own project you can just delete them.


I run `npm test --ignore-scripts` dozens of times per day because I'm developing tests, I don't need the linting to run.

ljharb:
Sure, that makes sense for `--ignore-scripts` - I thought this RFC was for a *distinct* option that disables install scripts for all dependencies.

wraithgar:
> I thought this RFC was for a distinct option that disables install scripts for all dependencies.

Yep! It is.  I am just trying to think through all the uses cases here and find the edges.  I think we have found the edge here and now we know that "ignore-scripts" is not going anywhere.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+### Allowlist format
+
+The allowlist is a JSON object with a single top-level field:
+
+```json
+{
+  "approvedInstallScripts": {
+    "node-gyp-build": "allowed",
+    "esbuild": "allowed",
+    "esbuild@>=0.25.0": "allowed",
```
JamieMagee: Allowing `"esbuild@>=0.25.0": "allowed"` means a future compromised `esbuild@0.27.0` is auto-approved. I think think this is incompatible with the goal you stated above "any _new_ unapproved script is caught immediately."

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+   }
+   ```
+
+All sources are merged (unioned). If the same key appears in multiple sources,
+the most permissive value wins (`"allowed"` > `"ignored"`). The goal is to
+reduce toil — an organization can maintain one shared list and have every
+project inherit it automatically, while individual projects can still add their
+own entries. Whether a project should also be able to _override_ a shared entry
+(e.g. downgrade `"allowed"` to `"ignored"`) is an open question, but not a hard
+requirement either way.
```
JamieMagee: I think yhis is where the merge rule and the central-management motivation collide. The paragraph admits it's open, but I don't think it can be. If the project list can override the org list upward, the org list isn't enforcing anything.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+When adding a new dependency that has install lifecycle scripts, npm could
+interactively ask whether to approve it, similar to how `apt` confirms disk
+usage or how Composer prompts for plugin permissions.
+
+### Should there be an `npm approve-scripts` command?
+
+A helper command that scans `node_modules` (or the package tree) and
+generates/updates the allowlist would make adoption much easier. Something like
+`npm approve-scripts --init` to populate the allowlist with all
+currently-installed packages that have scripts.
```
JamieMagee: Without this adoption requires manual copy-paste from error output. IMO this is essential to help with adoption.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+- `off` (default): Current behavior. All install lifecycle scripts run
+  unconditionally.
+- `ignore`: Unapproved scripts are silently skipped. At the end of
+  `node_modules` construction, npm prints a summary of every package whose
+  scripts were skipped.
+- `error`: Unapproved scripts are skipped. At the end of `node_modules`
+  construction, npm exits non-zero and prints the list of unapproved packages
+  that had scripts.
```
JamieMagee: What if `foo@1.0.0` is approved but a transitive pulls in `foo@2.0.0` too? Per-name or per-(name, version)?

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+    "sqlite3@4.x": "allowed",
+    "husky": "ignored",
+    "sponsorware-nag": "ignored"
+  }
+}
+```
+
+Keys are `<package-name>` or `<package-name>@<semver-range>`. A bare package
+name (no `@range`) is equivalent to `@*` and matches all versions. Semver ranges
+in this list match pre-release versions.
```
JamieMagee: The pre-release rule says ranges match prereleases. That's the opposite of what `package.json` ranges do by default. `^1.0.0` in a dep won't pull in `1.1.0-beta.1`, but the same string in the allowlist would approve it. Worth aligning with the default so allowlist entries match the same versions a `package.json` range would.

Also: git deps don't have a trustworthy name (the `package.json` name is whatever the fork author put there). How are those keyed in the allowlist?

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+- Implicit `node-gyp rebuild` (when a `binding.gyp` is detected and no `install`
+  script is defined)
+- `prepare` scripts of git dependencies (which run during install to build the
+  package from source)
+
+It does not affect:
+
+- Scripts in the root project and workspaces' `package.json` (the package being
+  developed). It's specifically for installed dependencies.
+- `npm run` / `npm test` / `npm start` etc.
```
JamieMagee: What about `npm rebuild` and `npx`? Both run scripts. Also: is an explicit `"install": "node-gyp rebuild"` gated the same as the implicit case?

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+generates/updates the allowlist would make adoption much easier. Something like
+`npm approve-scripts --init` to populate the allowlist with all
+currently-installed packages that have scripts.
+
+### Interaction with workspaces
+
+In a monorepo, should each workspace have its own `approvedInstallScripts` in
+its `package.json`, or should only the root's list apply? The external-file
+approach naturally handles this (point all workspaces at the same file), but the
+`package.json` story needs clarification.
```
JamieMagee: I'd like to see this resolved before merge. Hoisted `node_modules` plus per-workspace `approvedInstallScripts` is going to behave nondeterministically depending on hoist order. The external-file workaround pushes the problem to users instead of fixing it.

accepted/0000-install-script-allowlist.md
```diff
@@ -0,0 +1,274 @@
...
+name (no `@range`) is equivalent to `@*` and matches all versions. Semver ranges
+in this list match pre-release versions.
+
+Values:
+
+- `"allowed"` — the package's install lifecycle scripts will run.
+- `"ignored"` — the package's install lifecycle scripts will not run, and the
+  package will not be listed as an error in `error` mode. This is useful for
+  packages whose scripts are non-essential (telemetry, sponsorship messages,
+  optional native compilation with a JS fallback, etc.).
```
JamieMagee: What does `"ignored"` actually buy you over just leaving a package off the list? The only delta I can find is "doesn't show up in the error report," and that's uncomfortably close to "hides a warning you might want." If it's really just for sponsorship/telemetry noise, maybe rename it `"silenced"` so nobody confuses it for a security primitive?

JamieMagee:
A few things I'd want addressed, separate from the selector-vs-config-shape question:

  1. "Most-permissive wins" inverts the central management story. If the org list says `"sketchy-pkg": "ignored"` and a project's `package.json` says `"sketchy-pkg": "allowed"`, the project just escalated past org policy. Either flip to most-restrictive-wins, or give shared entries a way to mark themselves locked. The current rule makes the shared list a floor, when the use case described in the motivation needs it to be a ceiling.
  2. `"esbuild@>=0.25.0": "allowed"` approves every future 0.x release, including the compromised one. The motivation says the goal is to catch new unapproved scripts, but ranges make that impossible by construction. I'd push for exact versions only, or for the allowlist to bind to lockfile-pinned (name, version, integrity) tuples. Otherwise this leans on the honor system of the registry.
  3. The lockfile already has integrity hashes for every dep. Approving "what's in the lockfile right now" feels like a much better primitive than approving names. It makes the diff visible in code review when something changes. Worth at least an implementation subsection.

accepted/0000-install-script-allowlist.md
everett1992: I'll tone the wording down. I think introducing a closed set of packages is a significant improvement over the status quo, and managing a list of allowed package versions is a good balance of convenience vs security compared to managing a list of package versions.

The RFC gives package owners  the flexibility to allowlist single package versions, or ranges.

accepted/0000-install-script-allowlist.md
everett1992: I'd rather defer to a followup RFC / pull request, I don't want to get bogged down in designing the UX of that command.

accepted/0000-install-script-allowlist.md
everett1992: Should be the name@version or whole spec

accepted/0000-install-script-allowlist.md
everett1992: I including this with shared configs in mind, because it's hard to write a range that matches multiple pre-releases so the shared config would have to enumerate many options. Looks like there may be some [precedent for using includePrerelease](https://github.com/npm/cli/blob/27567ab27bef8af303bde0b2d7da0386da182b81/workspaces/arborist/lib/vuln.js#L15).

I'm happy to loose this.

I think changing to use `npm query` addresses git dependencies - they would allowlisted by repository matcher.

accepted/0000-install-script-allowlist.md
everett1992: rebuild, npx, update, anything that runs lifecycle script of dependencies would be controlled by this setting.
explicit and implicit scripts are treated the same.

accepted/0000-install-script-allowlist.md
everett1992: It covers cases like core-js that function without their scripts, or esbuild that only needs scripts as a fallback. I'm not attached to any of the config names I proposed, ignored, silenced or do-not-run seem fine.

karlhorky:
> @JamieMagee in [comment 4270774870](https://github.com/npm/rfcs/pull/861#issuecomment-4270774870): 2\. `"esbuild@>=0.25.0": "allowed"` approves every future 0.x release, including the compromised one. The motivation says the goal is to catch new unapproved scripts, but ranges make that impossible by construction. I'd push for exact versions only, or for the allowlist to bind to lockfile-pinned (name, version, integrity) tuples. Otherwise this leans on the honor system of the registry.

As reference, over in pnpm, the `minimumReleaseAgeExclude` setting was first introduced with package names only, and then pattern support and then finally version numbers:

- 9 Sep 2025: original `minimumReleaseAgeExclude` impl, with package names only: https://github.com/pnpm/pnpm/pull/9957
- 16 Sep 2025: pattern support: https://github.com/pnpm/pnpm/pull/9984
- 9 Oct 2025: version number support: https://github.com/pnpm/pnpm/pull/10059

Not saying that this is the right way (maybe only allowing exact version numbers would constrain the API to be more secure by default) - just offering some prior art with discussion around it.

itsnotmike:
Commenting so I receive updates on this thread. Apologies if that's not the intended purpose. 

In case it hasn't been raised, we are blocked from implementing min-release-age until we can exclude internally developed packages. 
eg. npm config set min-release-age-exempt @internal
or: npm config set @internal:min-release-age 'empty string/exempt/ignored etc.'

Thanks for the work on this team, mods please delete if this comment doesn't belong here.

everett1992:
Closing this RFC in favor of https://github.com/npm/rfcs/pull/868

Issue #94, opened 2020-01-28
[FEATURE] Allow customizing packages on install


begriffs:
# What / Why

Currently running `npm install` fetches and installs a package as-is. However it would be more flexible to be able to specify a patch to apply at installation time to customize a package. One use case is to modify the package source to remove donation banners in install logs.

This is complementary to -- and more general than -- #80.

## When

At install time, with the presence of a new flag.

## How

Add a `--patch /path/to/my.patch` option to `npm install`. NPM would then fetch the source as usual,  run `patch -p0 < /path/to/my.patch` to apply the customization, and continue with the normal installation process.

### Example

Suppose you want to install [core-js](https://github.com/zloirock/core-js) without its donation banner breaking your continuous integration. You can already use `npm install --loglevel silent`, but this will silence other potentially useful messages.

Instead, the user could create a patch file, e.g. `nobanner.patch`:

```patch
--- packages/core-js/postinstall.js
+++ packages/core-js/postinstall.js
@@ -52,5 +52,3 @@
   // eslint-disable-next-line no-console,no-control-regex
   console.log(COLOR ? BANNER : BANNER.replace(/\u001B\[\d+m/g, ''));
 }
-
-if (isBannerRequired()) showBanner();
```

Then, using the proposed feature, run:

```bash
npm install --patch nobanner.patch core-js
```

isaacs:
If this is really about donation request banners, they're being phased out in favor of the `funding` package.json stanza and `npm fund` command anyway.  So I'd like to see another use case, at least, since that one won't matter soon enough.

What happens when the patch doesn't apply?

This feels like one hell of a footgun.  I have deep misgivings about supporting it.

begriffs:
> If this is really about donation request banners, they're being phased
> out in favor of the `funding` package.json stanza and `npm fund`
> command anyway.  So I'd like to see another use case, at least, since
> that one won't matter soon enough.

There is no other concrete example I know of. Basically I saw some
Github issues that were full of drama about the banners, and wondered to
myself if there was a technical solution that would allow project
authors to keep the banners if they wanted while letting package
consumers customize it if they cared enough.

The patching idea seemed a simple mechanism that is open-ended enough to
possibly apply to other things.

> What happens when the patch doesn't apply?

On failure the patch utility exits with a non-zero code, and npm install
could fail with an error.

> This feels like one hell of a footgun.  I have deep misgivings about
> supporting it.

After thinking about the idea again today, I realized that it may not
[...]

PR #862, opened 2026-05-03
RFC: Native Dependency Patching

manzoorwanijk:
## Summary

Adds first-class, install-time patching of installed dependencies to the npm CLI, on parity with `pnpm patch`, `yarn patch`, and `bun patch`. Introduces a new `npm patch` command with subcommands `add` / `commit` / `update` / `ls` / `rm` (with `npm patch <pkg>` as shorthand for `npm patch add <pkg>`), a `patchedDependencies` field in `package.json`, and a `patched.{path,integrity}` record in `package-lock.json` (`lockfileVersion: 4`). Patches apply during Arborist's reify step, uniformly across every supported `install-strategy` (`hoisted`, `nested`, `shallow`, `linked`).

## Why now

The third-party `patch-package` is currently the only path to dependency patching for npm users, and it is structurally limited:

- **Silently disabled by `--ignore-scripts`.** `patch-package` runs as a `postinstall` script. In environments that disable lifecycle scripts — increasingly common in hardened CI and after recent supply-chain incidents like the [`Shai-Hulud` worm (Sept/Nov 2025)](https://www.reversinglabs.com/blog/shai-hulud-worm-npm) — declared patches simply do not apply, with no error and no warning. Production code can be installed missing fixes that are committed in the project.
- **Broken with workspaces** ([ds300/patch-package#277](https://github.com/ds300/patch-package/issues/277)).
- **Broken with `install-strategy=linked`** ([ds300/patch-package#595](https://github.com/ds300/patch-package/issues/595)).
- **Unmaintained.**

The headline outcome: reproducible, source-controlled dependency hotfixes that survive `--ignore-scripts` and work across every npm install strategy and across workspaces.

## Relationship to npm/rfcs#94

This RFC is a direct response to [#94](https://github.com/npm/rfcs/issues/94) (closed in 2020 as a footgun). The 2020 proposal was an ad-hoc `npm install --patch foo.patch` flag with no manifest record, no lockfile linkage, no transitive-dep support, and no failure-mode story; @isaacs's footgun objection was correct for that shape. This RFC is structured the opposite way — explicit manifest, lockfile-hashed, fail-loud-by-default, version-gated, publish-isolated. A row-by-row response is in the RFC's _Prior Art → npm/rfcs#94_ section.

See the RFC for the full design, alternatives considered, implementation plan, tests, and unresolved questions.

---

> **Disclosure**: [Claude Code](https://claude.com/claude-code) was used to draft this PR description and the initial version of this RFC and to iterate on it during review.


james-pre:
Hey @manzoorwanijk,

There are a few questions I have:

1. How are multiple patches for a single package handled? Let's say I depend on `duck-quack-js` and want to apply a patch to add a new feature from a PR and another patch to fix a security issue.

2. How are patches required for dependents handled? Take for example this dependency chain:
- `@ducks/quack` patches `duck-quack-js` to add a feature PR that is required at runtime
- `@ducks/mallard` depends on `@ducks/quack`. `npm install @ducks/quack` needs to result in the patch being applied otherwise we get runtime errors.

3. Why not wrap everything under a single `npm patch` command (`npm patch add`, `... ls`, `... rm`, `... commit`)?

manzoorwanijk:
Thanks @james-pre - three good questions.

**1. Multiple patches for the same package.**

Punted in v1: see *Unresolved Questions and Bikeshedding* item 6 ("Stacking patches"). The RFC matches at most one patch per resolved node, and stacking is left as an additive extension (selector value becomes an array). The reason for deferring is that with `fuzz=0` (also v1 default), stack order is load-bearing and composition rules need their own design pass - does order 1 → 2 produce the same tree as 2 → 1? what does `npm patch-remove` do with a stack? etc.

Practical workaround for now: merge the two diffs into a single `.patch` file (`cat a.patch b.patch | <git-apply-style-merge>`, or hand-merge). Not pretty, but unblocks the case until the stacking design lands.

**2. Patches needed by dependants / transitive consumers.**

This is intentionally not supported, and the constraint is by design rather than oversight. From the RFC:

- `patchedDependencies` is honoured **only in the root `package.json`** of the consuming project.
- `npm publish` / `npm pack` **strip** `patchedDependencies` from the published manifest and exclude the `<patches-dir>/` from the tarball.

So in your example, if `@ducks/quack` publishes a `patchedDependencies` entry for `duck-quack-js`, it never travels through the registry, and consumers of `@ducks/mallard` will not have the patch applied. This is deliberate: a published package being able to silently mutate its consumers' transitive dependency trees is a supply-chain abuse vector we shouldn't introduce. (It would be a strictly bigger lever than `postinstall` scripts - applied silently, inside the install pipeline, with no `--ignore-scripts`-style escape hatch.)

The right tool for "I want consumers of `@ducks/quack` to get a patched `duck-quack-js`" is *Alternative 2* in the RFC: `@ducks/quack` declares a normal dependency on a published `@ducks/duck-quack-js` fork (or uses an `overrides` entry that the consumer also opts into in their root). That makes the patched code a real registry artifact with a normal version, auditable in the lockfile.

**3. Why not consolidate under `npm patch <subcommand>`?**

Fair point - the current mix (`npm patch`, `npm patch-commit`, `npm patch-remove`, `npm patch ls`) is inconsistent, and `npm patch add/ls/rm/commit` would be more uniform and more idiomatic for npm (compare `npm pkg get/set/delete/fix`, `npm cache add/clean/verify/ls`, `npm team create/destroy/add/rm/ls`). I followed pnpm/yarn naming originally for prior-art familiarity, but you're right that npm's own convention is the better fit.

I'd propose:

- `npm patch <pkg>` - shorthand for `npm patch add <pkg>` (the most common entry point, kept short).
- `npm patch add <pkg>` - start an edit session.
- `npm patch commit <edit-dir>` - finalise (no dash; aligns with `npm pkg set`, etc).
- `npm patch ls` - list registered patches.
- `npm patch rm <pkg>[@<version>]` - remove a patch.

Happy to update the RFC to use this shape if there's broader agreement.

manzoorwanijk:
The rename is done — `npm patch add` / `commit` / `ls` / `rm`, with `npm patch <pkg>` kept as shorthand for `npm patch add <pkg>`. RFC and PR description updated.

manzoorwanijk:
@owlstronaut, while this RFC is still in review, is it fine to create a draft PR of the implementation in the CLI repo for demonstration?

owlstronaut:
> @owlstronaut, while this RFC is still in review, is it fine to create a draft PR of the implementation in the CLI repo for demonstration?

Absolutely. Love the idea

manzoorwanijk:
Here is the first PR (draft) https://github.com/npm/cli/pull/9439. The PR is feature complete, and the only part left is to support patching for aliases via `npm:`, which I will do in a follow-up. Please let me know if I should mark the PR as ready for review.

ehoogeveen-medweb:
Under this proposal, what would be the practical workflow for updating an existing patch that still applies to a new version?

With [patch-package](https://github.com/ds300/patch-package), it still attempts to apply patches to newer versions (but prints a warning), so you just run `npx patch-package package-name` to generate a new patch from the already modified files, then delete the old patch. Obviously things get more complicated if there are conflicts.

With the proposed native patching, I think you'd have to do something like this:
```bash
npm install <pkg>@new-version --allow-unused-patches

npm patch add <pkg>@new-version

pushd /path/printed/by/npm/patch/add
git apply --ignore-whitespace /path/to/project/patches/<pkg>@old-version.patch
popd

npm patch commit /path/printed/by/npm/patch/add

npm patch rm <pkg>@old-version
```
...and then commit the changes.

It would be great if there was a better way like an explicit `npm patch update <pkg>`.

manzoorwanijk:
> Under this proposal, what would be the practical workflow for updating an existing patch that still applies to a new version?

`npm patch update` is a great idea. Thanks. Let me add that to the proposal.

manzoorwanijk:
Added `npm patch update <pkg>[@<old-version>] [--to <new-version>]` to the RFC: resolves the existing entry, extracts a fresh copy of the new version, attempts a 3-way `git apply` of the existing patch, and either re-commits cleanly (renaming the selector, deleting the old patch file subject to the shared-reference rule) or leaves the conflict markers in the edit dir for you to resolve and re-run `npm patch commit`. Single command for the happy path, same finalisation flow for the conflict path. Runnable without a successful prior install - designed so it can recover from a fuzz-drift-aborted install.

accepted/0000-native-dependency-patching.md
```diff
@@ -0,0 +1,544 @@
...
+      "path": "patches/lodash@4.17.21.patch",
+      "integrity": "sha512-..."
+    }
+  }
+}
+```
+
+`patched.path` is the relative path from the project root. `patched.integrity` is the SSRI hash of the patch file's contents.
+
+This bumps `lockfileVersion` from `3` to `4` ([`workspaces/arborist/lib/shrinkwrap.js`](https://github.com/npm/cli/blob/latest/workspaces/arborist/lib/shrinkwrap.js) defines the current default and the parse path supports forward-version detection). The bump is the load-bearing safety mechanism: an older npm reading a v4 lockfile with any `patched` record **must error and abort the install**. A warn-only fallback is explicitly rejected — it would let an older CLI silently install unpatched code from a project that declares patches, which is exactly the regression this RFC is built to prevent. A v4 lockfile with **no** `patched` records anywhere remains safe to install under older clients; the abort only kicks in when a patch is present.
```
owlstronaut: There isn't a way to retroactively get this error for older npm, right? I guess it is load-bearing for people on the version of npm >x with the safeguard?

accepted/0000-native-dependency-patching.md
```diff
@@ -0,0 +1,544 @@
...
+
+## Unresolved Questions and Bikeshedding
+
+1. **Patches directory path**. `patches/` matches `patch-package`, pnpm, and bun. Is there an objection to occupying that name at the project root? Alternative: `.npm/patches/`.
+2. **Interaction with `overrides`**. Patches apply to the **resolved** version after override resolution. If `overrides` swaps `foo@1` → `foo@2`, a `patchedDependencies["foo@1"]` entry becomes unused and errors under default settings — which seems correct, but confirms that overrides + patches require coordinated edits. Is the "patch the resolved version" rule the right composition order in every case, or should range selectors be evaluated against the pre-override identity?
+3. **`npm audit fix` + patches**. If a patch is the only thing keeping a transitive dep secure, should `npm audit fix` refuse to upgrade away from the patched version, or upgrade and require the user to re-author the patch? Strawman: refuse, with `--force` to upgrade.
+4. **Exclude `package.json` from the diff by default?** `patch-package` does this because version bumps make patches non-portable. Carry the default forward, with `--include-package-json` to override.
```
owlstronaut: currently the implementation diffs package.json but Arborist resolves the pre-patch manifest, so patched stuff ends up on disk without being honored. How do we avoid silent partial application?

accepted/0000-native-dependency-patching.md
```diff
@@ -0,0 +1,544 @@
...
+- **Deprecated packages**: deprecation is a registry warning, not a resolution change. Patches on deprecated packages work normally; `npm ls` and `npm audit` show both the deprecation warning and the `[patched: ...]` annotation. Patching is in fact one of the better responses to a transitive deprecation that has no published successor yet.
+
+### `npm publish` and `npm pack`
+
+`patchedDependencies` is consumer-side state. A package being published to a registry must never carry it: doing so would either silently leak project-local edits to consumers (if the field were honoured) or look like a configuration error to anyone reading the published tarball.
+
+Therefore:
+
+- `npm publish` and `npm pack` **strip** `patchedDependencies` from the manifest written into the tarball. This is a stronger guarantee than `devDependencies` (which is left in the tarball but ignored by consumers) because `patchedDependencies` has install-time effects and would actively confuse consumers if it survived publish.
```
owlstronaut, line 370: the curent implementation just strips the registy side metadata today, so `npm pack --dry-run` still ships the patches. How will the tarball-side strip/exclusion get delivered?

accepted/0000-native-dependency-patching.md
```diff
@@ -0,0 +1,544 @@
...
+
+4. Cleans up the temp edit directory (unless `--keep-edit-dir`).
+5. Updates `package-lock.json` to record the patch's `sha512` against every node that the patch resolves to (see "Lockfile" below).
+
+Flags:
+
+- `--patches-dir <dir>`: override the destination directory for this invocation.
+- `--keep-edit-dir`: do not remove the edit directory after committing.
+
+#### `npm patch update <pkg>[@<old-version>] [--to <new-version>]`
```
owlstronaut, line 114: i don't think this can be deferred to a follow-up or we lose the 'fail loudly' bit until it is added (per the follow-up work in your pr)

accepted/0000-native-dependency-patching.md
manzoorwanijk: Correct, there's no way to make an already-released npm error beyond what it already does. The abort isn't new code in older clients; it's the existing `lockfileVersion > maxLockfileVersion` guard in `shrinkwrap.js`, which throws `Unsupported lockfileVersion 4 … Please upgrade npm` and refuses to install. Since every npm before this feature ships `maxLockfileVersion = 3`, it hits that guard on a v4 lockfile and stops.

So yes, it's load-bearing for the band of npm versions that have the lockfile-version gate (npm 7 → the release just before v4 support); those abort cleanly. npm 6 and older predate that gate and wouldn't be protected, but they're long EOL. The v4 bump is deliberately leaning on that pre-existing guard rather than inventing a new one.

accepted/0000-native-dependency-patching.md
manzoorwanijk: Nice catch! Thanks.

Fix for v1: **exclude `package.json` from the diff by default** (patch-package's default, resolves https://github.com/npm/rfcs/pull/862#discussion_r3356964205), with a warning when it's the only edit. That removes the vector - a patch can't touch resolution-affecting fields. Honoring manifest patches would mean applying them before ideal-tree resolution; out of scope for v1. I'll update the implementation and RFC.

accepted/0000-native-dependency-patching.md
manzoorwanijk: Right, today only the registry manifest is stripped (libnpmpublish); the tarball's `package.json` still carries it. I'll move the strip into the pack path (libnpmpack) so `npm pack` output drops the field too.

Open question: Should we also exclude the `patches/` dir from the tarball? It's dead weight for consumers. I lean toward yes.

accepted/0000-native-dependency-patching.md
manzoorwanijk: Fair enough. I'll pull `update` into v1. Fail-loud-on-drift only works with a first-class recovery; without `update` the alternatives are manual `rm` > `add` > re-edit or `--ignore-patch-failures` (installs unpatched, the thing fail-loud prevents). Moving it out of the follow-up list and into this PR.

accepted/0000-native-dependency-patching.md
manzoorwanijk: > How will the tarball-side strip/exclusion get delivered?

Today only the registry packument manifest is stripped (`libnpmpublish`). The tarball still ships the field and `patches/` because its `package.json` and file list come from **pacote** (`tar.c` over raw on-disk files) and **npm-packlist** - neither editable from the CLI. So delivering this needs:

- **pacote**: strip `patchedDependencies` from the `package.json` it writes for a `directory` pack;
- **npm-packlist**: exclude the resolved patches dir (passed through so it honors `--patches-dir`), plus the "force-added via `files`" warning.

**Question:** OK to land these as separate pacote + npm-packlist PRs (then bump deps here), with this PR keeping only the registry-manifest strip for now? Or is there a pack-time manifest seam you'd prefer we hook? We'd rather not re-implement file-selection/manifest handling in the CLI.

owlstronaut: yes, that's the right path forward. The repo changes will need to be a breaking change/new major. Pacote's engine bump is still pending, so that one was getting one anyway. So this will be an npm 12 feature.

I'll have to do the deps updates myself for npm 12 once it is possible, so this pr will stay open until all of the upstream dependencies are available to it.

accepted/0000-native-dependency-patching.md
owlstronaut: I'm having trouble finding the gate you say existed in npm 7+. Could you help point me to it?

manzoorwanijk: Sorry, I think I was confused. When we had to deal with such npm (v7) upgrade issues in Gutenberg (https://github.com/WordPress/gutenberg/issues/34801, https://github.com/WordPress/gutenberg/issues/46652), I thought there was some kind of gate that results in such errors for npm version mismatches.

I also handed those two Gutenberg issues to Claude to research in the context of this RFC, and I don't know why it gave me that false sense of confidence. 🤷 

So you are right; there's no retroactive error, and it's weaker than load-bearing. Released npm (v7+) has no `lockfileVersion` ceiling: it reads the version as-is and ignores unknown fields like `patched`, so an old client reading a v4 lockfile installs **unpatched** rather than erroring. The `> maxLockfileVersion` abort is new in this PR, so it only helps clients that ship the feature (and guards them against a future v5+).

So v4 doesn't make old clients fail *closed* - they fail *open* (silent unpatched install). I'll fix the RFC wording. A real fail-closed guarantee would need something released npm already hard-errors on (e.g. `engines`/`engine-strict`) - open to options.

accepted/0000-native-dependency-patching.md
manzoorwanijk: Perfect, that works.

manzoorwanijk: Here are the two upstream PRs:

- https://github.com/npm/pacote/pull/497
- https://github.com/npm/npm-packlist/pull/291

owlstronaut (approved):

saifktk5383 (approved):

Issue #860, opened 2026-04-02
[RRFC] Extend npm audit to detect supply chain compromise signals beyond CVEs


ranjithrajv:
## Motivation ("The Why")

The axios compromise on March 31st exposed a gap that `npm audit` can't close on its own. A threat actor hijacked the primary maintainer's npm account and published two backdoored versions within 40 minutes — each with a previously undeclared dependency running a RAT via post-install. There was no CVE at the time of publish. The advisory database had nothing to flag.

This isn't isolated. In 2025 alone, 45,777 CVEs were reported averaging 130 per day, and multiple npm supply chain attacks hit packages with billions of weekly downloads. A growing share of these attacks arrive before a CVE exists — through account takeovers, dependency injection, and post-install abuse.

### Example

A developer runs `npm install` after a routine `package.json` update. One dependency received a patch version bump overnight. The bump introduced a new post-install script and was published by an account that had never touched the package before. `npm audit` returns clean. The machine is compromised within seconds of install completing.

#### Current Behaviour

`npm audit` queries the advisory database for known CVEs against the resolved dependency tree. It returns clean if no CVEs exist, regardless of other signals.

#### Desired Behaviour

`npm audit` (or a new `npm audit --provenance` / `npm audit --supply-chain` flag) additionally checks:

- Whether the publish account for each installed version has previously published that package
- Whether new post-install or install scripts were introduced in the installed version vs. the prior one
- Whether new transitive dependencies were introduced that weren't present in the prior version
- Whether the installed version has a valid SLSA provenance attestation, especially for packages that previously carried one

These checks don't require a new database — npm already has publish history and provenance metadata. Surfacing them at audit time gives developers a signal before a CVE is ever filed.

### References

- [Microsoft Security Blog — Mitigating the Axios npm supply chain compromise](https://www.microsoft.com/en-us/security/blog/2026/04/01/mitigating-the-axios-npm-supply-chain-compromise/)
- [The Hacker News — Axios Supply Chain Attack Pushes Cross-Platform RAT](https://thehackernews.com/2026/03/axios-supply-chain-attack-pushes-cross.html)
- [Palo Alto Networks — Widespread npm Supply Chain Attack Puts Billions of Weekly Downloads at Risk](https://www.paloaltonetworks.com/blog/cloud-security/npm-supply-chain-attack/)
- Related: #739

ljharb:
How do you propose npm do such detections, considering that's a large value proposition of many SCA businesses?

[...]

PR #868, opened 2026-05-14
[RFC] Make install scripts opt-in

JamieMagee:
## Summary

Block dependency install scripts (`preinstall`, `install`, `postinstall`, and auto-detected `node-gyp` builds) by default during `npm install`. Projects opt in to running scripts for specific dependencies via a new `allowScripts` field in `package.json`. Two new CLI commands, `npm approve-scripts` and `npm deny-scripts`, help users build and maintain the allowlist.

npm is the only remaining major package manager that runs dependency install scripts by default. pnpm v10+, Yarn Berry, Bun, and Deno all block them.

## Why now

Recent attacks:

- **Shai-Hulud worm** (September 2025): self-replicating postinstall payload compromised 500+ npm packages by stealing maintainer tokens and republishing infected versions.
- **chalk, debug, and 17 other packages** (September 2025): phished maintainer account used to inject Web3 wallet-draining code into packages with over 2B combined weekly downloads. Delivered via `postinstall`.
- **Axios** (March 2026): hijacked lead maintainer published versions with a phantom dependency that existed only to trigger its `postinstall` hook, deploying a cross-platform RAT. The malicious package was never imported in Axios source code.

Install scripts run automatically the moment a package lands in the dependency tree. They don't require any `require()` or `import` from the application. A typo, a transitive dep change in a lockfile a reviewer didn't read, or a maintainer compromise becomes immediate code execution.

## Relationship to other RFCs

- Supersedes [RFC #488](https://github.com/npm/rfcs/pull/488) (the 2021 ancestor, 369 👍, rejected as too disruptive at the time). The phased migration plan in this RFC addresses that concern directly.
- Supersedes [RFC #861](https://github.com/npm/rfcs/pull/861) per author's note.

### DIFF
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -279,13 +279,17 @@ The two commands share implementation. Splitting them avoids the `!`-prefix synt
 
 ### Enforcement behavior
 
-When a dependency has install scripts and is not in the `allowScripts` allowlist:
+The policy is a tri-state per package:
 
-- The install continues by default. Scripts are skipped, and a warning is printed listing the blocked packages with a suggestion to run `npm approve-scripts`.
-- In strict mode (`strict-script-builds=true` in `.npmrc`), the install fails with an error instead.
-- The `--dangerously-allow-all-scripts` flag overrides the allowlist and runs all scripts.
+- `true`: install scripts run.
+- `false`: install scripts are silently skipped.
+- absent: install scripts run (Phase 1 default) and a post-install advisory warning lists the package so users can review it. In a future major release this default flips to skip-with-warning.
 
-The existing `--ignore-scripts` flag continues to work as before, disabling all scripts including root project scripts.
+In strict mode (`strict-allow-scripts=true`), the install fails before any scripts run if any dependency has install scripts that aren't covered by a `true` or `false` entry. The `--dangerously-allow-all-scripts` flag bypasses the policy and runs every install script.
+
+The `--allow-scripts` CLI flag is restricted to one-off and global contexts (`npm exec`, `npx`, `npm install -g`). Passing it during a project-scoped `npm install`, `ci`, `update`, or `rebuild` is an error: team-wide policy belongs in `package.json#allowScripts` or `.npmrc`, not in a command-line flag that gets copy-pasted into READMEs.
+
+The existing `--ignore-scripts` flag continues to work as before, disabling all scripts (including root project scripts) regardless of `allowScripts`.
 
 ### Affected commands
 
@@ -379,11 +383,11 @@ The primary enforcement point is in the `@npmcli/run-script` and `@npmcli/arbori
 
 New `.npmrc` settings:
 
-| Setting                         | Type                 | Default | Description                                                      |
-|---------------------------------|----------------------|---------|------------------------------------------------------------------|
-| `allow-scripts`                 | Comma-separated list | (empty) | Packages allowed to run install scripts (for global/npx context) |
-| `strict-script-builds`          | Boolean              | `false` | When `true`, blocked scripts cause install to fail               |
-| `dangerously-allow-all-scripts` | Boolean              | `false` | When `true`, all scripts run (escape hatch)                      |
+| Setting                         | Type                 | Default | Description                                                                                                                  |
+|---------------------------------|----------------------|---------|------------------------------------------------------------------------------------------------------------------------------|
+| `allow-scripts`                 | Comma-separated list | (empty) | Packages allowed to run install scripts. Valid only for `npm exec` / `npx` / `npm install -g`; rejected in project installs. |
+| `strict-allow-scripts`          | Boolean              | `false` | When `true`, packages with install scripts that aren't covered by `allowScripts` cause the install to fail before scripts run. |
+| `dangerously-allow-all-scripts` | Boolean              | `false` | When `true`, bypass `allowScripts` entirely and run every install script (escape hatch).                                     |
 
 ## Prior Art
 
@@ -406,20 +410,19 @@ New `.npmrc` settings:
 
 ## Migration Plan
 
-### Phase 1: Tooling and advisory warnings (next minor release)
+### Phase 1: Enforcement for opt-in entries; advisory for unlisted (next minor release)
 
 - Ship `npm approve-scripts` and `npm deny-scripts` (with `--pending` mode for previewing on `approve-scripts`).
-- Recognize the `allowScripts` field in `package.json`.
-- Print advisory warnings when dependency install scripts run that are not covered by an `allowScripts` field.
-- No change in default behavior: scripts still run.
+- Recognise the `allowScripts` field in `package.json`.
+- `true` entries run scripts; `false` entries silently skip them. Both do exactly what the user wrote.
+- Packages with install scripts that aren't covered by an `allowScripts` entry still run as before, with a post-install advisory warning. Nothing changes for anyone who hasn't opted in.
+- `strict-allow-scripts=true` and `--dangerously-allow-all-scripts` are wired up and enforce as described above.
+- `--allow-scripts` CLI flag is rejected in project-scoped installs; it remains available for `npm exec`, `npx`, and `npm install -g`.
 
 ### Phase 2: Default-deny (next major release)
 
-- Dependency install scripts are blocked by default.
-- Scripts for packages listed in `allowScripts` with `true` still run.
-- Blocked scripts produce a warning with remediation instructions.
-- `strict-script-builds=true` available for CI environments that want hard failures.
-- `--dangerously-allow-all-scripts` available as an escape hatch.
+- Packages with install scripts that aren't covered by `allowScripts` are skipped by default instead of running.
+- All other semantics from Phase 1 are unchanged.
 
 ### Phase 3: Ecosystem stabilization
 
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -1,3 +1,14 @@
+---
+title: Make install scripts opt-in
+number: null
+status: proposed
+created: null
+accepted_at: null
+implemented_at: null
+withdrawn_at: null
+implementation: null
+---
+
 # Make install scripts opt-in
 
 ## Summary
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -330,7 +330,9 @@ In a workspace (monorepo) context:
 
 ### Optional dependencies
 
-If a package in `optionalDependencies` has install scripts that are blocked, it is treated as a failed optional dependency installation. This is consistent with existing behavior where optional dependencies that fail to build are silently skipped.
+If a package in `optionalDependencies` has install scripts that are not allowed, npm still installs the package and skips only its scripts, exactly as it would for a regular dependency. The package is not treated as a failed optional dependency. Packages that use install scripts to optimize themselves but still function without them keep working when listed in `optionalDependencies`.
+
+The existing failed optional dependency behavior is unchanged. It applies when a script that is allowed to run exits non-zero, or when the package fails to extract. In those cases npm drops the optional dependency from the tree, the same as it does now. A script that the `allowScripts` policy skips never runs, so it cannot trigger that path.
 
 ### Bundled dependencies
 
--- a/accepted/0000-make-scripts-install-opt-in.md
+++ b/accepted/0000-make-scripts-install-opt-in.md
@@ -1,0 +1,439 @@
+# Make install scripts opt-in
+
+## Summary
+
+Dependency install scripts (`preinstall`, `install`, `postinstall`, and auto-detected `node-gyp` builds) should be blocked by default during `npm install`. Projects opt in to running scripts for specific dependencies by listing them in a new `allowScripts` field in `package.json`. An `npm approve-scripts` command helps users build and maintain this allowlist.
+
+This proposal aligns npm with pnpm (v10+), Yarn Berry, and Bun, all of which already block dependency install scripts by default.
+
+## Motivation
+
+### The security case
+
+Install scripts execute arbitrary shell commands the moment a package is installed. Unlike application code, which must be explicitly imported and run, install scripts fire automatically as a side effect of `npm install`.
+
+Attackers have exploited this repeatedly:
+
+- **event-stream (November 2018)**: A social engineering attack added a malicious `postinstall` dependency (`flatmap-stream`) targeting a specific Bitcoin wallet application.
+- **Lazarus Group campaigns (2024-2025)**: North Korean state-sponsored actors ran sustained campaigns publishing typosquatted packages (`is-buffer-validator`, `react-event-dependency`, and others) with `postinstall` scripts that deployed the BeaverTail malware and InvisibleFerret backdoor, stealing browser credentials and cryptocurrency wallet data. [Source](https://www.bleepingcomputer.com/news/security/north-korean-lazarus-hackers-infect-hundreds-via-npm-packages/)
+- **chalk, debug, and 17 other packages (September 2025)**: A phished maintainer account was used to inject Web3 wallet-draining code into 19 packages with over 2 billion combined weekly downloads. The payload was delivered via `postinstall` scripts in packages that had never needed install scripts before. [Source](https://www.ox.security/blog/npm-packages-compromised/)
+- **Shai-Hulud worm (September 2025)**: A self-replicating `postinstall` payload compromised 500+ npm packages by stealing maintainer tokens and automatically publishing infected versions of the victim's other packages. CERT/CC issued advisory [VU#534320](https://www.kb.cert.org/vuls/id/534320). [Source](https://www.stepsecurity.io/blog/ctrl-tinycolor-and-40-npm-packages-compromised)
+- **Axios (March 2026)**: Attackers hijacked the lead maintainer of Axios (100M+ weekly downloads) and published versions containing a "phantom dependency" that existed solely to trigger its `postinstall` hook, deploying a cross-platform RAT. The malicious package was never imported in Axios source code. [Source](https://www.trendmicro.com/en_us/research/26/c/axios-npm-package-compromised.html)
+
+### Install scripts are a distinct threat class
+
+A common response is "you're going to run the code anyway." But install scripts are different from application code in several ways that matter:
+
+1. Install scripts run without any `require()` or `import`. They fire just by being in the dependency tree. This means they can attack systems that never intend to run the package's code at all: frontend-only projects, build machines, CI pipelines.
+
+2. Install scripts may run under different privileges than the application. The widespread use of `--unsafe-perm` as a troubleshooting fix means many environments run install scripts as root.
+
+3. Many organizations trigger `npm install` automatically in response to pull requests. Install scripts are often the only code execution surface on build machines that otherwise only compile and bundle.
+
+4. A typo like `npm install canvsa` triggers the install script immediately, before the developer can notice the mistake and cancel. Without install scripts, the typosquatted package would sit inert until explicitly imported.
+
+5. Adding a dependency to `package-lock.json` in a pull request is easy to miss: GitHub hides large lock file diffs by default, and few reviewers read them carefully. The install script runs the next time anyone runs `npm install`, without the attacker getting any application code reviewed.
+
+6. Process sandboxing, import policies, and code review all target runtime code. None of them cover code that runs during installation.
+
+### The ecosystem has moved on
+
+When this RFC was first proposed in 2021, a common objection was that too many packages depend on install scripts, particularly native addons that compile via `node-gyp`. Since then, the ecosystem has shifted to prebuilt platform-specific binaries distributed via `optionalDependencies`:
+
+- esbuild -> `@esbuild/linux-x64`, `@esbuild/darwin-arm64`, etc.
+- SWC -> `@swc/core-linux-x64-gnu`, `@swc/core-darwin-arm64`, etc.
+- Sharp -> `@img/sharp-linux-x64`, etc.
+- Rollup -> `@rollup/rollup-linux-x64-gnu`, etc.
+- lightningcss, Biome, oxc: all follow this prebuild pattern.
+
+[Node-API](https://nodejs.org/api/n-api.html) provides ABI stability across Node.js versions, making prebuilt binaries viable without recompilation. The packages that still require install scripts are a small minority and getting smaller.
+
+### npm is the last holdout
+
+Every other major JavaScript package manager already blocks dependency install scripts by default:
+
+| Package manager | Default behavior                                                   | Since                           |
+|-----------------|--------------------------------------------------------------------|---------------------------------|
+| pnpm            | Blocked; allowlist via `allowBuilds` in `pnpm-workspace.yaml`      | v10 (January 2025)              |
+| Yarn Berry      | Blocked; per-package opt-in via `dependenciesMeta.built`           | v2 (2020)                       |
+| Bun             | Blocked (except ~400 default-trusted); `trustedDependencies` array | Since install support was added |
+| Deno            | Blocked; per-package opt-in via `--allow-scripts=<packages>`       | Since npm compat was added      |
+| npm             | Runs all scripts by default                                        | —                               |
+
+npm is the only one that still runs everything by default.
+
+## Scope
+
+This RFC covers the following lifecycle scripts when they are defined by dependencies (not the root project):
+
+- `preinstall`
+- `install`
+- `postinstall`
+- Auto-detected `binding.gyp` files (which trigger implicit `node-gyp rebuild`)
+- `prepare` (for non-registry sources only: git dependencies, local file/link dependencies)
+
+The following are out of scope:
+
+- Scripts defined in the root project's `package.json` (these always run, as they are under the developer's direct control)
+- Lifecycle scripts triggered by explicit user action (`test`, `start`, `stop`, `restart`, `publish`, etc.)
+- Registry-level changes (2FA requirements, package signing, provenance attestations)
+- Runtime code isolation or sandboxing
+
+## Detailed Explanation
+
+### Design overview
+
+The design is modeled on pnpm v10's `allowBuilds` system, adapted for npm conventions:
+
+1. A new `allowScripts` field in `package.json` declares which dependencies are permitted to run install scripts.
+2. A new `npm approve-scripts` command writes allowlist decisions to `package.json` from the command line.
+3. Dependency install scripts not covered by the allowlist are blocked, with a clear warning and remediation instructions.
+4. A phased rollout allows the ecosystem to migrate gradually.
+
+### The `allowScripts` field
+
+A new top-level field in `package.json` maps package name patterns to `true` (allowed) or `false` (denied):
+
+```json
+{
+  "allowScripts": {
+    "canvas": true,
+    "sharp": true,
+    "core-js": false,
+    "nx@21.6.4 || 21.6.5": true
+  }
+}
+```
+
+The field uses three values:
+
+| Value   | Behavior                                          |
+|---------|---------------------------------------------------|
+| `true`  | Install scripts are permitted                     |
+| `false` | Install scripts are blocked silently (no warning) |
+| Absent  | Install scripts are blocked with a warning        |
+
+This three-value design (allow / deny / unreviewed) matches pnpm's `allowBuilds` semantics. The distinction between `false` and absent is important: `false` means "I have reviewed this package and decided it does not need scripts," while absent means "this package has not been reviewed yet."
+
+Package entries may include version constraints using the `@` separator:
+
+```json
+{
+  "allowScripts": {
+    "sharp": true,
+    "nx@21.6.4 || 21.6.5": true,
+    "sqlite3@5.1.7": true
+  }
+}
+```
+
+A name-only entry (e.g., `"sharp": true`) allows all versions. A versioned entry (e.g., `"nx@21.6.4 || 21.6.5": true`) restricts the allowance to specific versions, using exact versions joined by `||`. Semver ranges like `^`, `~`, `>=`, or `<` are not supported. This is intentional: a range like `nx@<21.6.4` would automatically trust future versions that haven't been reviewed, which defeats the purpose of an allowlist. If both a name-only entry and a versioned entry exist for the same package, the versioned entry takes precedence for matching versions.
+
+This matches pnpm's `allowBuilds` design, which also restricts versioned entries to exact versions with `||` disjunction.
+
+If overlapping versioned entries assign different values to the same resolved version (for example, `"pkg@1 || 2": true` and `"pkg@2 || 3": false` both match version `2`), the `false` value wins. Deny-wins is the safer default; users who want a specific version allowed despite an overlapping deny should narrow the deny entry.
+
+Non-registry dependencies (git, file, tarball) use [`package-spec`](https://docs.npmjs.com/cli/v11/using-npm/package-spec) syntax. See [Identity matching](#identity-matching) below for the supported key forms.
+
+The `allowScripts` field is only read from the root project's `package.json` (or workspace root). `allowScripts` fields in dependency `package.json` files are ignored. This is a consumer-side policy, not a publisher declaration.
+
+### Identity matching
+
+Package names in npm are not a trustworthy identity for security policy. Each `allowScripts` key is matched against a node's resolved identity from the lockfile, not against the package's self-reported name.
+
+Two ecosystem flaws make name-based matching unsafe:
+
+1. The npm alias mechanism. `npm install trusted@npm:naughty` installs `naughty` under the folder `node_modules/trusted`. In the resulting tree, `node.name` is `"trusted"` (the alias / folder name) and `node.package.name` is `"naughty"` (the tarball's self-report). An allowlist that matched on either field would let an attacker bind a malicious package to a trusted name by getting it installed via an alias.
+
+2. Manifest confusion. The npm registry does not validate that a published tarball's internal `package.json` matches the manifest the registry serves. (Disclosed [March 2023](https://blog.vlt.sh/blog/the-massive-hole-in-the-npm-ecosystem); unfixed as of 2026.) A package published as `naughty` can ship a tarball whose `package.json` declares `name: "trusted"`. After install, `node.package.name` is `"trusted"` regardless of where the tarball came from.
+
+Both problems apply to any allowlist that matches by package name, including a map-based one. The fix is to match against the lockfile's `resolved` field, which is set by the resolver, not by the tarball.
+
+#### What the key matches against
+
+Keys are parsed with [`npm-package-arg`](https://github.com/npm/npm-package-arg), the same parser that handles `npm install <spec>` and `package.json` dependency entries. The parsed spec is matched against each node's resolved identity, recorded in `package-lock.json`:
+
+| Dependency type         | Resolved identity                                                                                                                                     |
+|-------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Registry (version)      | `name@version` (e.g. `sharp@0.34.0`)                                                                                                                  |
+| Registry (range)        | `name` matches any resolved version; `name@version` matches exactly                                                                                   |
+| Alias                   | The underlying registered package, not the alias. `npm install trusted@npm:naughty@1.0.0` matches against `naughty@1.0.0`, never `trusted`.           |
+| Git                     | The resolved git ref (e.g. `npm/cli#c12ea07`). A name-only entry like `"npm/cli"` permits any resolved commit of that repository.                     |
+| File / tarball / folder | The resolved URL or path (e.g. `https://example.com/foo.tgz`, `file:./vendor/foo`).                                                                   |
+
+Examples:
+
+```json
+{
+  "allowScripts": {
+    "sharp": true,
+    "sharp@0.34.0": true,
+    "sharp@0.33.2 || 0.33.3 || 0.34.0": true,
+    "npm/cli": true,
+    "npm/cli#c12ea07": true,
+    "@myorg/internal-tool": true
+  }
+}
+```
+
+A name-only entry permits any resolved version or commit: `"npm/cli": true` matches any commit of that repository just as `"npm": true` matches any version of the package. A fully-qualified entry (`"npm/cli#c12ea07"`, `"sharp@0.34.0"`) pins exactly one.
+
+Three implementation notes follow from the table. For registry-resolved deps, the matcher uses `name@version` rather than the full `resolved` URL. This handles the `omitLockfileRegistryResolved` config (a user setting that omits `resolved` for registry packages from the lockfile). For git deps, the key's committish (the part after `#`) is matched as a prefix of the resolved full SHA: `"npm/cli#c12ea07": true` matches a node resolved to `git+ssh://git@github.com/npm/cli.git#c12ea07a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e`. Users typically write short SHAs; the resolver always pins to the 40-character commit hash. For file, tarball, and remote URL deps, the `resolved` value is matched as an exact string.
+
+#### Fields the key MUST NOT match against
+
+Implementations MUST NOT match keys against:
+
+- `node.name` (the install location / folder name)
+- `node.package.name` (the tarball's self-reported `package.json#name`)
+- `node.package.version` (the tarball's self-reported `package.json#version`)
+- `node.package.repository` (self-reported and unverified)
+
+All four are attacker-controllable through manifest confusion or alias installs. The lockfile's `resolved` field, set by the resolver from the registry response or the user's git/tarball/file spec, is the only field set independently of the tarball.
+
+#### Forward compatibility with a query selector grammar
+
+The key syntax above is forward-compatible with npm's existing [dependency selectors](https://docs.npmjs.com/cli/v11/using-npm/dependency-selectors) (`npm query`). Today, only package-specs are accepted as keys. If a future arborist version exposes a trustworthy queryable identity (see Unresolved Questions), the same key position could accept full selector syntax (`":root > .prod"`, `":type(git)"`, etc.) alongside today's bare-name and `name@spec` keys. The bare-name keys would then read as sugar over a name-equality selector against the identity field.
+
+The full selector grammar is intentionally out of scope. Today's `#name` and `[name=...]` selectors match against `node.name` and `node.package.name`, which the identity rule above excludes. Shipping a selector-based allowlist before arborist has a trustworthy queryable identity would inherit the same exposure; extending the query engine to add one is a design effort of its own. [RFC #861](https://github.com/npm/rfcs/pull/861) reached the same conclusion ("the query language has no trustworthy package identity").
+
+### Policy layering
+
+Script permissions are resolved from a single layer in the precedence chain below (highest to lowest). Layers are **not** merged:
+
+1. CLI flags (`--allow-scripts`, `--no-scripts`, `--dangerously-allow-all-scripts`)
+2. Project `package.json` (`allowScripts` field)
+3. Project `.npmrc` (`allow-scripts` setting)
+4. User `.npmrc`
+5. Global `.npmrc`
+
+The first layer that defines any allowlist configuration wins for the entire install; lower-precedence layers are not consulted at all. This avoids surprising merge semantics where, for example, a project `package.json` entry for one version is silently augmented by a `.npmrc` entry for another. If a project sets `"allowScripts": { "sharp@0.34.0": true }` in `package.json` and the installed version resolves to `0.35.0`, the install fails the way any other unmatched-version case fails — regardless of what `.npmrc` says.
+
+The `.npmrc` setting is intended for contexts without a project `package.json` `allowScripts` field at all, such as `npm install -g` and `npx`:
+
+```ini
+; ~/.npmrc
+allow-scripts = canvas, sharp, sqlite3
+```
+
+### The `npm approve-scripts` command
+
+A new command writes approval decisions to the `allowScripts` field in `package.json`:
+
+```sh
+# Approve specific packages
+npm approve-scripts canvas sharp
+
+# Approve all packages with pending scripts
+npm approve-scripts --all
+```
+
+Approved packages are written with `true`, and newly approved packages are immediately rebuilt. Denial is a separate command, [`npm deny-scripts`](#the-npm-deny-scripts-command), described below.
+
+`--pin` controls only how approvals are written. Denials are always recorded name-only (`"pkg": false`), since pinning a denial to a specific version would silently allow a future version to run scripts again. This is intentional: pinning is conservative for approve and permissive for deny, so the asymmetry favours the safer default in both directions.
+
+When an entry for a package already exists in `allowScripts` and the installed version differs, the command's behaviour depends on the shape of the existing entry. Single-version pins are auto-rewritten to track the installed version; multi-version statements (`pkg@a.b.c || x.y.z`) are never modified, since they likely capture user intent the command can't infer; entries for removed packages are left alone; and existing `false` entries always win:
+
+| Existing entry                | Installed version             | `--pin=true` (default)                              | `--pin=false`                                       |
+|-------------------------------|-------------------------------|-----------------------------------------------------|-----------------------------------------------------|
+| `pkg: true`                   | (none, package removed)       | no edit                                             | no edit                                             |
+| `pkg@a.b.c: true`             | (none, package removed)       | no edit                                             | no edit                                             |
+| `pkg@a.b.c: true`             | `pkg@x.y.z`                   | rewrite to `pkg@x.y.z: true`                        | rewrite to `pkg: true`                              |
+| `pkg: true`                   | `pkg@x.y.z`                   | upgrade to `pkg@x.y.z: true`                        | leave as `pkg: true`                                |
+| `pkg@a.b.c \|\| d.e.f: true`    | `pkg@x.y.z`                   | add `pkg@x.y.z: true`                               | add `pkg: true`                                     |
+| `pkg@a.b.c \|\| x.y.z: true`    | `pkg@x.y.z` (already covered) | no edit                                             | no edit                                             |
+| (no entry)                    | `pkg@a.b.c` and `pkg@x.y.z`   | write both pinned                                   | write `pkg: true`                                   |
+| `pkg: false`                  | any                           | no edit (existing deny wins)                        | no edit (existing deny wins)                        |
+| `pkg@a.b.c: false`            | `pkg@x.y.z`                   | no edit; `pkg@x.y.z` remains unreviewed, with a warning | no edit; `pkg@x.y.z` remains unreviewed, with a warning |
+
+Pinned deny entries (`pkg@a.b.c: false`) are not auto-rewritten when the package version changes, since that would either silently re-allow or silently re-deny the new version. The currently-installed version is left unreviewed and the command prints a warning suggesting the user run `npm deny-scripts pkg` (to re-deny name-only) or remove the pinned deny entirely.
+
+The deny-wins conflict rule from the [allowScripts field](#the-allowscripts-field) section applies here too: an existing `false` entry is never overwritten by `--all`. Users who want to flip a deny to an approve must edit the entry by hand.
+
+List packages with install scripts that are not yet covered by the resolved policy, without writing any changes:
+
+```
+$ npm approve-scripts --pending
+The following packages have install scripts that are not approved:
+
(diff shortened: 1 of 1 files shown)


brunoborges:
I'm sure this was considered by the team, but it would be great if they can provide insights: what is the rationale for keeping install scripts feature in the first place? In other words: why not get rid of this feature completely?

leobalter:
@brunoborges The short answer is that a small but legitimate tail of packages still relies on install scripts (`canvas`, `sharp`, `sqlite3`, and similar native addons), and `--ignore-scripts` today is all-or-nothing, which is why it isn't a practical substitute.

The ecosystem has largely shifted to prebuilt binaries via `optionalDependencies` (`esbuild`, `SWC`, `Sharp`, `Rollup`, `Biome`, etc.), so the remaining surface is small - but it's non-zero, and removing the feature would break those packages with no clean migration path.

The actual problem isn't that the feature exists, it's arbitrary execution from unaudited transitive deps, and that's addressable by making scripts opt-in with an explicit per-package allowlist. That's the direction this RFC takes, and it leaves room to tighten further in future majors.

brunoborges:
@leobalter I am afraid the RFC does not address the root of the problem that libraries are allowed to have install scripts in the first place. It still leaves room for laziness as an attack vector.

> The actual problem isn't that the feature exists, it's arbitrary execution from unaudited transitive deps, and that's addressable by making scripts opt-in with an explicit per-package allowlist. 

I'd argue that the existence of the feature is the actual problem. Once the feature exists, any enhancement is an additional guardrail. 

One approach to consider is to distinguish between executable tools and libraries as published packages.

AllanOricil:
It would also be useful if `--ignore-scripts` combined with `--verbose` displayed the exact lifecycle scripts that would have been executed, including their resolved paths (real ones not symlinks) and content hashes.

With this output we can then edit the new `--allowScripts` flag with the exact output showed, including its sha, instead of just the package name

```json
"allowScripts" : [
    "path/to/script.extension?sha256=foo"
]
```

The current allowScripts proposal is still too permissive if approvals are scoped only to package names.

A package can change its lifecycle scripts between versions — or even within the same version range — which means the developer may end up executing code that was never actually reviewed or approved.

Because of that, approvals should be bound to the exact script file being executed, not just the package identity.

JamieMagee:
> It would also be useful if `--ignore-scripts` combined with `--verbose` displayed the exact lifecycle scripts that would have been executed, including their resolved paths (real ones not symlinks) and content hashes.

@AllanOricil I like this idea, but I think it's orthogonal to the rest of your proposal. A dry-run view of "here's what would actually run, with paths resolved through symlinks" is a useful diagnostic on its own, however approvals end up scoped. Worth its own RFC IMO.

> With this output we can then edit the new `--allowScripts` flag with the exact output showed, including its sha, instead of just the package name
>
> ```json
> "allowScripts" : [
>     "path/to/script.extension?sha256=foo"
> ]
> ```

The part I'm not sold on is hashing the script file itself. Lifecycle script values in `package.json` aren't file paths, they're shell command strings. A real `postinstall` looks like `node scripts/setup.js && chmod +x ./bin/*`. To meaningfully hash "what's about to run" you'd have to resolve every binary referenced on `PATH`, hash the transitive JS closure reachable from the entry script, and re-hash any time the Node.js or toolchain version changes. So every dependency update forces a full re-review of every script. In practice that's the dependency-approval equivalent of a default-deny TOFU prompt; people click through, and the signal disappears.

> The current allowScripts proposal is still too permissive if approvals are scoped only to package names.

Agreed when it's name-only. The RFC's answer is that you can pin a version: `"allowScripts": { "esbuild@0.20.0": true }` only approves that exact version. Bumping to 0.20.1 puts esbuild back in `npm approve-scripts --pending`. The name-only form is there for convenience on packages whose install scripts you genuinely don't care to gate per-version.

> A package can change its lifecycle scripts between versions — or even within the same version range — which means the developer may end up executing code that was never actually reviewed or approved.

Pinned entries handle the version-to-version case, range keys like `^0.20.0` are rejected, and same-version byte changes are caught by the lockfile's SHA-512 `integrity`.

> Because of that, approvals should be bound to the exact script file being executed, not just the package identity.

`name@version` + lockfile `integrity` already binds approvals to the exact bytes. Per-script hashing only helps if `integrity` is wrong. Do you have a concrete attack in mind that name-plus-version-plus-integrity misses?

JamieMagee:
There's a Phase 1 implementation up at https://github.com/npm/cli/pull/9360 for anyone who wants to look at it alongside the RFC.

It's advisory-only. Scripts still run, but install ends with a grouped warning listing packages whose install scripts haven't been approved via `allowScripts`. The matcher, the `approve-scripts` / `deny-scripts` commands, policy layering, and the three new configs are all wired in. The `arb.rebuild()` gate is what Phase 2 will flip.

Reviews on either thread welcome.

sheplu:
+1 on having something around that
do you want to include part of this RFC something similar for the `allow-git` config?

AllanOricil:
@JamieMagee One case I was thinking about that name@version + lock file integrity does not fully cover is when lifecycle scripts fetch remote content dynamically.

For example, a postinstall script may download an executable from a stable URL:

```json
{
  "postinstall": "node scripts/install.js"
}
```

where install.js does something like:

```js
https.get("https://example.com/tool/latest")
```

In that scenario:

- the package version remains unchanged,
- the lockfile integrity remains valid,
but the actual executed payload can still change server-side at any time.

So name@version + integrity guarantees the integrity of the npm package itself, but not necessarily the integrity of the code ultimately executed during install.

That is partly why I was thinking about approvals at the executable/script level rather than only at the package level, although I agree it still would not completely solve dynamic remote execution.

AllanOricil:
I will open a RFC for the other idea of outputing the list of all lifecycle scripts that would have run, if that isn't shown already. Maybe that info should be displayed as a warning, in a parsable way. We could use the output to manually analyze each executable or use a consensus of many LLMs analysis to better warn developers. 

Maybe NPM should do this LLM analysis at publish time before making a package available?

AllanOricil:
Someone else can write the RFC for the output I mentioned above. I'm too occupied and won't have time for that until I finish my other project. I think the list of scripts can be available when install is combined with --dry-run (not sure if it exists for install) and --verbose.

Like I said above, I don't believe allow scripts alone will protect users when scripts can be dynamic without changing the integrity hash. A human, or AI, review must be done and for this to happen it would be necessary to have a list of all scripts that would have run so that a manual review can be performed.

AllanOricil:
Our friend Claude wrote this script to let us know all packages that would have run during installation. (needs review)


```bash
#!/usr/bin/env bash
# install-script-audit.sh
# Lists every script that would have run during install, so each can be audited.
# Run your install with `--ignore-scripts` first, then run this.

set -euo pipefail

ROOT="${1:-.}"
cd "$ROOT"

[ -d node_modules ] || { echo "node_modules not found in $ROOT" >&2; exit 1; }

TMP=$(mktemp -d)
trap 'rm -rf "$TMP"' EXIT

# 1. Declared lifecycle scripts in every package.json under node_modules.
find -L node_modules -name package.json -type f 2>/dev/null \
| while IFS= read -r pj; do
    jq -r --arg path "$pj" '
      select(.scripts != null) | . as $r |
      ["preinstall","install","postinstall","prepare"][] as $h |
      select($r.scripts[$h] != null) |
      [$r.name // "?", $r.version // "?", $h, $r.scripts[$h], $path] | @tsv
    ' "$pj" 2>/dev/null || true
  done > "$TMP/declared.tsv"

# 2. Packages with binding.gyp -> implicit `node-gyp rebuild` on install.
find -L node_modules -name binding.gyp -type f 2>/dev/null \
| while IFS= read -r gyp; do
    pj="$(dirname "$gyp")/package.json"
    [ -f "$pj" ] || continue
    jq -r --arg path "$gyp" '
      [.name // "?", .version // "?", "binding.gyp",
       "(implicit: node-gyp rebuild)", $path] | @tsv
    ' "$pj" 2>/dev/null || true
  done > "$TMP/native.tsv"

# 3. Merge, dedupe by name@version + hook, sort.
cat "$TMP/declared.tsv" "$TMP/native.tsv" \
| awk -F'\t' 'NF>=4 && !seen[$1"@"$2"|"$3]++' \
| sort -t$'\t' -k1,1 -k2,2 -k3,3 \
> install-script-audit.tsv

# 4. Cross-check against package-lock.json if present.
if [ -f package-lock.json ]; then
  jq -r '
    .packages // {} | to_entries[] |
    select(.value.hasInstallScript == true) | .key
  ' package-lock.json | sort -u > install-script-audit.lockfile.txt
fi

total=$(wc -l < install-script-audit.tsv | tr -d ' ')
pkgs=$(awk -F'\t' '{print $1"@"$2}' install-script-audit.tsv | sort -u | wc -l | tr -d ' ')
prepares=$(awk -F'\t' '$3=="prepare"' install-script-audit.tsv | wc -l | tr -d ' ')

echo "Wrote install-script-audit.tsv"
echo "  $total entries across $pkgs unique packages"
[ "$prepares" -gt 0 ] && echo "  $prepares 'prepare' entries — only run for git/file deps; skip for registry deps."
[ -f install-script-audit.lockfile.txt ] && {
  echo "  Cross-check: install-script-audit.lockfile.txt"
  echo "  Anything in the lockfile set not in the audit list is suspicious."
}
echo
echo "Columns: name  version  hook  command  path"
echo "View:  column -t -s \$'\\t' install-script-audit.tsv | less -S"
```

With this script, the process for a human or AI audit would be as follows

```bash
npm install --ignore-scripts
chmod +x install-script-audit.sh
./install-script-audit.sh
column -t -s $'\t' install-script-audit.tsv | less -S
```

With this output people can manually, or using AI, review each entry to determine (open the file and read to determine its whole flow) if it has dangerous executables, like dynamic ones that wouldn't have been caught even if the lock integrity hash hasn't changed.

I do believe NPM should do this analysis for us using LLM consensus before making packages visible. NPM can track false positives until the system becomes reliable. Maybe you guys already have access to Mythos. Cant it be used for this?

AllanOricil:
@JamieMagee @leobalter this RFC can be used as a "beacon" to actually help attackers.

This RFC turns a private security signal into a public one.

Today, the list of packages whose install scripts a company permits lives where attackers can't see it: internal CI configs, private Dockerfiles, npmrc files in CI secrets.
After this RFC, the same list lives in a public field in package.json.

The lockfile tells attackers what a company uses. allowScripts tells them what will execute. Only one of those was previously available. Only one of them is what attackers need.

A scraper can collect every public allowScripts field and produce a ranked target list: "compromise this package, land code execution on these companies."
The RFC effectively does that "reconnaissance work" (Google about it) for them.

naugtur:
There's prior art in this repo somewhere. (Sorry, I'm on the phone right now and I was sent a link to it) 

Also check out:   https://www.npmjs.com/package/@lavamoat/allow-scripts  
https://www.npmjs.com/package/can-i-ignore-scripts

AllanOricil:
@naugtur so it is a common practice to share the trust boundaries publicly? It feels insecure to let attackers know their targets. Can someone explain why it is not insecure, please? I would not add this info in my package.json, only as CLI arguments and in memory only

Thank you for sharing this lava moat tool. I never heard of it before.

@bakkot disliked my comment, so I would like to hear counter arguments, please.

naugtur:
The list of packages you allow to run scripts from is easy to derive from your already public dependencies with few false positives.

The realistic goal for disabling scripts by default is to remove the attack surface for the most common way malware reaches people now. Having to compromise a specific package with a preexisting install script is already making your reach as an attacker orders of magnitude smaller. and if you add a specific version to that, the information becomes useless (assuming npm registry remains immutable)

AllanOricil:
I run `npm install package --ignore-scripts` and the scripts I run are listed in my env variables only and never shared.  To target me, an attacker will have to go thorugh N possible dependencies, which can all be determined like mentioned by @naugtur 

After this RFC, I add `allowScripts` in a publicly facing file, and start using `npm install package` because I know that only the trusted scripts will run. Now the attacker knows upfront that he doesn't need to look for N targets, only M (M <= N), which are the ones that are explicitly in the package.json.

Isn't this helping an attacker to do his work faster (reconnaissance work already done for him)? If yes, how isn't this just making things worse?

naugtur:
But in this context, M and N are both numbers of packages they'd have to replace existing versions of.

Meanwhile there's F~=1000*M other packages in the dependency tree that they have to choose as targets to publish new versions of with install scripts now without scripts being disabled.

If you focus on targeted attacks, there are a dozen ways to compromise a project with all scripts off if you assume they can replace a specific package they choose. This is one of the least likely scenarios. 
And with tooling that only runs scripts from precisely identified registry+package+version the information being public doesn't matter. You're going to run scripts from the packages you need to work or uninstall the packages when they don't work. 


TLDR: If I can see your lockfile, I can determine the M packages with a lot of precision by choosing the dependencies that have an install script and won't work without it


This conversation would be easier if I could wave my hands. Can we bring back the NPM RFC call? 😅

AllanOricil:
I don't see where my reasoning drifted from the RFC specs.

Making the list of scripts my org needs to run publicly available, hardcoded in a package.json, is a security flaw because it reduces the number of dependencies the attacker will invest time on. Im defending that I would want to use `allowScripts` but only as CLI arguments, never stored in the repo.

jwalton:
> This RFC turns a private security signal into a public one.

Today, for the vast majority of people using npm, it’s already public.  The list of packages your package will run scripts in is all of the scripts in package.json.  In that sense, this PR doesn’t make anything worse, and if you enable this feature by default (so you have to whitelist pakckages) it makes things vastly better for the majority of people.

Even if you are publicly alllowing, say, express and lodash to run scripts, this hugely reduces the attack surface considering most large projects inevitably have some dependency in their tree with a small number of downloads a week and an overworked maintainer who is happy to take submissions from anywhere.  Most npm packages (including lodash and express) have no reason to run scripts.  A very large number of projects will just disable all install scripts from running, except perhaps their own.

I don’t think anything here is stopping you from continuing to run some alternative solution to this if you’re the paranoid type who’d rather keep this private.

naugtur:
The information we're talking about keeping private is easily guessable from a lockfile or package.json. I don't know how to say it more clearly.

AllanOricil:
@naugtur in my example there is no way to guess which scripts I would run. You can only guess that I would run all or none (N or 0), and then as an attacked you would have to spend time trying to inject attacks in N targets to try to catch me.

After the RFC, my org decided to stop using `--ignore-scripts` and started using `allowScripts` and made it available in package.json. Now attackers will have a much smaller list of targets to try injecting attacks. Isn't this reducing work for an attacker? I didn't see arguments that commiting allowScripts in package.json isn't doing "reconnaissance work". And if it is, then it is a security flaw by definition.

Again, I'm not against using allowScripts, just that it shouldn't read from a file that is usually publicly available.

AllanOricil:
I created an analogy to help you understand my reasoning.

Imagine an attacker wants to breach a company's production environment. The company has 1,000 employees, but only 5 have production deployment access. If the attacker doesn't know who those 5 are, they are forced to cast a wide net. They have to phish broadly, which is slow, noisy, expensive, and highly likely to trigger security alarms.

​Now, imagine the company publishes a directory on their public website: "These are the 5 specific engineers with production access."

​The attacker stops guessing immediately. They skip 99% of their reconnaissance, pick the most vulnerable engineer on that list, and execute a precise, highly targeted attack.

​This is exactly what happens if we culturally accept putting an allowScripts field directly in package.json.

​**To be clear, the mechanism of allowing specific scripts to bypass --ignore-scripts is not the problem**—security-conscious organizations absolutely need that capability to function. **The problem is where that information lives.**

​Currently, the list of trusted packages that are permitted to execute code at install time lives in an opaque layer: private CI configurations, or secure internal documentation. Moving this list to package.json takes a sensitive security configuration and places it in a file that is fundamentally designed to be public, shared, and published.

​By normalizing this in package.json, we are handing attackers a public attack roadmap. **We are explicitly telling them: "If you want to achieve arbitrary code execution on our developers' machines or in our CI pipeline, don't bother guessing—just compromise this specific upstream dependency.**" It drastically reduces the friction of a supply chain attack by doing the reconnaissance work for the attacker. **Security exceptions belong in private environment configurations, not public metadata.**

jwalton:
What packages they actually use are already in package.json, and which of these packages need to run scripts is already public knowledge, no?

AllanOricil:
> What packages they actually use are already in package.json, and which of these packages need to run scripts is already public knowledge, no?

True, but what is not public knowledge is which ones I actually execute. If all were to be executed why would "allowScripts" be needed? Without this information attackers have to try injecting attacks in all. Check if the analogy above makes sense.

jwalton:
I think this is true in the abstract, but practically speaking if I npm install a package and it works fine, I’m not going to allow it to run scripts, and if I npm install a package and it doesn’t work because it uses a post-install script to build some native library, then I’m going to have to add it to my allowScripts.  It makes little sense to install a package that requires script access to function and then deny it the permission to run scripts, and it’s even more strange to install a package that doesn’t need to run scripts and grant it that permission.  So figuring out which scripts have been granted access seems like a fairly trivial exercise.

I also very much suspect of you add this feature, especially if there’s some scary language when you “npm allow scripts …” (Are you SURE?  This could steal all your moneys!) then scripts that require access to run post-install scripts will suddenly decline in popularity, which is also a good thing from a security perspective.

(Also, worth mentioning that whether or not a package can run scripts or not at install time is fine, but once you import the package in node.js, it can run whatever commands it likes and open whatever network connections it wants.  This is only part of the puzzle.)

JamieMagee:
> +1 on having something around that do you want to include part of this RFC something similar for the `allow-git` config?

@sheplu [The default for `--allow-git` is already planned to go to `none` in npm 12](https://github.blog/changelog/2026-02-18-npm-bulk-trusted-publishing-config-and-script-security-now-generally-available/).

Keeping this RFC scoped to install scripts.

---

@AllanOricil `allowScripts` doesn't have to live in `package.json`. The RFC also reads from `--allow-scripts` on the CLI and from `.npmrc` at the project, user, and global levels. Both take precedence over `package.json`, so if you'd rather not publish your `allowlist`, you can keep it in `.npmrc` or pass it via CLI in CI.

Pinning entries to a version (`pkg@1.2.3`) helps too: the registry packages are immutable, so a leaked allowlist isn't very useful. @naugtur already made that point.

And on the bigger framing: most of the recent npm supply-chain incidents (Shai-Hulud, Axios) didn't come from an attacker targeting some specific company's allowlist. They came from a fresh malicious version of some dep in the tree running its install script in passing. That's the surface this RFC is aimed at, and whether your allowlist is public or private doesn't really move the needle there.

AllanOricil:
> Pinning entries to a version (pkg@1.2.3) helps too: the registry packages are immutable, so a leaked allowlist isn't very useful. @naugtur already made that point.

The pinning argument doesn't fully hold here. I mentioned a dynamic-fetch case earlier in the thread — pinning a version guarantees the same package bytes, but it doesn't guarantee what those bytes go fetch at install time. A postinstall that does something like https.get("https://example.com/tool/latest") produces a different executed payload depending on what the URL serves, even with the tarball pinned and the integrity hash valid.

On the broader point: putting the allowlist in package.json does help attackers narrow their target set. Today, knowing which packages a security-conscious org permits to run scripts requires reconnaissance work. With a public allowlist, that work is largely done for them. It introduces a reconnaissance surface that doesn't exist today, separate from whether the RFC succeeds at its stated goal of blocking typosquats.

For example,

Pinning to pkg@1.2.3 guarantees you get the same bytes. It doesn't guarantee what those bytes go fetch at install time.

If pkg@1.2.3's postinstall does https.get("https://example.com/tool/latest"), the tarball is immutable but the executed payload is whatever the URL serves. A leaked allowlist tells attackers which orgs trust which dynamic-fetch scripts. They don't need to compromise the npm package, publish a new version, or phish a maintainer — just control the URL the script fetches from.

Originally-author-malicious is the most realistic path: publish a "clean" version that does dynamic fetching, wait for orgs to allowlist it, flip the URL content later. No npm-side metric flags this — provenance valid, integrity hash matches, no new version published.

So the pinning argument only protects against attacks that modify the package bytes. The public allowlist enables a different attack class that pinning explicitly doesn't cover.

With that said, NPM must not normalize declaring allow scripts in a public facing file. It must not even be possible to do it.

JamieMagee:
> Pinning to pkg@1.2.3 guarantees you get the same bytes. It doesn't guarantee what those bytes go fetch at install time.

You're right on the technical point. I shouldn't have framed pinning + lockfile `integrity` as a defense against leaked allowlists. They defend against new-version compromise (someone publishing a malicious 1.2.4). Your dynamic-fetch case (postinstall calls `https.get`, attacker controls what comes back) is a real attack and pinning doesn't touch it.

> So the pinning argument only protects against attacks that modify the package bytes. The public allowlist enables a different attack class that pinning explicitly doesn't cover.

First sentence yes, second sentence no.

The dynamic-fetch attack works whether the allowlist is public or private. A malicious maintainer doesn't need anyone's `package.json` to know who trusts their script, they get the install traffic directly. For a third-party attacker compromising the fetched-from endpoint, the targeting list is also derivable from public lockfiles: install each dep with `--ignore-scripts`, observe which ones break without their install script, and that's the candidate allowlist. @jwalton made that
point above. A public allowlist makes derivation cheaper, not possible-vs-impossible.

> NPM must not normalize declaring allow scripts in a public facing file. It must not even be possible to do it.

Forbidding the public layer has real costs. Open-source projects legitimately want to share their install-script trust in-repo, the same way they share `.eslintrc`, `.prettierrc`, and `.npmrc`, so the project builds the same for every contributor. Pushing that use case to `CONTRIBUTING.md` instructions and ad-hoc bash is less auditable, not more. The RFC already supports `.npmrc` and CLI for anyone who wants the allowlist private. You can use that path on your own projects regardless of the default. The disagreement is whether to forbid the public option for everyone else.

> Hey, look! These guys trusted this script which fetches executables from this hardcoded URL.

The dynamic-fetch class deserves a real defense, but the right layer for it isn't where the allowlist is stored. It's at install-time network egress: hash-pinned downloads, `--offline` in CI, sandbox network policy, or an RFC that puts `integrity` on fetched content the way the lockfile does for tarballs. Whether allowScripts is in `package.json` or `.npmrc` doesn't change any of that.

AllanOricil:
I asked Gemini Pro to synthesize my reasoning into a final comment.


I’ve been following the discussions around introducing an allowScripts parameter to package.json to let specific packages bypass the --ignore-scripts flag during npm install. While I completely agree that we need a built-in mechanism to selectively allow lifecycle scripts, I strongly believe that culturally accepting its placement inside package.json is a massive security mistake.

​I didn’t read the official MITRE documentation directly before this, so I ran my logic through an AI chat to stress-test my theory. It pulled an exact mapping to a real-world security framework that I think everyone should look at to see if it makes sense.

**​The Analogy: Giving Away the Keys**
​Imagine an attacker wants to breach a company's production environment. The company has 1,000 employees, but only 5 have deployment access. If the attacker doesn't know who those 5 are, they are forced to cast a wide net. They have to phish broadly, which is slow, expensive, and noisy.

​Now, imagine the company publishes a directory on their public website: "These are the 5 specific engineers with production access." The attacker stops guessing immediately, picks a target on that list, and skips 99% of their reconnaissance.

​This is exactly what happens if we put an allowScripts field directly in package.json.

**​The Counter-Argument: "But everyone knows which packages run scripts!"**

​The immediate pushback I got was: "What packages a project uses are already public in package.json, and which of those packages contain install scripts is already public knowledge."

​True, but what is not public knowledge is your internal security policy. **An attacker might see 15 dependencies that want to run scripts, but they have no idea if your CI blocks them all by default. Without knowing what you actually whitelist, an attacker has to gamble on which package to compromise.**

​The moment you add allowScripts: ["package-a"] to a public package.json, you solve the attacker's dilemma. You explicitly tell them: **"Don't bother trying to inject malware into the other 14 packages; we block them. Focus 100% of your energy on Package A, because we have explicitly carved out an active hole in our firewall for it."**

**​The MITRE ATT&CK Mapping (What the AI Found)**
​When I asked the AI to look up how threat models view this, it pointed out that placing security bypass configurations in a public metadata file maps directly to MITRE ATT&CK Reconnaissance Tactic, Technique T1593.003: "Search Open Websites/Domains: Code Repositories."

​Under T1593.003, adversaries actively scrape open-source repositories and public registries for target intelligence to optimize their supply chain attacks. By putting allowScripts in package.json, we are effectively performing the attacker's passive reconnaissance for them. It allows a threat actor to quietly identify vulnerabilities without ever touching private infrastructure or triggering a single internal security log.

**Conclusion**
​The mechanism of allowScripts is great, but the location is the hazard. package.json defines what the software needs to run. Security exceptions define how a specific environment trusts its inputs.

​Take a look at the MITRE ATT&CK T1593 concept alongside this RFC. Does the logic hold up to you? To me, keeping these exceptions in private configurations (like .npmrc or environment flags) isn't just security by obscurity—it’s denying a targeted attacker a free roadmap.

ljharb:
The allow list would only go in applications, not packages, so it'd never be public.

AllanOricil:
> The allow list would only go in applications, not packages, so it'd never be public.

I believe that people will start publishing apps with package.json with "allowScripts" to public github repositories because, as seen here, most don't seen to realize that they will be given away relevant information for an attack.

ljharb:
It could be restricted to only appear in package.json's with `private: true`, or the registry could strip it out at publish time, etc. Either way, attackers already know what scripts are attackable - all of them. Explicitly listing the ones that are allowed is still an improvement, because it's a much smaller list.

paschun:
pnpm v11 disables them by default, but also exits with an error code `1` on install , and in order to get no error code, the user is pressured into auditing the scripts and putting them into the config file under `allowBuilds: pkgname: true|false`. I quite like that workflow.

also, pnpm has [a git repo](https://github.com/pnpm/plugin-trusted-deps) that has a list of "well-known" trustworthy packages with install scripts that you can add to your config. Look in allow.json.
But looks like pnpm just uses the list from bun, [`bun pm default-trusted`](https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt)
[Here](https://github.com/naugtur/can-i-ignore-scripts/blob/main/data.json) is a list from someone that tried to classify the scripts, which is better if possible incase a "well-known" package is compromised. Scripts could be scanned/scored if they do anything potentially naughty. Could also run the scripts sandboxed with node.js [permissions API](https://nodejs.org/api/permissions.html), with no network or fs access.

naugtur:
This conversation is now dominated by a single concern that is the least of our problems and has no supporting evidence of being a problem at all.

In a targeted attack on a public project the information about which scripts run is useless because there's an order of magnitude more packages that the attacker knows are invoked in build scripts or the app itself and make for much better targets (unless all they know is bash, in which case it is unlikely to affect their likelihood of compromising any package at all)

In a non-targeted attack the signal from allowlist doesn't matter because you want to cast a wide net, so what matters is finding popular packages to compromise or find a way to pop many low hanging fruit at scale.

MetaMask has been maintaining a public list of allowed scripts for about 5 years now. 

Let's talk about how to solve adoption and avoid people allowing everything all the time (which is still an improvement on what we currently have)

AllanOricil:
> This conversation is now dominated by a single concern that is the least of our problems and has no supporting evidence of being a problem at all.\n\nIn a targeted attack on a public project the information about which scripts run is useless because there's an order of magnitude more packages that the attacker knows are invoked in build scripts or the app itself and make for much better targets (unless all they know is bash, in which case it is unlikely to affect their likelihood of compromising any package at all)\n\nIn a non-targeted attack the signal from allowlist doesn't matter because you want to cast a wide net, so what matters is finding popular packages to compromise or find a way to pop many low hanging fruit at scale.\n\nMetaMask has been maintaining a public list of allowed scripts for about 5 years now.\n\nLet's talk about how to solve adoption and avoid people allowing everything all the time (which is still an improvement on what we currently have)

That's a fair perspective. I still believe we will see a shift in how automated recon is done once this scales, but I entirely agree that developer friction and the 'allow everything' anti-pattern is the immediate hurdle npm needs to solve for adoption. I'll step back and let the thread focus on the adoption strategies.

matteokoi:
We live in a time of heavy automation: CI/CD, bots, dependency updates, and automated builds. Because of that, control over what can execute during install is more important than ever

MilosHavranek:
NPM packaging should have a major change. Any type of script execution for dependency resolution or project build is insecure and unwanted, if provided with option to whitelist anything, it will be still heavily used and the attack surface while lowering still exist. Best would be to force package maintainers to release final immutable packages that require no script execution and are treated just as files.

riderx:
All script should NEVER run by default.
99% of user dont need them, remove it will fix 99% of problems!

For package who need it still, there is already a NPM command for this `npm rebuild sharp --ignore-scripts=false --foreground-scripts` 

Adding `allowScripts` by the publisher is wrong as it will make bad actor self allow themself.

There is no need to add anything in ANYWHERE.

 People will create themself a script in they own package to run all the postinstall they need in a command.
 
 This has been so much overdue ! 
 
 PLEASE stop overthink this:
 
 Disable ASAP auto run scripts. let people still allow the old behavior in `.npmrc` then that they responsibility not yours.
 
 Instead of making the life of maintainer harder and for no benefit in security with OICD ....

brunoborges:
> We live in a time of heavy automation: CI/CD, bots, dependency updates, and automated builds. Because of that, control over what can execute during install is more important than ever

Which is why I advocate for the complete removal of supporting install scripts for libraries/dependencies. Eventually, developers will just add whatever flag enables install scripts again.

bakkot:
@riderx You have misread the RFC. This does not give publishers the ability to allowlist themselves. It (eventually) disables all scripts by default and gives _consumers_ the ability to opt in to running their own dependencies' install scripts, via `.npmrc` or by the consumer's own `package.json`. It also provides more granularity than is currently possible so that people who need to opt in for certain packages aren't thereby allowing all packages, which is a considerable improvement over the current options.

JamieMagee:
Pushed a revision following feedback on npm/cli#9360

- Phase 1 now actually enforces `true` and `false` entries. Unlisted packages still run, with an advisory warning.
- `strict-script-builds` -> `strict-allow-scripts` to more closely match `allowScripts` in `package.json`
- `--allow-scripts` CLI flag is restricted to `npm exec`, `npx`, and `npm install -g`. Passing it during a project-scoped install/ci/update/rebuild errors out. Project policy belongs in `package.json` or `.npmrc`.
- `dangerously-allow-all-scripts` now actually does what it says in phase 1 (was previously a no-op).

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+
+6. Process sandboxing, import policies, and code review all target runtime code. None of them cover code that runs during installation.
+
+### The ecosystem has moved on
+
+When this RFC was first proposed in 2021, a common objection was that too many packages depend on install scripts, particularly native addons that compile via `node-gyp`. Since then, the ecosystem has shifted to prebuilt platform-specific binaries distributed via `optionalDependencies`:
+
+- esbuild -> `@esbuild/linux-x64`, `@esbuild/darwin-arm64`, etc.
+- SWC -> `@swc/core-linux-x64-gnu`, `@swc/core-darwin-arm64`, etc.
+- Sharp -> `@img/sharp-linux-x64`, etc.
```
bakkot, line 56: Not the best examples, since at least [esbuild](https://www.npmjs.com/package/esbuild?activeTab=code),  [swc](https://www.npmjs.com/package/@swc/core?activeTab=code), and [sharp](https://www.npmjs.com/package/sharp?activeTab=code) still have install or postinstall scripts. Possibly they work without them; I didn't verify.

Did you have AI write this? It's polite to disclose that, if so. (And also you should be verifying every claim it makes yourself, not relying on reviewers to do so.)

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+
+- The root `package.json` `allowScripts` field is the single source of truth for the entire workspace.
+- `allowScripts` is read from the workspace root's `package.json` (the project containing the `workspaces` array) regardless of where the user runs `npm install`. Running it from inside a sub-workspace like `packages/foo/` does not bypass the root policy.
+- Individual workspace `package.json` files do not have their own `allowScripts` fields. All script permissions are managed at the root.
+- If `allowScripts` appears in a non-root workspace `package.json`, npm prints a warning and ignores the field. Silently dropping it would be confusing for developers who placed it there expecting it to do something.
+- This avoids ambiguity about merge semantics and ensures security policy is set in one place.
+
+### Optional dependencies
+
+If a package in `optionalDependencies` has install scripts that are blocked, it is treated as a failed optional dependency installation. This is consistent with existing behavior where optional dependencies that fail to build are silently skipped.
```
bakkot: Not sure this one makes sense. There are a number of projects which have install scripts to optimize something but still function when the install script isn't run. This would make it impossible to use such a project in optionalDependencies.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: These packages use the [`os`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#os)/[`cpu`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#cpu) filter + `optionalDependencies` pattern to ship prebuilt subpackages. Yes, I verified this works with `--ignore-scripts` set (the `postinstall` scripts are merely launch-path optimisations, not required)

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: It wouldn't make them impossible to use, but you would need to add them to `allowScripts`.

That said, I double checked what the implementation, and it already does what you describe (the package installs, but it's install scripts are skipped).

I'll fix the RFC wording in the next iteration to match the implementation. I think this was left over from an earlier draft.

AllanOricil:
@JamieMagee can you enforce the approval of a script to be required at least once per host? I don't want Dev B to blindly accept Dev's A trust policy when he never touched the codebase, specially because of dynamic scripts that can bypass integrity hashes. For example, given a list of allowScripts, each script listed will require a review and approval on every new host before it can ever be executed. Could map host approval/acknowledgement to package, version and integrity and read/write it from ~/.npm ?

bakkot:
@AllanOricil You seem to want this to be addressing a threat model it's just not for. This is about addressing the case where one of your transitive dependencies is compromised and a malicious build script is added, a situation we have seen several times recently. It is not and should not be about addressing threats from other maintainers of a package you're building.

AllanOricil:
> It is not and should not be about addressing threats from other maintainers of a package you're building.

That's a fair read of what I wrote — my framing wasn't clear. But it's not what I meant. I'm not talking about maintainers of my own package, and I'm not distrusting a specific person.

The concern is that allowScripts turns one developer's review decision about a transitive dependency's script into a committed artifact everyone inherits. Reviews are fallible. A developer can approve a script by mistake — for example, allowlisting one whose install step fetches and runs code from a remote server, without realizing that makes it a live compromise vector. Once that approval is in allowScripts, every colleague who clones and installs runs that script on the strength of a single, possibly-mistaken review — including after the original reviewer has left the org.

So this is inside the compromised-transitive-dependency model you described, not outside it. allowScripts defends against that case by relying on human review to decide which scripts are safe. My point is that human review is fallible, and the proposal propagates one person's review — mistakes included — to the whole org without anyone else looking.

The mitigation: require each host to review each allowlisted script at least once before it runs there (acknowledgment keyed to package + version + integrity, stored in ~/.npm). A single reviewer's mistake then isn't silently inherited — each developer gets a chance to catch it, the way independent review catches errors one reviewer misses. It doesn't make any individual review infallible; it stops one fallible review from becoming an org-wide fact.

If this is better as a follow-up RFC than in scope here, that's a fine answer — I wanted it clear it's a supply-chain concern about inherited, fallible review, not a complaint about any maintainer.

bakkot:
In order for one developer's decision to allowlist a script to affect another, it must be committed and pushed to a shared `package.json`. Teams which care about security are going to require a different person to review that before it lands. So it will not be relying on a single person unless your team is already in the habit of landing unreviewed changes or approving changes without actually reviewing them, in which case forcing every member of the team to personally approve everything in `allowScripts` will not save you.

AllanOricil:
> Teams which care about security are going to require a different person to review that before it lands. So it will not be relying on a single person unless your team is already in the habit of landing unreviewed changes or approving changes without actually reviewing them, in which case forcing every member of the team to personally approve everything in allowScripts will not save you.

Fair — PR review does add a second reviewer. On a team that requires it, the allowScripts change gets reviewed before it lands, so I'll drop the "single reviewer" framing.

But there's a population PR review doesn't reach at all: everyone who clones an open-source project and runs npm install. Contributors, people building from source, downstream projects, forks — none of them took part in the maintainer's review. They inherit the maintainer's allowScripts decision without ever seeing it.

That's where this stops being a team-trust question. For a public repo, three things line up:
- the allowScripts list is committed and public, so an attacker can read exactly which scripts every cloner will run;
- if one of those scripts is dynamic — fetches and executes remote code — its payload can change server-side, with no version bump, no integrity change, nothing for PR review or the lockfile to catch;
- everyone who builds the project after that change runs the new payload, on a review they never made.

So a popular project's public allowScripts becomes a targeting list: compromise one listed package's remote payload, hit everyone who builds the project. And the timeline is the whole problem — the review happens at time T, when the script is benign; the payload turns malicious at T+N; everyone who builds between T+N and the day someone finally notices runs the compromised payload, on a review made before the threat existed. **PR review is point-in-time: it covers T, not T+N, and it never involved the cloners at all.**

To be clear about what I'm not claiming: this isn't allowScripts making things worse than today — today those scripts run on every clone with no review at all. The point is that allowScripts records trust in a package's identity (version + integrity), while the risk lives in behavior the reviewer couldn't see and the cloner never reviewed. Another approval step doesn't change that; surfacing what the script does — e.g. "this fetches and executes remote code" — at the point it would run at least lets someone decline the pattern.

Probably its own follow-up rather than in scope here, but it's why I don't think identity-based approval is sufficient on its own.

bakkot:
All review is point in time. If your threat model includes install scripts changing behavior after review, requiring each person to review makes little difference. The compromise could just as well happen after any given person has already reviewed. And surfacing information like whether a script is fetching remote code - given that many scripts are running arbitrary shell commands anyway - is not feasible without completely redoing how scripts execute.


This conversation really gives me the feeling of arguing with someone who is just copy pasting LLM output. Regardless of whether that's true I don't think it will be productive for me to continue discussing this with you so I'm not going to respond further.

AllanOricil:
> This conversation really gives me the feeling of arguing with someone who is just copy pasting LLM output. Regardless of whether that's true I don't think it will be productive for me to continue discussing this with you so I'm not going to respond further.

Calling my argument LLM output is a convenient ad hominem exit, but the threat model is real. "Stale trust" when developers leave a company is a massive enterprise security risk. But since you've checked out, let's leave it there.

accepted/0000-make-scripts-install-opt-in.md
bakkot: Ah, makes sense. Maybe a note? E.g.

> Some of these packages still include install scripts to optimize layout after installation, but they all function without them.

BrunoBernardino:
@AllanOricil I'm nobody, but I think your request makes sense as a separate RFC, since this is focusing on the described threat model which is a subset of what seems to be yours, and this one would already be an amazing improvement to `npm`. It definitely feels like "scope creep" to me, even if it's a valid concern and security improvement.

owlstronaut (dismissed):

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Bundled dependencies
+
+Bundled dependencies are packages shipped inside a parent package's tarball. The lockfile marks them with `inBundle: true`, but they have no independent `resolved` URL since they were never fetched on their own. Bundled deps with install scripts are treated as unreviewed and blocked with a warning. Allowlisting them is deferred to a follow-up RFC: matching by `name@version` from the bundled tarball would reintroduce manifest confusion, and a safe parent-qualified syntax needs its own design.
```
everett1992: I was thinking more about bundled dependencies. I know  the code base I manage has a few instances of packages that depend on bundlers that have scripts. These instances won't be able to use `strict-allow-scripts` because the unreviewed (unreviewable) bundled dependency will fail. 

What do you think about approvals for the bundler to apply to the bundled dependencies? 
```
"allowScripts": {
  "a-builder@1.0.0": true,
},
"dependencies": {
  "a-builder": "^1.0.0"
}
```
all scripts in packages bundled by the approved `a-builder` are allowed to run.

* When a package@version is approved we're protected by integrity / immutability. a-bundler can't add new dependencies with scripts, so we've approved an explicit set
* If a whole package name is approved there are more risks, an update could pull in compromised packages,
  you're not only exposed to the approved package being compromised. 

I'm still happy with the RFC and deferring this for a followup.

owlstronaut:
Moderation note: I've removed a comment in this thread for tone. The substantive concerns about stale trust and dynamic-fetch payloads are real and worth pursuing, and the community is welcome to open an RFC for further discussion of that. For this RFC, let's keep discussion focused on the mechanism's herein and civil per the Code of Conduct

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+The first layer that defines any allowlist configuration wins for the entire install; lower-precedence layers are not consulted at all. This avoids surprising merge semantics where, for example, a project `package.json` entry for one version is silently augmented by a `.npmrc` entry for another. If a project sets `"allowScripts": { "sharp@0.34.0": true }` in `package.json` and the installed version resolves to `0.35.0`, the install fails the way any other unmatched-version case fails — regardless of what `.npmrc` says.
+
+The `.npmrc` setting is intended for contexts without a project `package.json` `allowScripts` field at all, such as `npm install -g` and `npx`:
+
+```ini
+; ~/.npmrc
+allow-scripts = canvas, sharp, sqlite3
+```
+
+### The `npm approve-scripts` command
```
manzoorwanijk, line 234: IMHO the syntax here is brought in from `pnpm approve-builds`, and we should probably use what npm uses already.
So it could be `npm scripts approve` and `npm scripts deny`, which matches other npm commands like `npm teams <subcommand>`, and `npm cache <subcommand>`.

That way the docs reference stays on a single page, and those are not two different commands.

owlstronaut: This is a good point, while the inverse commands aren't totally unprecedented here either. I'm landing what we have for phase 1 now. I think we could consider making the approve-scripts and deny-scripts aliases and add the scripts namespace for phase 2. @JamieMagee do you haver a gut  reaction?

JamieMagee: Yeah, I mention it in various places throughout the RFC that pnpm was inspiration for the design of this functionality. I can certainly implement more npm-like commands (`npm scripts approve` and `npm scripts deny`) alongside the existing proposals (`npm approve-scripts` and `npm deny-scripts`).

manzoorwanijk: Yes, I also did the same for native dependency patching proposal - #862 and then updated it to use npm style sub command approach.

I think once we ship this feature, it’s probably time to handle #862 😁

JamieMagee:
Phase 1 has been completed and published 🎉

- https://github.com/npm/cli/releases/tag/v11.16.0
- https://www.npmjs.com/package/npm/v/11.16.0

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: I've had a little time to think on this more, and my worry is that an `npm scripts` command may be confused with the `scripts` block in `package.json`. They do related, but different, things.

manzoorwanijk: `--ignore-scripts` is similar too, and I think that is fine.

manzoorwanijk: But I think instead of `npm scripts`, we could call it <s>`npm trust`, with subcommands like

- `npm trust add lefthook`
- `npm trust rm lefthook`
- `npm trust ls`
- `npm trust clear`</s>

EDIT: I didn't realize that `npm trust` command already exists.

Other candidates are

- `npm install-scripts`
    - `npm install-scripts approve lefthook`
    - `npm install-scripts deny lefthook`
    - `npm install-scripts ls`
- `npm hooks`
- `npm lifecycle`

bakkot: I like `npm install-scripts`. A tiny bit more verbose, but it's much clearer.

accepted/0000-make-scripts-install-opt-in.md
everett1992: Bundled dependencies are also the packages that _will be_ included in a packages own tarball when published.

I can see two ways to address those:
 * Do not allow those scripts to be allowlisted, because consumers would not be able to allowlist them. Make that explicit in the errors.
 * Allow the package to approve the scripts, the trust issues outlined here don't apply, these packages were fetched from the registry. 

I think the phase-1 implemenation merged into latest doesn't distinguish between packages that were bundled and packages that will be bundled.

naugtur: Scripts for bundled dependencies should never run. If a package wants to run them, it needs to expose them as its own lifecycle scripts.

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+
+Non-registry dependencies (git, file, tarball) use [`package-spec`](https://docs.npmjs.com/cli/v11/using-npm/package-spec) syntax. See [Identity matching](#identity-matching) below for the supported key forms.
+
+The `allowScripts` field is only read from the root project's `package.json` (or workspace root). `allowScripts` fields in dependency `package.json` files are ignored. This is a consumer-side policy, not a publisher declaration.
+
+### Identity matching
+
+Package names in npm are not a trustworthy identity for security policy. Each `allowScripts` key is matched against a node's resolved identity from the lockfile, not against the package's self-reported name.
+
+Two ecosystem flaws make name-based matching unsafe:
```
naugtur, line 155: npm also trusts the names of builtins and other non-registry sources.

`npm rebuild bcrypt` would run install scripts from `bcrypt` but also a bundled dependency of any other package that called itself `bcrypt` in package.json in the bundled `node_modules/bcrypt` folder.

`pnpm` ignores scripts of bundled dependencies (seems to be ignoring them entirely, just leaving the folder there)

Another way to go is identify packages on the allowlist with prefixes referring to where they're installed from if not a registry (`file: bundled: git: url:`)

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,442 @@
...
+
+- `true`: install scripts run.
+- `false`: install scripts are silently skipped.
+- absent: install scripts run (Phase 1 default) and a post-install advisory warning lists the package so users can review it. In a future major release this default flips to skip-with-warning.
+
+In strict mode (`strict-allow-scripts=true`), the install fails before any scripts run if any dependency has install scripts that aren't covered by a `true` or `false` entry. The `--dangerously-allow-all-scripts` flag bypasses the policy and runs every install script.
+
+The `--allow-scripts` CLI flag is restricted to one-off and global contexts (`npm exec`, `npx`, `npm install -g`). Passing it during a project-scoped `npm install`, `ci`, `update`, or `rebuild` is an error: team-wide policy belongs in `package.json#allowScripts` or `.npmrc`, not in a command-line flag that gets copy-pasted into READMEs.
+
+The existing `--ignore-scripts` flag continues to work as before, disabling all scripts (including root project scripts) regardless of `allowScripts`.
```
naugtur, line 307: @JamieMagee  I missed this earlier - hope it's not too late.

This will make the migration story more risky.

There's no way to prevent old `npm` versions from running when someone has them installed, so it'd be much safer to make it a reasonable default to keep `ignore-scripts=true` in the config while the allowlist is applied. 

As a result, an `.npmrc` that specifies both  `ignore-scripts=true` and an allowlist will **fall back to not running any scripts** on older versions instead of running all scripts on older versions.

The layering section above *suggests* (but doesn't clearly state, at least to me) that it would work that way, so if this is only about the CLI flag, I'm happy with it as-is.

naugtur: Observed behavior is not working with a safe fallback: https://github.com/npm/cli/issues/9450

bakkot: Disagree; see https://github.com/npm/cli/issues/9450#issuecomment-4593516939

accepted/0000-make-scripts-install-opt-in.md
everett1992: That might address both of my comments. It explains the trust model, and forces bundlers to setup the scripts. 

* **Trust model** consumers must review+approve the bundler, they don't need to check bundled deps.
* **Bundler setup** npm would not run scrips of to-be-bundled packages, or allow them to be approved, it could guild the bundler thru setting up a script to run their bundled package's scripts.

However I don't see how that's better than my recommendation that approvals for the bundler carry thru to the bundled dependencies? It's the same as trusting a bundler with postinstall hook 'run-my-bundled-postinstall-scripts', and has the downside of not working for any existing packages, since they don't have this script setup.

naugtur: The recommendation seems like an improvement for the author of a very rare case of a package with bundled dependencies. What it doesn't take into consideration is how much of a burden it is for the vast majority where all tooling we have for analyzing what happens at install time needs to handle this additional complex case. 
When we eliminate the low hanging fruit of ambient install scripts running everywhere, the attackers will have to become more sophisticated to reach the same scale and (besides going for runtime execution) they'll reach for gaps and complexities in how we decide to handle it.

My point here: the convenience for the few authors of packages with bundled dependencies to not have to forward the necessary scripts manually is not worth the burden on absolutely everyone else and the authors of security scanners etc.

JamieMagee: Both cases collapse to the same rule now: bundled scripts don't run. A package that wants its bundled dep's script forwards it as one of its own lifecycle scripts, which then goes through `allowScripts` like anything else. Called this out in the rewritten section.

JamieMagee: Marking this as resolved in your other comment thread.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: Good catch. Added a note to Identity matching: the `<name>` argument resolves to a registry-resolved node's identity, never a self-named bundled folder, and bundled nodes are excluded from rebuild.

accepted/0000-make-scripts-install-opt-in.md
```diff
@@ -0,0 +1,450 @@
...
+$ npm approve-scripts --pending
+The following packages have install scripts that are not approved:
+
+  canvas (postinstall: node-gyp rebuild)
+  sharp (install: node install/libvips && ...)
+
+To approve them, run: npm approve-scripts
+```
+
+`--pending` consults the same precedence stack as install-time enforcement (CLI flag → root `package.json` → project `.npmrc` → user `.npmrc` → global `.npmrc`), so the listed packages match what would be blocked during a real install.
```
JamieMagee, line 280: One piece of feedback we received from the npm v11.16.0 release is that `approve-scripts --pending` is unclear in what it does. Does it approve pending scripts? Or does it just list pending scripts?

I'll investigate splitting this functionality out into a separate command e.g. `npm pending-scripts`, `npm manage-scripts`, etc.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: Updated the wording of the RFC to match the implementation, as discussed above.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: I left it as-is.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: This doesn't need to be resolved in this RFC, and the `approve-scripts --allow-scripts-pending` has already shipped in npm v11.16.0 anyway. We can deprecate it, and add a new top-level command, without the need for the full RFC process.

accepted/0000-make-scripts-install-opt-in.md
manzoorwanijk: @JamieMagee any update on this?

JamieMagee: I think this is a good proposal, and I want to implement it, but I don't think it's a blocker for this RFC. Let me create an issue in `npm/cli` to track it.

owlstronaut (approved):

ljharb:
No package should only be installed globally, and few packages ever _should_ be installed globally.

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: Created https://github.com/npm/cli/issues/9546 to track this

accepted/0000-make-scripts-install-opt-in.md
JamieMagee: Created https://github.com/npm/cli/issues/9545 to track this

mdjr1877-droid:
#860

PR #891, opened 2026-06-06
[RFC] fail installs on publish-trust downgrades

JamieMagee:
Adds an RFC proposing a `trust-policy` setting for `npm install`.

It warns or fails when a package's publish-trust level drops, for example when a version that used to ship via trusted publishing (OIDC) or provenance is later published with weaker evidence or none. The check is trust-on-first-use: it only fires when a package that earned trust loses it, and it runs during resolution, before lifecycle scripts, so it gates a malicious postinstall instead of catching it afterward. It relies only on metadata npm already fetches (`_npmUser.trustedPublisher`,`_npmUser.approver`, `dist.attestations`).

Configuration:
- `trust-policy` (off, warn, error)
- `trust-policy-exclude`
- a date-valued `trust-policy-enforce-since` to grandfather the pre-provenance back-catalogue
- `trust-lockfile`
- a one-off `--allow-trust-downgrade`.

## References

#860

### DIFF
--- a/accepted/0000-trust-policy-downgrade.md
+++ b/accepted/0000-trust-policy-downgrade.md
@@ -1,0 +1,376 @@
+---
+title: Fail installs on publish-trust downgrades (trust policy)
+number: null
+status: proposed
+created: null
+accepted_at: null
+implemented_at: null
+withdrawn_at: null
+implementation: null
+---
+
+# Fail installs on publish-trust downgrades (trust policy)
+
+## Summary
+
+Add a `trust-policy` setting to the npm CLI that detects when a package's publish-trust level drops.
+For example, a package was previously published via trusted publishing (OIDC) with provenance, and a newer version arrives with weaker evidence: provenance only, or nothing at all.
+That drop is a strong signal of a supply-chain compromise, such as a stolen publish token or a hijacked maintainer account.
+Depending on the configured policy, npm can warn (the proposed default) or fail the install (`error`) when it sees one, before any lifecycle scripts run.
+Four knobs tune the behavior: `trust-policy-exclude`, `trust-policy-enforce-since`, `trust-lockfile`, and a one-off `--allow-trust-downgrade` escape hatch.
+This is a trust-on-first-use (TOFU) check modelled on pnpm's `trustPolicy` family.
+
+## Motivation
+
+npm already lets publishers establish strong, verifiable trust in a release:
+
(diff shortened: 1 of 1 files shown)


ljharb:
This is a very bad idea. First of all, OIDC ("trusted" publishing) is one factor, so this is a fundamentally untrustworthy mechanism. Second, provenance provides no actual guarantees, and it is a perfectly legitimate decision for a maintainer to decide to stop publishing from CI or using provenance.

Simply put, provenance is a misfeature that imo should even be removed, and implying it's a "trust signal" is dramatically shredding the security of the npm ecosystem.

The best approach is to forbid any non-2FA publishing - meaning, everything either goes through a staged publish, or uses actual 2FA (iow, local publishing + 2FA, which prior to staged, is the only secure method of publishing to npm that has ever existed).

piiiico:
Red Hat Miasma shows the gap in provenance-only TOFU: 32 @redhat-cloud-services packages republished through the *existing* CI/CD pipeline. Valid SLSA provenance on every malicious version. Trust level never downgraded — a TOFU check sees green.

What actually broke: 32 packages in a 3-hour batch (normal cadence: weekly). Orphan commits with no parent vs. PR-based flow. New `preinstall` hook where none existed before.

If `trust-policy` tracked behavioral cadence — publish frequency, hook introduction, commit ancestry — against a package's own baseline, it catches credential-theft attacks where the pipeline stays the same but the behavior doesn't.

ljharb:
I continue to wait for anyone to point to an actual npm security incident/campaign where provenance would have been a material barrier - in almost a year, the GitHub security team hasn't been able to provide a single one, and nor has anyone else.

Relying on provenance before it would have been useful _once ever in the history of npm_, let alone in a significant number of incidents, seems like a massive waste of resources on security theater.

piiiico:
Miasma is a sharp counterexample. SLSA was *correct*, and 2FA wouldn't have helped: the credential was the CI/CD's own OIDC token. The attacker pushed orphan commits to a repo with `id-token: write` and the legitimate pipeline did the rest.

Anomalous in registry metadata alone:
- 32 packages republished in ~3 hours (normal cadence: weekly)
- New `preinstall` hooks where none existed in prior versions
- Orphan commits in tarball ancestry, vs the usual PR-merged history

None rely on OIDC or any identity signal. Behavior baselines on the package itself, a different axis from a TOFU check.

Wiz: https://www.wiz.io/blog/miasma-supply-chain-attack-targeting-redhat-npm-packages

Issue #817, opened 2025-04-22
[RRFC] Include a `globallink` install strategy


TheBrenny:
## Motivation ("The Why")

There's no doubt that NPM is the butt of a lot of memes relating to how heavy the local `node_modules` folder can get. Especially on big projects, and whne working on multiple big projects over time, the duplication of the same packages across different project spaces can cause a significant impact on hard drive space. A tool already exists to address this issues: PNPM. However, PNPM is not a 1-for-1 replacement - there are some quirks with the commands, and while for the most part it works, these nuances can get in the way of efficient package management, including when sharing code among developers (especially in the Open Source community).

A `globallink` install strategy should be included to address this and would work similar to how the `linked` strategy works now. However, instead of the packages being saved in `./node_modules/.store`, they would be saved in `/path/to/npm/global/node_modules/.store`.

This will also cause an update to the `clean-install` command where the packages (and exact dependencies) for a project are _deleted_ from the global store and re-installed.

Of course, this install strategy can be overridden by specifying a different type at the command line or local config.

#### Current Behaviour

The hoisting of packages to the root of the node modules folder helps to deduplicate packages across the project. However, this is only localised to the project. As an example, I work on many web apps that all use SvelteKit and I need to download and store the entire SvelteKit package every time for each project ([36MB](https://pkg-size.dev/@sveltejs%2Fkit)). Add on the extra frameworks that I use, and that number starts to add up pretty quickly to a couple hundred megabytes.

#### Desired Behaviour
By specifying the `install-strategy` as `globallink` all the packages would be installed into a global store. npm would resolve the package requirements of an install command, determine that a package exists within the store which meets the version criteria and links to it instead of downloading it again. If no package in the store meets the criteria, one is simply downloaded.

I might be oversimplifying the complexity, but the implementation could be an increase in capability of the `linked` install strategy, in that the store that it points to is the global store, rather than the local one.

stefanobaghino:
> This is already what "isolated layout" mode does, I believe.

[...]

Issue #409, opened 2021-07-03
[RRFC] node_modules using high disk space


haneenmahd:
## Motivation ("The Why")

The node_modules is consuming a large disk space. So,. It would be better if it installs the packages in the root of the user in the computer.
In windows the path would look like: `C:\Users\user_name\node_modules`.
Users could install the different versions of a package.
And it would import packages matching the version inside the `package.json`.
Python packages📦 use this kind of concept and it consumes less storage.It only does one installation.And we don't neee to install it again.

### Example

Using less storage space.
Avoiding the large node_modules folder in every single project.
Just one node_modules in the root of the user.
Users could install the different versions of a package.
And it would import packages matching the version inside the `package.json`.

#### Current Behaviour

Currently npm does install the whole project packages in the project folder📁.This takes up the storage of the system.

#### Desired Behaviour

When we run `npm install` it should only install in the root user folder.

### References

* n/a

PR #911, opened 2026-07-11
RFC: Global Virtual Store for Linked Installs

manzoorwanijk:
## Summary

Adds an RFC for an opt-in `global-virtual-store` config for projects using `install-strategy=linked`. The projections that `linked` writes into each project's `node_modules/.store/` move to a single machine-global store root; the project's `node_modules` keeps only the symlinks (junctions on Windows) that `linked` already creates. Each globally eligible projection is extracted once per machine, so every additional project or git worktree resolving the same dependency subtree costs a few megabytes of symlinks and nearly zero install time.

The proposal changes only `@npmcli/arborist` and the npm CLI — no changes to `pacote`, `cacache`, `package-lock.json`, the registry, or `npm publish`.

## Motivation

Parallel coding agents have made git worktrees routine, and every worktree pays for a full `node_modules` today (~2.2 GB and ~24 s per worktree measured on Gutenberg). pnpm shipped [`enableGlobalVirtualStore`](https://pnpm.io/next/git-worktrees) for exactly this workload; a 2.5 GB production monorepo measured its `node_modules` drop to 1.4 MB. A machine-global store is also one of npm's most persistently requested features ([#278](https://github.com/npm/rfcs/issues/278), [#409](https://github.com/npm/rfcs/issues/409), [#817](https://github.com/npm/rfcs/issues/817), [npm/cli#8242](https://github.com/npm/cli/issues/8242) — closed with a pointer to this repo; no RFC followed).

With [RFC 0042](https://github.com/npm/rfcs/pull/436) shipped, the layout already exists: this RFC reparents the store root out of the project, keeping the projection key, the reifier, the symlink shape, and the lockfile unchanged.

## Notable semantics

- Requires `install-strategy=linked`; errors under other strategies.
- Projections become visible only in final form (extract → patch → read-only sweep → atomic rename) and are read-only, files **and** directories — an in-place write anywhere fails loudly instead of silently corrupting every consumer.
- Packages that may run lifecycle scripts are never globally projected — they stay project-local, exactly as plain `linked` today. This is stricter than pnpm, which shares built projections ([pnpm/pnpm#12302](https://github.com/pnpm/pnpm/issues/12302)).
- Git dependencies stay project-local (pack output is not reproducible across npm versions, per [RFC 0048](https://github.com/npm/rfcs/pull/525)).
- Garbage collection is exact via a project registry — a projection referenced by a live project is never pruned.
- Hard caveats stated up front: deployment flows that archive `node_modules` must dereference or reinstall, and the store assumes a single trust boundary.

A separate proposal, [RFC: Content-Addressable Store (#912)](https://github.com/npm/rfcs/pull/912), covers file-level content-addressable storage for every install strategy; the two share a store root layout but neither depends on the other.

See the RFC for the full design, alternatives, implementation plan, tests, and unresolved questions.

---

> **Disclosure**: [Claude Code](https://claude.com/claude-code) was used to draft this PR description and the initial version of this RFC, and to iterate on it — including adversarial review rounds against the npm CLI source with Codex — before submission.

### DIFF
--- a/accepted/0000-global-virtual-store.md
+++ b/accepted/0000-global-virtual-store.md
@@ -1,0 +1,189 @@
+# Global Virtual Store for Linked Installs
+
+## Summary
+
+Add an opt-in `global-virtual-store` config for projects using `install-strategy=linked`. When enabled, the package projections that `linked` today writes into each project's `node_modules/.store/` move to a single machine-global store root, and the project's `node_modules` contains only the symlinks (junctions on Windows) that `linked` already creates — now pointing at shared, machine-global projections. Each globally eligible projection is extracted exactly once per machine; every additional project or git worktree that resolves the same dependency subtree costs a few megabytes of symlinks and nearly zero install time. (Packages excluded from global projection — those that may run lifecycle scripts, and git dependencies — still extract and build per project, exactly as plain `linked` does.)
+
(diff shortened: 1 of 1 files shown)


manzoorwanijk:
@reggi what do you think about this and #912

